PR #1, opened 2016-04-15
initial commit with cluster membership, and a basic local clusterImplementation

CaitieM20:


clusterDemo/main.go
```diff
@@ -0,0 +1,21 @@
+package main
```
dbentley: Most go directories I've seen are all lowercase (no spaces, underscores, or camel case)

clusterDemo/main.go
```diff
@@ -0,0 +1,21 @@
...
+
+import "fmt"
+import ci "scoot/clusterImplementations"
```
dbentley: maybe github.com/scootdev/scoot/clusterimplementations

clusterDemo/main.go
```diff
@@ -0,0 +1,21 @@
...
+
+func main() {
+
+  var cluster = ci.StaticLocalNodeClusterFactory(10)
```
dbentley: Go has too many decl forms. I'd just do cluster := Function()

clusterMembership/clusterMembership.go
```diff
@@ -0,0 +1,21 @@
...
+type Node interface {
+	GetId() string
+	SendMessage(msg string) error
+}
+
+type Cluster interface {
+	/*
+	 * Returns a Slice of Node Ids in the Cluster
+	 */
+	GetMembers() []string
```
dbentley: []string and not []Node?

clusterMembership/dynamicCluster.go
```diff
@@ -0,0 +1,72 @@
...
+		index++
+	}
+
+	return nodes
+}
+
+/*
+ * Adds A Node from the Cluster
+ */
+func (c dynamicCluster) AddNode(n Node) {
```
dbentley: So this access has to be protected for concurrency. Either with a lock or by having it be owned by a goroutine.

clusterProof/main.go
```diff
@@ -0,0 +1,27 @@
+package clusterProof
```
dbentley: Is this still used? (there's another main.go, and this isn't package main, so I bet it doesn't become a binary)

clusterMembership/clusterMembership.go
```diff
@@ -0,0 +1,21 @@
+package clusterMembership
+
+
+type Node interface {
+	GetId() string
+	SendMessage(msg string) error
```
dbentley: How are you supposed to get a Node?

clusterMembership/clusterMembership.go
```diff
@@ -0,0 +1,21 @@
...
+	 * Returns a Slice of Node Ids in the Cluster
+	 */
+	GetMembers() []string
+
+	/*
+	 * Sends a Message to the specified Node in the Cluster,
+	 * Returns an Error if the Node is not in the cluster or sending
+	 * the message fails
+	 */
+	SendMessage(msg string, nodeId string) (error)
```
dbentley: Ahh, so, really it's cluster membership and communication, right? So you might use ZK for the membership, which then gives you an address you use RPC to talk to?

clusterMembership/clusterMembership.go
CaitieM20: The idea here is we don't want users directly invoking the SendMessage method on the nodes.

There is a race condition where I give a process the list of nodes, and then for a dynamic cluster a node is removed.  We want SendMessage to go through the Cluster interface.  So if the message is trying to be sent to a node that no longer exists, we return an error.  

This may need to be modified if we want to implement work distributions that depend on Node specific information like load etc.. that the application state machine can't keep track of itself.

clusterMembership/dynamicCluster.go
CaitieM20: Yes, so my plan is to run ClusterMembership in its own goroutine.  This way all updates to the underlying cluster data structure are single threaded.

clusterProof/main.go
CaitieM20: oops nope, missed deleting this in the rename.

clusterMembership/clusterMembership.go
CaitieM20: So the idea is you don't you just interact with nodes by sending them messages, via the Cluster Interface

clusterMembership/clusterMembership.go
CaitieM20: Yes different Implementations of the Node interface are how we define a worker that exposes a thrift api or http endpoint for example if its on a different machine.

We can also have different implementations of dynamic cluster membership, one that relies on ZK for update or dynamic config files, or even a gossip protocol for example

clusterDemo/main.go
CaitieM20: done!

clusterDemo/main.go
CaitieM20: done!

clusterDemo/main.go
CaitieM20: done! except when declaring empty slice in test files to avoid allocating unnecessary memory

PR #2, opened 2016-05-03
Caitie/workDistribution

CaitieM20:
Work Distribution Interface for selecting a node from the cluster to send work to

Includes two basic implementations Random and RoundRobin

scheduler/main.go is an implementation of this that can be ran.  Uses a static cluster and generated dummy work to local nodes.


clusterMembership/dynamicCluster.go
```diff
@@ -14,13 +11,13 @@ type dynamicCluster struct {
 /*
  * Returns a Snapshot of the Cluster Membership state
  */
-func (c dynamicCluster) Members() []string {
+func (c dynamicCluster) Members() []Node {
```
dbentley: Hrm, it might be worth changing these from (c dynamicCluster) to (c *dynamicCluster). The former can mean it has to make a copy of the struct for each method call. As a result, most code seems to be in terms of pointers, not value types.

clusterMembership/staticCluster.go
```diff
@@ -11,46 +8,19 @@ import "errors"
...
  */
 type staticCluster struct {
-	members    []string
-	membersMap map[string]Node
+	members []Node
 }
 
-func (c staticCluster) Members() []string {
+func (c staticCluster) Members() []Node {
 	return c.members
```
dbentley: It annoys me that we end up returning mutable pointers to our internal data. This is the idiomatic way to do it, just, mentioning that it bugs me each time.

distributor/distributor.go
```diff
@@ -0,0 +1,13 @@
...
+import cm "scootdev/scoot/clustermembership"
+
+/*
+ * Interface for choosing a node in a cluster to send
+ * work to.  When DistributeWork is called a node from
+ * the passed in cluster should be returned, where work
+ * can be scheduled.
+ */
+type Distributor interface {
+	DistributeWork(cluster cm.Cluster) cm.Node
```
dbentley: This is a good intermediate stage.

I think this will grow in at least two ways:
a) take the piece of work (or at least some info about it) to distribute it meaningfully
b) take all the pieces of work in the queue to assign them at once

Do I have that right?

distributor/distributor.go
```diff
@@ -7,7 +7,10 @@ import cm "scootdev/scoot/clustermembership"
  * work to.  When DistributeWork is called a node from
  * the passed in cluster should be returned, where work
  * can be scheduled.
+ *
+ * work is the work to be scheduled
+ * cluster is the Cluster to schedule the work on
  */
 type Distributor interface {
-	DistributeWork(cluster cm.Cluster) cm.Node
+	DistributeWork(work string, cluster cm.Cluster) cm.Node
```
dbentley: Ah, answered my previous comment. Great.

clusterMembership/staticCluster.go
CaitieM20: Same, I was trying to avoid it but then having the Cluster Manager own sending work to nodes seemed broken.  This basically relies on the fact that users won't mutate things, and this should be treated as readonly.

distributor/distributor.go
CaitieM20: Yes, I totally missed that and realized it last night, since the early toy ones don't rely on work.  Added to review.

clusterMembership/dynamicCluster.go
CaitieM20: yes will fix in a separate pull request, also want to make Members() not make the copy but instead store a list which is updated on add or remove since these should happen less frequently

PR #9, opened 2016-05-26
Add Scheduler Job Message & Refactor

CaitieM20:
- Add Scheduler Job Message Struct. This will be used by Scoot Scheduler and Scoot Api
- Refactor ClusterMembership and Work Interfaces to support accepting a JogMessage instead of a string
- Moved Demo main.go into demo folder, deleted test cases because refactoring would be time consuming, and this is soon to be replaced by real scoot implementation
- Refactored TestNode into ClusterMembership package, generally useful should be provided as part of the base package.  
- Remove Codecoverage.txt should not be checked in


dbentley:
LGTM

PR #10, opened 2016-06-01
Sagas

CaitieM20:
Implementing [Sagas](https://www.youtube.com/watch?v=xDuwrtwYHu8).  Sagas are how we are going to ensure all Tasks in a Job are completed.  

### Code
- Saga Struct & Methods give users an interface to program against, backed by a SagaLog
- SagaLog Interface allows us to implement the durable backend.  
- SagaState tracks SagaState.  Ensures messages added to a Saga are valid.  Currently not thread safe because there should be only One Go Routine updating an individual Saga at a time, this gives us thread safety and message ordering guarantees for free!
-InMemorySagaLog, A non-durable Implementation of a SagaLog for demo & unit testing code.  

### Tests
- various test cases
- introduced [gomocks](https://github.com/golang/mock) into the codebase for testing the Interface.  Checking in code generated code for now will add to build process in subsequent pull request

### Demo
- updated Demo code to use Sagas. 
- InMemory Saga prints received messages to the command line
- run demo: go run sched/demo/main.go


saga/inMemorySagaLog.go
```diff
@@ -0,0 +1,72 @@
...
+import "fmt"
+import "sync"
+
+/*
+ * In Memory Implementation of a Saga Log, DOES NOT durably persist messages.
+ * This is for local testing purposes.
+ */
+type inMemorySagaLog struct {
+	sagas map[string]*SagaState
+	mutex *sync.RWMutex
```
jasonbs10: I don't believe we want locking in this testing/example code. GetSagaState() gives the caller an unlocked pointer to state which LogMessage() can then alter concurrently. The code that operates on Saga structs should lock as necessary instead (assuming we continue to pass by pointer).

saga/sagaState.go
```diff
@@ -0,0 +1,248 @@
...
+func (state *SagaState) IsSagaCompleted() bool {
+	return state.sagaCompleted
+}
+
+/*
+ * Updates a SagaState with the specified SagaMessage.
+ * Returns an InvalidSagaState Error if applying the message would result in an invalid Saga State
+ * Returns an InvalidSagaMessage Error if the message is Invalid
+ */
+func (state *SagaState) UpdateSagaState(msg sagaMessage) error {
```
jasonbs10: I don't know, can do without the saga messages and factories? Are these messages truly opaque or do both SagaState and a real SagaLog impl have to look inside? If it's not opaque then I think it's cleaner to split UpdateSagaState() and LogMessage() into functions corresponding to each SagaMessage type.

jasonbs10:
I think this is designed well in terms of minimizing testing verbosity, but is it accommodating testing too much? Otherwise, lgtm, it's easy enough to read and understand it as written

saga/inMemorySagaLog.go
CaitieM20: Yes agree.  Actually just talked to Dan about a interface change that will make exposing SagaState Unnecessary.

Also inMemorySaga has locks just because you can concurrently run multiple sagas that are updating the same in memory dictionary concurrently.

saga/sagaState.go
CaitieM20: So if Go had Types I'd just make sub types of an abstract class SagaMessage where each base class only had the properties it used.  Seems to be Go convention to hide structs inside a package and expose Factory methods that initialize the structs correctly.  

I want the message structure to be opaque, because these are being constructed by implementations of SagaLog which won't be in this package, etc...

saga/inMemorySagaLog.go
jasonbs10: I understand why it's locking, but it's not sufficient. I'll wait to see Dan's suggested change, but what I was getting at is all locking should be moved to scheduleWork().

saga/sagaState.go
jasonbs10: What do we gain by having SagaMessages at all though, vs passing the correct args directly to a set of functions? I can imagine cases where having them is necessary but their usefulness is not immediately clear from this initial checkin.

CaitieM20:
Refactored Saga methods to take in a SagaState and return a SagaState as discussed with Dan.
This is more functional, easier to test etc... Still has a side-effect of durably persisting the data to the SagaLog though.  SagaStates cannot be created outside the Saga package without calling Saga.StartSaga(...)

The next TODO piece after this implementation is the Recovery part.  Reading the Durable SagaLog to reconstruct a list of all in progress sagas and their state, and then hand it back to the caller to reschedule as necessary.  

This also probably needs a pass tomorrow to re-evaluate if some of the SagaMessage functions still need to be exposed outside the package.  They probably do not.  SagaMessages should also have a default serialize/deserialize method provided for SagaLog Implementations to use to persist the data.

jasonbs10:
I thought the message factories were perhaps not needed and I wanted to know if you took that direction because it would facilitate testing, or because you tried a couple variations and this way was best, etc. I'm simply looking to reduce the lines of code where possible if it doesn't hurt readability.

saga/inMemorySagaLog.go
```diff
@@ -0,0 +1,71 @@
+package saga
+
+import "errors"
```
dbentley: We've been preferring import (
  "foo"
  "bar"
)

WDYT?

saga/inMemorySagaLog.go
```diff
@@ -0,0 +1,71 @@
...
+ */
+type inMemorySagaLog struct {
+	sagas map[string][]sagaMessage
+	mutex *sync.RWMutex
+}
+
+/*
+ * Returns an Instance of a Saga based on an InMemorySagaLog
+ */
+func InMemorySagaFactory() Saga {
```
dbentley: In my desire to avoid writing Java in go, I've been steering away from Factory. WWYT of "MakeInMemorySaga"?

saga/saga.go
```diff
@@ -0,0 +1,118 @@
...
+	BackwardRecovery SagaRecoveryType = iota
+	ForwardRecovery
+)
+
+/*
+ * Saga Object which provides all Saga Functionality
+ * Implementations of SagaLog should provide a factory method
+ * which returns a saga based on its implementation.
+ */
+type Saga struct {
```
dbentley: Hrm, I'm not sure this will work for implementations outside the package, because they won't be able to set log...

saga/saga.go
```diff
@@ -0,0 +1,118 @@
...
+/*
+ * Log an EndTask Message to the log.  Indicates that this task
+ * has been successfully completed. Returns an error if it fails.
+ */
+func (s Saga) EndTask(state *SagaState, taskId string, results []byte) (*SagaState, error) {
+	return s.logMessage(state, EndTaskMessageFactory(state.sagaId, taskId, results))
+}
+
+/*
+ * Log a Start a Compensating Task if Saga is aborted, and rollback
```
dbentley: English (and capitalization) here.

saga/sagaState.go
```diff
@@ -0,0 +1,292 @@
...
+import "fmt"
+
+/*
+ * Data Structure representation of the current state of the Saga.
+ */
+type SagaState struct {
+	sagaId string
+	job    []byte
+
+	taskStarted   map[string]bool
```
dbentley: What are the keys and values? How do we know that tasks have unique names?

saga/sagaState.go
```diff
@@ -0,0 +1,292 @@
...
+	taskCompleted map[string]bool
+	taskResults   map[string][]byte
+
+	compTaskStarted   map[string]bool
+	compTaskCompleted map[string]bool
+	compTaskResults   map[string][]byte
+
+	sagaAborted   bool
+	sagaCompleted bool
+	version       int
```
dbentley: What is version? Do we use it yet?

saga/sagaState.go
```diff
@@ -0,0 +1,292 @@
...
+func (state *SagaState) Job() []byte {
+	return state.job
+}
+
+/*
+ * Returns true if the specified Task has been started,
+ * fasle otherwise
+ */
+func (state *SagaState) IsTaskStarted(taskId string) bool {
+	started, ok := state.taskStarted[taskId]
```
dbentley: I think you can just return state.taskStarted[taskId]

saga/sagaState.go
```diff
@@ -0,0 +1,292 @@
...
+
+	case EndSaga:
+
+		//A Successfully Completed Saga must have StartTask/EndTask pairs for all messages or
+		//an aborted Saga must have StartTask/StartCompTask/EndCompTask pairs for all messages
+		for taskId := range state.taskStarted {
+
+			if state.sagaAborted {
+				if !(state.compTaskStarted[taskId] && state.compTaskCompleted[taskId]) {
+					return nil, errors.New(fmt.Sprintf("InvalidSagaState: End Saga Message cannot be applied to an aborted Saga where Task %s has not completed its compensating Tasks", taskId))
```
dbentley: fmt has a method Errorf to create an error with a formatted string.

saga/sagaState.go
```diff
@@ -0,0 +1,292 @@
...
+ * Creates a deep copy of mutable saga state.  Does not Deepcopy
+ * Job field since this is never mutated after creation
+ */
+func copySagaState(s *SagaState) *SagaState {
+
+	newS := &SagaState{
+		sagaId:        s.sagaId,
+		sagaAborted:   s.sagaAborted,
+		sagaCompleted: s.sagaCompleted,
+		version:       s.version + 1,
```
dbentley: Incrementing the version is not a copy. I'd make this just a copy, and then update the version in the apply function above.

saga/saga_test.go
```diff
@@ -0,0 +1,331 @@
...
+	state, err = s.EndCompensatingTask(state, "task1", nil)
+	if err != nil {
+		t.Error(fmt.Sprintf("Expected EndCompensatingTask to not return an error"))
+	}
+}
+
+func TestEndCompTaskLogError(t *testing.T) {
+	entry := EndCompTaskMessageFactory("testSaga", "task1", nil)
+
+	mockCtrl := gomock.NewController(t)
```
dbentley: I think this would be clearer and less brittle with a fake instead of a mock. A fake saga log would let us get the messages at the end. Then the test would run and then verify that the contents of the log were what we expected.

saga/inMemorySagaLog.go
jasonbs10: I have gofmt set up to run on every save. Seems like a good thing to have if your editor supports it.

saga/inMemorySagaLog.go
jasonbs10: The golang/go repo uses Make* and New*, I'm not sure when to use one or the other. We haven't really been consistent in the existing code.

saga/inMemorySagaLog.go
CaitieM20: Can do!

I have gofmt set up on save to and as part of our build!  Dan was saying for imports it mimics however the file was set up.  If we want enforce this overall we may want to set up a linter, or just do it via code reviews.

saga/inMemorySagaLog.go
CaitieM20: I'll change to Make for now and then if we want to align on New we can do a refactor!

saga/saga.go
CaitieM20: yea totally right duh, fixed with a MakeSaga function that should be used

saga/saga.go
CaitieM20: fixed

saga/sagaState.go
CaitieM20: added comments to each map.  The key is always the taskId.
  
TaskId has to be unique for Sagas to work just like SagaId must be unique for each saga.  In scoot, we will have unique TaskIds as part of the API to schedule work.

saga/sagaState.go
CaitieM20: yup totally right, fixed

saga/sagaState.go
CaitieM20: cool, that's much nicer, updated!

saga/sagaState.go
CaitieM20: removed version, not using it anymore, since Saga is all logic and just stores state.

saga/saga_test.go
CaitieM20: I disagree, then we are just testing that the fake works correctly (and doing a lot more work writing fakes) inMemorySaga and the demo already does this, but all it proves is that my implementation of inMemorySaga is correct.  Our implementation if DistributedLogSaga could still be wrong.  

Mocks are useful for unit tests to ensure you are passing the right things to the interface, to actually test that the interface implementation is correct & is persisted correctly we need an integration test.

dbentley: OK, I'll take this offline and see if I can convince you later, but we don't need to block on this.

CaitieM20:
@jasonbs10 back to the MakeMessage Methods (MessageFactoryMethods) I created them because I want to enforce stronger guarantees of what's in a message, what fields are set etc... than the Go Type System allows me to use.  If we were in a Strongly typed Language there would be

- abstract class SagaMessage
- case class StartSagaMessage extends SagaMessage
- case class EndSagaMessage extends SagaMessage
- abstract TaskMessage extends SagaMessage
- case class StartTaskMessage extends TaskMessage 
etc ...

While the code is more verbose, its more maintainable, if we choose to add data to a message, the compiler will enforce that all parameters necessary to initialize this object are provided.  If we just initialize based on the struct we are liable to miss updating things, since its not compile time enforced.

dbentley:
(Jumping in, so apologies if I'm missing context)

Will the factory methods ever be called from another package? It seems like in the current setup, they won't, and so we're trying to stop ourselves from doing something wrong.

I like it if it stops clients from doing the wrong thing, but I don't think it will. (sagaMessage isn't even exported, so they can't create a variable of that type)

CaitieM20:
I assume they will be called when users are testing their SagaLog implementations which will be in other packages.  The interface of LogMessage uses sagaMessages, which cannot be created directly.  

Maybe this is just for testing, but I think its more readable code and a nice API for low overhead easy to maintain code.

dbentley:
Ah, I see. Yeah, in memory saga log is currently privileged because it lives in this package.

I think this is something where we need more experience and shouldn't bikeshed now.

I'm happy to trust your sense, but I want to make sure we don't ignore concerns and end up with too much code (which is my understanding of Jason's concerns).

I don't know what the best thing to do is. Maybe revisit this at some point to see what it's buying us as we have our second implementation of log?

CaitieM20:
Yes happy to have this evolve if it becomes burdensome.  @dbentley gave a verbal Ok to Check In @jasonbs10 anything outstanding before submitting?

jasonbs10:
Nope, nothing outstanding, feel free to merge.

PR #12, opened 2016-06-06
Scoot Local Server skeleton.

dbentley:
localserver will serve against a Unix Domain Socket (currently $HOME/.scoot/socket)

running scootcl echo <args> will echo against a local server

Currently uses grpc over Unix Domain Sockets. We may want to investigate this, because right now it appears that the RTT is ~600 usecs. Possible suspects: unix domain sockets may not be properly buffered. grpc's http2 transport may be too high overhead for these kidns of RPCs.

NB (that took me hours to figure out): SOCK_SEQPACKET for unix domain sockets is supported on Linux but not OS X.

Also, create a directory for binaries. These binaries should have very little code, deferring to libraries. It should exist to do dependency injection and that's it.


binaries/localserver/main.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/scootdev/scoot/local/protocol"
+	"github.com/scootdev/scoot/local/server"
+	"log"
+)
+
+// A Local Scoot server.
+func main() {
+	scootdir, err := protocol.Locate()
+	if err != nil {
+		log.Fatal("Error locating Scoot instance: ", err)
```
CaitieM20: Probably should return an exit code signifying an error happened

binaries/scootcl/main.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"github.com/scootdev/scoot/local/client/cli"
+	"github.com/scootdev/scoot/local/client/conn"
+	"log"
+	"os"
+)
+
+// A Scoot command-line client
+func main() {
+	dialer, err := conn.UnixDialer()
```
CaitieM20: Same here, if != nil err needs to return an exit code otherwise nil pointer exceptions or it'll use garbage.

local/client/cli/cli.go
```diff
@@ -0,0 +1,45 @@
...
+		c.conn = conn
+	}
+	return c.conn, nil
+}
+
+func NewCliClient(dialer conn.Dialer) (*CliClient, error) {
+	rootCmd := &cobra.Command{
+		Use:   "scootcl",
+		Short: "Scootcl is a command-line client to Scoot",
+		// Run: func(cmd *cobra.Command, args []string) {
```
CaitieM20: Commented out intentionally?

local/client/conn/conn.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"github.com/scootdev/scoot/local/protocol"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+	"log"
+	"net"
+	"time"
+)
+
+// A Dialer can dial a connection to the Scoot server.
```
CaitieM20: nit: how do we want to do block comments style wise.  I've been doing /* comment */  Looks like standard Go uses //

testCoverage.sh
```diff
@@ -3,12 +3,10 @@
...
 echo "" > coverage.txt
 
-for d in $(find ./* -maxdepth 10 -type d); do
-    if ls $d/*.go &> /dev/null; then
-        go test -coverprofile=profile.out -covermode=atomic $d
-        if [ -f profile.out ]; then
-            cat profile.out >> coverage.txt
-            rm profile.out
-        fi
+for d in $(find . -name \*.go | xargs -n 1 dirname | uniq); do
```
CaitieM20: yay thanks!

binaries/localserver/main.go
dbentley: Yeah, log.Fatal logs and then calls os.Exit.

https://golang.org/pkg/log/#Fatal

binaries/scootcl/main.go
dbentley: same

local/client/cli/cli.go
dbentley: I commented it out intentionally, but should have removed it entirely. Done.

local/client/conn/conn.go
dbentley: I'm fine with whatever style we pick. I'd even lean towards "it's not worth enforcing a consistent style for this", but I'm happy to pick one and stick with it. WDYT?

binaries/localserver/main.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"github.com/scootdev/scoot/local/protocol"
+	"github.com/scootdev/scoot/local/server"
+	"log"
+)
+
+// A Local Scoot server.
+func main() {
+	scootdir, err := protocol.Locate()
```
jasonbs10: Locate() is a bit vague, maybe LocateScootDir() or something similar?

binaries/localserver/main.go
jasonbs10: I had the same comment in an earlier review. It's not entirely intuitive that calling log exits the program.

local/client/cli/cli.go
```diff
@@ -0,0 +1,45 @@
...
+		conn, err := c.dialer.Dial()
+		if err != nil {
+			return nil, err
+		}
+		c.conn = conn
+	}
+	return c.conn, nil
+}
+
+func NewCliClient(dialer conn.Dialer) (*CliClient, error) {
```
jasonbs10: I'd want dialer to be optional when making a client. This separation is useful for testing but I'm not a big fan of having to construct trivial defaults when the lib can just do that for me. Also applies to NewServer().

binaries/localserver/main.go
```diff
@@ -0,0 +1,23 @@
...
+	if err != nil {
+		log.Fatal("Error locating Scoot instance: ", err)
+	}
+	s, err := server.NewServer()
```
jasonbs10: See NewCliClient() comment.

local/protocol/local.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
+
+// TODO(dbentley): integrate this into our go build
```
jasonbs10: Any particular reason for choosing this over thrift? Is it easier to work with when making something simple?

local/client/cli/cli.go
```diff
@@ -0,0 +1,45 @@
...
+	dialer  conn.Dialer
+	conn    conn.Conn
+}
+
+func (c *CliClient) Exec() error {
+	return c.rootCmd.Execute()
+}
+
+func (c *CliClient) Close() error {
+	return c.conn.Close()
```
jasonbs10: Should this set conn to nil as well?

local/client/conn/conn.go
jasonbs10: I agree. "//" by default, no big deal if someone forgets and uses "/**/". I imagine that godoc, if we use that, supports either.

local/client/conn/conn.go
```diff
@@ -0,0 +1,66 @@
...
+	socketPath string
+}
+
+func dial(addr string, timeout time.Duration) (net.Conn, error) {
+	return net.DialTimeout("unix", addr, timeout)
+}
+
+func (d *dialer) Dial() (Conn, error) {
+	log.Println("Dialing ", d.socketPath)
+	c, err := grpc.Dial(d.socketPath, grpc.WithInsecure(), grpc.WithDialer(dial))
```
jasonbs10: Not a big fan of this type of wrapper. This is idiomatic for grpc I assume?

local/protocol/locate.go
```diff
@@ -0,0 +1,34 @@
+package protocol
+
+import (
+	"fmt"
+	"os"
+	"path"
+)
+
+// Utilities to deal with Scoot instances.
```
jasonbs10: It's a little weird that this lives in the same place as the grpc protocol. What's the reasoning here?

jasonbs10:
Looks fine for a first cut. We should talk offline about when to use interfaces vs structs and how isolated components should be (versatility vs simplicity).

For one-shot connections, which is how I understand the cli to work, is SOCK_SEQPACKET useful?

local/client/cli/cli.go
dbentley: Done.

local/client/cli/cli.go
dbentley: This is Dependency Injection. I remember you said you thought sometimes the SOLID principles led to convoluted code.

I think it leads to a nice separation of concerns. The CLI client knows how to translate between command-line and the connection layer; it doesn't know how to create a connection.

Should we talk about how we feel about DI as a team?

local/client/conn/conn.go
dbentley: Yup.

What don't you like about it? What type of wrapper do you see it as?

local/protocol/local.proto
dbentley: I was sure it worked with Unix Domain Sockets, and Thrift wasn't clear on if it could.

Also, the previous bugs in the thrift implementation for Go made me feel uneasy.

I'm not convinced it's a perfect choice, which is why it's hidden and doesn't escape.

local/protocol/locate.go
dbentley: I figured agreeing on how to determine the socket was part of the protocol.

Put another way: the protocol is what's the server and client have to agree on, which includes this code.

binaries/localserver/main.go
dbentley: Done.

local/client/conn/conn.go
jasonbs10: It's just another couple levels of indirection vs having Dial() incorporate dial() directly (not advocating for that here). I'm not sure how this code would look with thrift though, maybe similar.

dbentley:
SOCK_SEQPACKET is useful even for one shot connections because it lets you know when a packet is over. But it doesn't matter for Thrift or GRPC because they're built to work on top of SOCK_STREAM.

local/client/conn/conn.go
dbentley: Oh, I see. I could move func dial to be an anonymous func inside Dial(). I tend to prefer explicit small func's and prefer to leave anonymous inner func's for when they're operating as a closure and capturing state.

WDYT?

local/protocol/locate.go
jasonbs10: In the broader sense of what protocol can entail, yes it fits. However there is the 'wire protocol' and then there is all the other stuff usually passed as flags or env vars, which I typically categorize as 'config'. I think it's ok to leave it this way though.

local/client/conn/conn.go
CaitieM20: Cool I've just been doing /* */ but that's a Java / Scala habit happy to change to whatever is more standard or works best with godoc

CaitieM20:
LGTM for a first cut, at some point when we lock down on the design I'd like to see test cases as well.  

In regards to Thrift vs GRPC, mixing protocols in a project is going to lead to a lot of complexity and confusion.  We should have a discussion about picking one and sticking with it, as this topic will keep coming up

PR #13, opened 2016-06-06
Caitie/makefile

CaitieM20:
- Adding a super basic Makefile
- Does Codegen automatically as part of test process
- Update Travis CI to run off of codegen


dbentley:
Did you look into using go generate? https://blog.golang.org/generate

Is it worth adding this to .gitignore, or do we want it checked in?

CaitieM20:
Did not know about go generate, really nice, added! 
Also yes we want to .gitignore all *_mock.go files added!

dbentley:
LGTM

PR #14, opened 2016-06-06
Compressing sagastate

CaitieM20:
Since we are copying SagaState on every state transition, compressing state to BitFlags instead of individual maps.

### DIFF
--- a/saga/sagaState.go
+++ b/saga/sagaState.go
@@ -5,6 +5,15 @@ import (
 	"fmt"
 )
 
+type flag byte
+
+const (
+	TaskStarted flag = 1 << iota
+	TaskCompleted
+	CompTaskStarted
+	CompTaskCompleted
+)
+
 /*
  * Data Structure representation of the current state of the Saga.
  */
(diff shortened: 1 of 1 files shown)


saga/sagaState.go
```diff
@@ -5,28 +5,28 @@ import (
 	"fmt"
 )
 
+type flag byte
```
dbentley: Do these need to be and'ed together? It seems like having just one of the states below would be enough to know what to do? Or am I missing something?

CaitieM20: (?) ANDing bitflags will always produce zero.

This is a list of all the messages received for this task(START, END, STARTCOMP, ENDCOMP)
they don't always all happen in a Saga.  i.e. if there is an END there is always a START, but if there is a STARTCOMP there isn't always an END message.  See the updateFunction for full rules.

dbentley: Sorry, I meant 'or', not 'and'.

If you see an ENDCOMP, you know you've seen a STARTCOMP. If you see an END, you know you've seen a START.

So I guess I'm saying I thought you could summarize the state of a Task with one state, instead of all of the messages you've seen. Am I misunderstanding?

CaitieM20: SortOf...  You may not see an End message if CompMessages are supplied.

These are all valid combos of messages
(Start)
(Start, End)
(Start, StartComp)
(Start, End, StartComp)
(Start, StartComp, EndComp)
(Start, End, StartComp, EndComp)

Given that I can't get the flags to be smaller than 8 bits (I'm only using 4), I think its worth the clarity of having a flag for each received message.

dbentley:
LGTM

FWIW, if you don't need to differentiate the states created by the sequences:
(Start, StartComp)
(Start, End, StartComp)

then I think you can just store the current state and get rid of the history, but that may be more change than you want to make right now (or maybe you do need to differentiate those states)

CaitieM20:
You _could_ infer data based on which messages you received so we could compress the state to 2 bits maybe(?) But we have to have 8 because of Go's types & we are only using 4.  This makes the code more complex so what's the benefit?

For instance the IsTaskStarted() method would now need to understand to check is the StartTask bit flag set or the EndTask bit flag set.  Instead of just testing if the StartTask bit is set.  Reading this code would also be confusing.  Instead of all the Saga State Transition logic being encapsulated in the updateSagaState method it's not littered all over the file.  I would prefer to never do this.

PR #15, opened 2016-06-06
Sagas: Store Data When Logging Task Messages

CaitieM20:
Adding the ability to store opaque data with Task Messages (StartTask, EndTask, StartCompTask, EndCompTask) this is useful for debugging / diagnostics.  Its also useful for storing intermediary results which can be rolled up when a Saga has Completed, or generating reports after the fact from the SagaLog. 

Exposing this data to the applications via the SagaState API, and can be added via Saga API for logging messages.


saga/sagaState.go
```diff
@@ -6,30 +6,40 @@ import (
...
+/*
  * Data Structure representation of the current state of the Saga.
  */
 type SagaState struct {
 	sagaId string
 	job    []byte
 
+	// map of taskID to task data supplied when committing
+	// startTask, endTask, startCompTask, endCompTask messages
+	taskData map[string]*taskData
```
dbentley: So if I commit endTask, it overwrites the data I set as part of startTask?

saga/sagaState.go
```diff
@@ -6,30 +6,40 @@ import (
 )
 
 /*
+ * Additional data about tasks to be committed to the log
+ * This is Opaque to SagaState, but useful data to persist for
+ * results or debugging
+ */
+type taskData struct {
```
dbentley: There are a lot of maps from taskID to some kind of data. Would it be worth making a taskState struct that stores all of them (and is also immutable)?

CaitieM20: This [pull requests] (https://github.com/scootdev/scoot/pull/14) reduces most of them.  So with both merged there will be 
* map of taskId -> flag
* map of taskId -> taskData

In theory yes we could combine the two next if we want which will save an extra string copy per task.    

I don't understand the immutable comment?  Go doesn't provide Immutable constructs.  This code provides SagaState Immutability based on the API (only read methods outside of package) and based on the copyMethod, which is updated to treat taskData as immutable by copying the (ptr, len, cap) slice object when copying a SagaState. 

There's only so much I can do to prevent a user from mutating the underlying []bytes since we give them the slice reference back with a pointer to the underlying slice.

dbentley: I'm not concerned about an extra string copy per task; I just thought it might be nice to put all the task state and data together, so it's:
map of taskId -> taskState

Then per-task methods could just take a taskState, instead of both a flag and taskData. Would that be useful or not?

The immutable comment just meant doing what you're doing, not any kind of hardening (which, yes, Go doesn't provide).

saga/sagaState.go
```diff
@@ -79,6 +88,19 @@ func (state *SagaState) IsTaskStarted(taskId string) bool {
 }
 
 /*
+ * Get Data Associated with Start Task, supplied as
+ * Part of the StartTask Message
+ */
+func (state *SagaState) GetStartTaskData(taskId string) []byte {
+	data, ok := state.taskData[taskId]
```
jasonbs10: You can return data directly right? The comment can indicate that it might be nil.

saga/sagaState.go
CaitieM20: If we combine them, it needs to happen in a later code review, since the taskState has been compressed in a separate pull request. 

Also none of this is exposed as part of the public API so less concerned with how we internally represent task state, It can continue to evolve to match our perf / clarity needs, feel free to add a Jira ticket if this is something you feel strongly about.  More concerned with getting it up and running end to end with a cohesive public API

dbentley:
LGTM

PR #16, opened 2016-06-07
Caitie/rename factory

CaitieM20:
rename sagaStateFactory to makeSagaState to be consistent

I missed sagaStateFactory when doing the rename from factory to make in a previous pull request


dbentley:
LGTM

PR #17, opened 2016-06-07
Scoot LocalServer: support running and querying for status.

dbentley:
Add Runner interface, and a mock implementation (in which every command succeeds immediately)

Wire this through from client to server and over rpc.


local/protocol/local.proto
```diff
@@ -22,4 +25,37 @@ message EchoRequest {
...
+  UNKNOWN = 0;
+  PENDING = 1;
+  RUNNING = 2;
+
+  COMPLETED = 3;
+  FAILED = 4;
+}
+
+message Duration {
+  int64 seconds = 1;
```
jasonbs10: Can you remind me, what's the reasoning behind splitting sec/nsec?

runner/runner.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// Only valid if State == COMPLETED
+	ExitCode int
+
+	// Only valid if State == FAILED
+	Error string
+}
+
+type Runner interface {
+	Run(cmd *Command) (*ProcessStatus, error)
```
jasonbs10: What is the expected return value here? Should it be anything other than pending?

local/server/server.go
```diff
@@ -34,3 +37,34 @@ func Serve(s *Server, scootdir string) error {
...
+func (s *Server) Status(ctx context.Context, req *protocol.StatusQuery) (*protocol.ProcessStatus, error) {
+	status, err := s.runner.Status(runner.RunId(req.RunId))
+	if err != nil {
+		return nil, err
+	}
+
+	return runnerToProtocolStatus(status), nil
+}
+
+func runnerToProtocolStatus(status *runner.ProcessStatus) *protocol.ProcessStatus {
```
jasonbs10: I'm not sure this belongs here. We don't want protocol to leak into the runner code, so it can't go there, but maybe somewhere in the protocol dir?

local/client/conn/conn.go
```diff
@@ -61,6 +68,40 @@ func (c *conn) Echo(arg string) (string, error) {
...
+
+func (c *conn) Status(run runner.RunId) (*runner.ProcessStatus, error) {
+	r, err := c.client.Status(context.Background(), &protocol.StatusQuery{string(run)})
+	if err != nil {
+		return nil, err
+	}
+	return protocolToRunnerStatus(r), nil
+}
+
+func protocolToRunnerStatus(status *protocol.ProcessStatus) *runner.ProcessStatus {
```
jasonbs10: See comment on runnerToProtocol

local/protocol/local.proto
dbentley: Ummm, it's what Google does? https://github.com/google/protobuf/blob/master/src/google/protobuf/duration.proto

But I think you're right, it's silly. I'll remove it.

runner/runner.go
dbentley: Sure:
1) you get the RunID
2) it could have capacity and have already started running it
3) it could have failed it, because, e.g., you gave it a nonsensical Argv
4) there are other reasons it could fail the request before even returning

local/client/cli/run_cmd.go
```diff
@@ -0,0 +1,30 @@
+package cli
+
+import (
+	"github.com/scootdev/scoot/runner"
+	"github.com/spf13/cobra"
+	"log"
+)
+
+func makeRunCmd(cli *CliClient) *cobra.Command {
```
jasonbs10: We have shadowing throughout our code, it's confusing. Can you rename?

local/client/conn/conn.go
```diff
@@ -19,6 +20,12 @@ type Dialer interface {
 
 type Conn interface {
 	Echo(arg string) (string, error)
+
+	// TODO(dbentley): this feels weird. We shouldn't expose our internal
```
jasonbs10: If runner is shared by client/server, can it go somewhere under protocol?

dbentley: Yeah, this why I feel uneasy about putting this here.

Runner is a useful concept without regard to the local server. It should also be used the Worker binary to run, and the Worker isn't involved in the local server protocol at all.

jasonbs10: As long as it's isolated so you can pull in exactly the dependencies you need, I don't see a huge problem with libs referring to higher level dirs.

local/server/server.go
dbentley: Done.

local/client/conn/conn.go
dbentley: Done.

runner/runner.go
jasonbs10: In general, for questions that come up in reviews, it makes sense to document the answer in the code.

local/client/cli/run_cmd.go
dbentley: Changed to c; wdyt? Is there something better?

jasonbs10:
Kinda lgtm, but tbh I can barely fit the DI into my head. So this is already pushing my comfort level with having everything be substitutable.

local/client/cli/run_cmd.go
jasonbs10: Tolerable for a method this short. I might have gone with cliC.

local/protocol/local.proto
dbentley: Done.

PR #18, opened 2016-06-08
Add Build Dependencies to MakeFile

CaitieM20:
Adding dependencies command to make file.  Using this command to install all dependencies needed for TravisCI build.

The goal is to :
- make it it easy to setup dev environment
- make it easy to sync dependencies when new ones are added
- ensure our build process & dev environment are similar

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ go:
   - 1.6
 
 before_install:
-  - go get -t $(go list ./...)
+  - make dependencies
   
 before_script:
   - go fmt ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,7 @@ go:
   - 1.6
 
 before_install:
-  - go get github.com/golang/mock/gomock
-  - go get github.com/golang/mock/mockgen
-  - go get github.com/onsi/gomega
-  - go get github.com/onsi/ginkgo
-  - go get golang.org/x/tools/cmd/cover
+  - make dependencies
   
 before_script:
   - go fmt ./...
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -6,11 +6,7 @@ go:
   - 1.6
 
 before_install:
-  - go get github.com/golang/mock/gomock
-  - go get github.com/golang/mock/mockgen
-  - go get github.com/onsi/gomega
-  - go get github.com/onsi/ginkgo
-  - go get golang.org/x/tools/cmd/cover
+  - make dependencies
```
dbentley: Go get should just download dependencies; can you just make this line go get -t scootdev/scoot/... ?

CaitieM20: so I tried go get -t $(go list./...)
and apparently that does not work correctly as input like go vet $(go list ./...) or go test $(go list ./...)
https://travis-ci.org/scootdev/scoot/builds/136274042

So we can manually list every project or ever dependency what are people's preferences

dbentley: Arg. that really ought to work...

Did you try go get -t scootdev/scoot/... ? go get is supposed to get dependencies so that should be enough...

Oh wait! Maybe it's because mockgen isn't actually listed as a dependency? What if you add:

import (
  _ "path/to/mockgen"
)
to sagalog.go (so that go install will install it)?

dbentley: ARG, ok, I hit this too.

I think it's that the test command in Makefile should include a go install github.com/golang/mock/mockgen

I haven't been able to figure out how it currently works.

CaitieM20: yay that worked!!! add it manually as part of dependencies with a comment

dbentley:
LGTM

PR #20, opened 2016-06-09
First take Proc Api

CaitieM20:


proc/api.thrift
```diff
@@ -0,0 +1,40 @@
+service Proc {
+  
+  exception InvalidRequest {
+    1: optional string message
+  }
+
+  exception ScootTooBusy {
```
dbentley: Maybe "CanNotSchedule"? E.g., maybe Scoot isn't that busy, but it wants to rate limit just you?

proc/api.thrift
```diff
@@ -0,0 +1,40 @@
...
+    1: optional i64 retryAfterMs
+  }
+  
+  enum JobType {
+    IRON_TESTS=1,
+    DEFAULT=2,
```
dbentley: Maybe UNKNOWN=1, IRON_TESTS=2, and we can add anything after that?

proc/api.thrift
```diff
@@ -0,0 +1,40 @@
...
+    4: optional string contextId
+  }
+
+  struct Job {
+    1: required string id,
+    2: required list<Task> tasks
+  }
+
+  Job RunJob(
+    1: optional string contextId,
```
dbentley: I'd suggest deleting this, and just having contextId on each Task.

(Also, I'd put the meat first, which in this case I think is tasks)

proc/api.thrift
```diff
@@ -0,0 +1,40 @@
...
+    2: optional JobType jobType,
+    3: required list<Task> tasks
```
dbentley: At some points I've heard that required is really hard to ever upgrade, and so a large school of thought is to make everything optional; wdyt?

dbentley:
Have some nits, but generally looks good.

proc/api.thrift
jasonbs10: When do we start enum at 0 for unknown vs 1?

proc/api.thrift
```diff
@@ -0,0 +1,40 @@
...
+    1: optional i64 retryAfterMs
+  }
+  
+  enum JobType {
+    IRON_TESTS=1,
+    DEFAULT=2,
+  }
+
+  struct Command {
+    1: list<string> cmds
```
jasonbs10: I remember Dan making a comment about command strings and the difficulty of shell parsing. Is that applicable at all here?

proc/api.thrift
CaitieM20: Required is hard to upgrade, changing it is a breaking change, but having no required fields is a little weird, it suggests your api is endlessly malleable, and can make progress without any real input.  

I chose to make the things that our API will not function without required.  JobId is core to the ScootAPI if we didn't send you one back then you could never check the status of your event.  Likewise a Job without tasks doesn't make sense, a Job needs to specify work to do.  If any of these things changed I would not modify the existing API I would just make a new one since likely the semantics of the API have drastically changed.  

Does that make sense?  Also it means less error handling boiler plate code for us to write, to check for things we always expect to be there

proc/api.thrift
CaitieM20: Yes that's why its a list of strings and not just a single string.

proc/api.thrift
CaitieM20: So my thoughts were:
- Job contextId is likely to be the build you are running against, this is more for future multi Job scheduling and taking advantage of Data Locality.  I agree that we know all the tasks will benefit from the same Data Locality. 
- Task contextId is additional info like target affinity in iron tests which will help us schedule more effectively at the per task level.  I.e. I'm using 10 machines to run this Job which have github commit data locality, and running task A&B on the same machine will give me added benefits since they build tests running similar targets.

proc/api.thrift
CaitieM20: Sure changed it to CanNotScheduleNow, I want the error name to convey that this error is related to time and not request.  A 500 vs a 400 level error in HTTP

proc/api.thrift
jasonbs10: Well yes, but what does that list represent? It says cmds, plural, so it must split things at the command level, not at the arg level. I'm floating the idea that it should be list<list<string>>.

proc/api.thrift
CaitieM20: So I set it to 1 because sometimes its difficult to tell between no value, (defaults to 0) and specifying default (also 0) since this is optional.  

Maybe it doesn't really matter in this case but typically I don't start IDL enums at 0

proc/api.thrift
dbentley: Maybe call it args, or argv?

proc/api.thrift
dbentley: SGTM

proc/api.thrift
jasonbs10: That makes sense, yet there is a reason Google proto3 removed 'required'/'optional', because they were more problematic than useful. I'm not sure what else went with that proto change though, so the same arguments may not apply directly to thrift.

dbentley: OK.

proc/api.thrift
dbentley: Yeah, that makes sense. I feel like this is a feature we know we'll need, but we're not 100% clear on the best way to do it.

I'd suggest leaving this until we're ready to start passing meaningful values here, and then it should be much clearer what API we want.

dbentley:
LGTM

jasonbs10:
I'm not receiving notifications for commits so I'm not sure I'm following along correctly. If you address cmds vs args field then lgtm.

CaitieM20:
@jasonbs10 yes Args and Cmds are the same I believe.  @dbentley maybe able to elaborate more?  This came out or our convo better

PR #21, opened 2016-06-09
Tests for CLI client (and fixing a few bugs found)

dbentley:


local/client/cli/echo_cmd.go
```diff
@@ -25,6 +24,6 @@ func (c *CliClient) echo(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	log.Println("Scoot server: ", result)
+	fmt.Println(result)
```
jasonbs10: Are we preferring log/fmt in different contexts?

local/client/cli/cli.go
```diff
@@ -33,12 +36,17 @@ func (c *CliClient) openConn() (conn.Conn, error) {
...
 func NewCliClient(dialer conn.Dialer) (*CliClient, error) {
+	r := &CliClient{nil, dialer, nil}
+
 	rootCmd := &cobra.Command{
-		Use:   "scootcl",
-		Short: "Scootcl is a command-line client to Scoot",
+		Use:                "scootcl",
+		Short:              "Scootcl is a command-line client to Scoot",
+		Run:                func(*cobra.Command, []string) {},
+		PersistentPostRunE: r.Close,
```
jasonbs10: I like this, removing responsibility from the caller when possible.

local/client/cli/cli_test.go
```diff
@@ -0,0 +1,204 @@
...
+	return out, outCh
+}
+
+func (o *output) WaitAndReset() (string, string) {
+	stdoutCh, stderrCh := o.stdoutCh, o.stderrCh
+	o.Reset()
+	return <-stdoutCh, <-stderrCh
+}
+
+func (o *output) ResetIfNecessary() {
```
jasonbs10: Not sure why this needs to be a separate fn.

local/client/cli/cli_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+func makeDiversion() (*os.File, chan string) {
+	reader, out, err := os.Pipe()
+	if err != nil {
+		panic("Cannot create pipe")
+	}
+	outCh := make(chan string)
+
+	go func() {
+		var buf bytes.Buffer
```
jasonbs10: Is this the best way to capture data, all at once vs streaming?

local/client/cli/cli_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+	output := captureOutput()
+	defer output.ResetIfNecessary()
+
+	err := cl.Exec()
+	stdout, stderr := output.WaitAndReset() // Reset early so we can use stdout/stderr and write to uncaptured stdout/stderr
+	return stdout, stderr, err
+}
+
+// Below here is code to capture stdout and stderr
```
jasonbs10: We should have a test utility class for this type of stuff.

jasonbs10:
lgtm

local/client/cli/echo_cmd.go
dbentley: yes; log includes info like a timestamp (and I think prints to stderr?)

fmt.Print* print straight to stdout, which is what you want for a Unix tool.

local/client/cli/cli_test.go
dbentley: It doesn't per se, I just thought it looked nicer. I'm happy to collapse them; would you prefer that?

local/client/cli/cli_test.go
dbentley: I bet io.Copy streams under the hood. As long as stdout and stderr will fit into a string I think this should be fine.

local/client/cli/cli_test.go
dbentley: Sure, I'm happy to move this if we find that other tests want to capture stdout and stderr, but I suspect the only tools that really care about stdout and stderr are command-line tools, and I hope we only have one of those.

jasonbs10: Good point. Would we want to capture output from worker?

local/client/cli/cli_test.go
jasonbs10: Yeah, nm.

local/client/cli/cli_test.go
jasonbs10: Unless you're communicating something subtle about resets, I don't think the caller cares about the details of a reset, so I'd collapse it.

local/client/cli/cli_test.go
dbentley: I think we want the Runner that's in the Worker to capture output of the commands it runs, but I don't think we care about "did the Worker log X message". stdout and stderr are part of the interface of command-line tools, but not servers.

WDYT?

jasonbs10: That's what I mean, testing the runner, not the worker. You should clearly leave this here for now and we can pull it out if needed.

dbentley: Yeah, then we'll want stdout and stderr of what we run, not just for testing but for implementation.

golang's built-in os/exec already do this: https://golang.org/pkg/os/exec/

The tricky thing that we're doing is that we want to capture output for only *part* of an execution (as opposed to for the lifetime of the process)

PR #24, opened 2016-06-14
Ported 'source' minfs to github.com/scootdev/scoot.

jasonbs10:


fs/utils/utils.go
```diff
@@ -0,0 +1,86 @@
+package utils
```
dbentley: how about fs/performance ?

(That could be in a separate, follow-up PR)

dbentley:
LGTM

fs/utils/utils.go
jasonbs10: Sure, that's certainly more descriptive.

PR #25, opened 2016-06-15
Converted binary/minfs/main.go to a small wrapper.

jasonbs10:
Converted binary/minfs/main.go to a small wrapper that calls common functions in fs/minfuse. Renamed utils->perf.

### DIFF
--- a/fs/minfuse/runfs.go
+++ b/fs/minfuse/runfs.go
@@ -19,13 +19,13 @@ import (
 )
 
 type Options struct {
-	src          string
-	mountpoint   string
-	strategyList []string
-	async        bool
-	trace        bool
-	threadUnsafe bool
-	maxReadahead uint32
+	Src          string
+	Mountpoint   string
+	StrategyList []string
+	Async        bool
+	Trace        bool
+	ThreadUnsafe bool
+	MaxReadahead uint32
 }
 
 func SetupLog() {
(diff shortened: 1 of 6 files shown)


fs/minfuse/runfs.go
```diff
@@ -0,0 +1,142 @@
...
+		default:
+			log.Fatal("Unrecognized strategy", strategy)
+		}
+	}
+	log.Print(opts)
+	return &opts, nil
+}
+
+func Runfs(opts *Options) {
+	snap := snapshots.NewFileBackedSnapshot(opts.src, "only")
```
dbentley: I'd move the snapshot creation to main.go, and then have this function take a Snapshot.

(To use the terminology, a Snapshot is a dependency of minfs, and so we should inject the dependency instead of having the fs create it)

fs/minfuse/runfs.go
```diff
@@ -0,0 +1,142 @@
...
+	src          string
+	mountpoint   string
+	strategyList []string
+	async        bool
+	trace        bool
+	threadUnsafe bool
+	maxReadahead uint32
+}
+
+func SetupLog() {
```
dbentley: I'd also just leave this in main.go

dbentley:
LGTM

PR #26, opened 2016-06-15
snapshots/fake: fix errors in go vet

dbentley:


local/protocol/translate.go
```diff
@@ -4,6 +4,8 @@ import (
 	"github.com/scootdev/scoot/runner"
 )
 
+// TODO(dbentley): we should go generate with command protoc local.proto --go_out=plugins=grpc:.
+
```
CaitieM20: typically when I put TODOs like this in code, I'll file a jira and also link the number.  Do we want to do that as well.  Super useful to know if it was actually ever done or punted on.  Ensures we don't lose track of tasks

CaitieM20:
LGTM

PR #27, opened 2016-06-16
Saga Recovery

CaitieM20:
Saga Recovery will happen at Startup.  Once a Scheduler starts up it should read, it's saga log, and then reschedule non completed work.  Startup returns a list of SagaIds, RecoverSagaState should then be called per SagaId.  These methods are separate to allow the caller to determine how to handle failures (how many times to retry etc...) also to handle the rate at which the SagaLog is polled, and how quickly tasks are rescheduled.  These decisions are specific to an implementation and therefore should not be abstracted from the user by the Saga package.  

This code review also contains Saga Recovery Logic, which reads Sagas from the SagaLog, replays messages and verifies state is correct, and checks if its safe to proceed otherwise aborts the Saga.  

The demo has been updated to poll the InMemorySagaLog (which just returns the ids of all sagas started since creation) after running all the work.  It then verifies that all sagas are in the completed state.

### DIFF
--- a/saga/inMemorySagaLog.go
+++ b/saga/inMemorySagaLog.go
@@ -69,3 +69,19 @@ func (log *inMemorySagaLog) GetMessages(sagaId string) ([]sagaMessage, error) {
 		return nil, nil
 	}
 }
+
+/*
+ * Returns all Sagas Started since this InMemory Saga was created
+ */
+func (log *inMemorySagaLog) GetActiveSagas() ([]string, error) {
+	log.mutex.RLock()
+	keys := make([]string, 0, len(log.sagas))
+
+	for key, _ := range log.sagas {
+		keys = append(keys, key)
+	}
+
+	log.mutex.RUnlock()
+
+	return keys, nil
+}
--- a/saga/inMemorySagaLog.go
+++ b/saga/inMemorySagaLog.go
@@ -27,32 +27,31 @@ func MakeInMemorySaga() Saga {
 }
 
 func (log *inMemorySagaLog) LogMessage(msg sagaMessage) error {
-	fmt.Println(fmt.Sprintf("Saga %s: %s %s", msg.sagaId, msg.msgType.String(), msg.taskId))
-
-	sagaId := msg.sagaId
-	var err error
 
 	log.mutex.Lock()
+	defer log.mutex.Unlock()
+
+	fmt.Println(fmt.Sprintf("Saga %s: %s %s", msg.sagaId, msg.msgType.String(), msg.taskId))
+	sagaId := msg.sagaId
+
 	msgs, ok := log.sagas[sagaId]
 	if !ok {
 		return errors.New(fmt.Sprintf("Saga: %s is not Started yet.", msg.sagaId))
 	}
 
 	log.sagas[sagaId] = append(msgs, msg)
-	log.mutex.Unlock()
-	return err
+	return nil
 }
 
 func (log *inMemorySagaLog) StartSaga(sagaId string, job []byte) error {
 
 	log.mutex.Lock()
+	defer log.mutex.Unlock()
 
 	fmt.Println(fmt.Sprintf("Start Saga %s", sagaId))
 
 	startMsg := MakeStartSagaMessage(sagaId, job)
 	log.sagas[sagaId] = []sagaMessage{startMsg}
-
-	log.mutex.Unlock()
 
 	return nil
 }
@@ -60,8 +59,9 @@ func (log *inMemorySagaLog) StartSaga(sagaId string, job []byte) error {
 func (log *inMemorySagaLog) GetMessages(sagaId string) ([]sagaMessage, error) {
 
 	log.mutex.RLock()
+	defer log.mutex.RUnlock()
+
 	msgs, ok := log.sagas[sagaId]
-	log.mutex.RUnlock()
 
 	if ok {
 		return msgs, nil
@@ -75,13 +75,13 @@ func (log *inMemorySagaLog) GetMessages(sagaId string) ([]sagaMessage, error) {
  */
 func (log *inMemorySagaLog) GetActiveSagas() ([]string, error) {
 	log.mutex.RLock()
+	defer log.mutex.RUnlock()
+
 	keys := make([]string, 0, len(log.sagas))
 
 	for key, _ := range log.sagas {
 		keys = append(keys, key)
 	}
-
-	log.mutex.RUnlock()
 
 	return keys, nil
 }
--- a/saga/saga.go
+++ b/saga/saga.go
@@ -1,22 +1,5 @@
 package saga
 
-type SagaRecoveryType int
-
-/*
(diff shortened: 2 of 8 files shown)


saga/inMemorySagaLog.go
```diff
@@ -69,3 +69,19 @@ func (log *inMemorySagaLog) GetMessages(sagaId string) ([]sagaMessage, error) {
 		return nil, nil
 	}
 }
+
+/*
+ * Returns all Sagas Started since this InMemory Saga was created
+ */
+func (log *inMemorySagaLog) GetActiveSagas() ([]string, error) {
+	log.mutex.RLock()
```
dbentley: Maybe just defer the unlock after acquiring it?

saga/saga.go
```diff
@@ -148,3 +111,48 @@ func (s Saga) StartCompensatingTask(state *SagaState, taskId string, data []byte
 func (s Saga) EndCompensatingTask(state *SagaState, taskId string, results []byte) (*SagaState, error) {
 	return s.logMessage(state, MakeEndCompTaskMessage(state.sagaId, taskId, results))
 }
+
+/*
+ * Should be called at Saga Creation time.
+ * Returns a Slice of In Progress SagaIds
+ */
+func (s Saga) Startup() ([]string, error) {
```
dbentley: Will Startup ever do anything other than just return GetActiveSagas? Maybe this should just be GetActiveSagas?

dbentley:
LGTM

saga/saga.go
CaitieM20: I named it Startup because you suggested calling it Startup when we were in New York, since you should always call this when creating a Saga

dbentley: Yup, I'm fine with Startup, and also acknowledge that I am not always right, so if there's something you think is better, I'm sorry for having led you astray.

WDYT is best going forward?

(Feel free to just do what you think is best and then merge, don't need to consider this blocking)

saga/inMemorySagaLog.go
CaitieM20: defer has a pretty high performance cost, given how simple this code is i just chose to explicitly place it

saga/saga.go
CaitieM20: I like Startup! you shouldn't be calling this method as part of normal execution.  Its only meant for Recovery mode which is at startup time!

saga/inMemorySagaLog.go
dbentley: The cost is fairly low, if this isn't a hot spot I'd suggest we follow the idiom.

If this is a hot spot, that sounds like performance optimization. Does it make sense to use the syntax for performance optimizations we discussed?

dbentley: (BTW, just googled an found this, https://github.com/golang/go/issues/6980 , which shows they know about it, and the cost is ~70ns)

CaitieM20: sure just switched to lock, defer unlock since its just test code

saga/sagaRecovery.go
```diff
@@ -0,0 +1,101 @@
...
+		// And compensating tasks should start
+		if !isSagaInSafeState(state) {
+			state, err = saga.AbortSaga(state)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+	case ForwardRecovery:
+
```
jasonbs10: Are you omitting the TODO here because that would mean having to create a new ticket? Would it be ok to attribute this to an existing/task story?

saga/sagaRecovery.go
```diff
@@ -0,0 +1,101 @@
...
+			return nil, err
+		}
+	}
+
+	// Check if we can safely proceed forward based on recovery method
+	// RollbackRecovery must check if in a SafeState,
+	// ForwardRecovery can always make progress
+	switch recoveryType {
+
+	case RollbackRecovery:
```
jasonbs10: I don't really understand why we're modifying SagaState here, that's more than just recovery. Also, how does the caller know if it should call compensating tasks due to an abort here (if that's the intent), and what is the expected action for a logged abort vs one we've applied right here?

saga/sagalog.go
```diff
@@ -24,4 +24,12 @@ type SagaLog interface {
 	 * specified saga.
 	 */
 	GetMessages(sagaId string) ([]sagaMessage, error)
+
+	/*
+	 * Returns a list of all in progress sagaIds.
+	 * This MUST include all not completed sagaIds.
+	 * It may also included completed sagas
```
jasonbs10: What's the definition of active saga? When would completed sagas be conditionally included?

sched/demo/main.go
```diff
@@ -30,21 +33,50 @@ func main() {
...
+
+		sagaState, err := saga.RecoverSagaState(sagaId, s.ForwardRecovery)
+		if err != nil {
+			// For now just print error in actual scheduler we'd want to retry multiple times,
+			// before putting it on a deadletter queue
+			fmt.Println(fmt.Sprintf("ERROR recovering saga state for %s: %s", sagaId, err))
+		}
+
+		// all Sagas are expected to be completed
+		if !sagaState.IsSagaCompleted() {
```
jasonbs10: Do we ever exercise this or is ids always empty on startup?

jasonbs10: I'd remove the todo in scheduleWork() if this is always expected to hold.

saga/sagaRecovery.go
CaitieM20: There is nothing really to do on ForwardRecovery, you are always in a safe state.  I can add a comment, explaining this, since Tasks have to be Idempotent.

saga/sagaRecovery.go
CaitieM20: So in Sagas, RollbackRecovery means we apply compensating actions when it fails.  When this method is called, the scheduler has failed so if there is no start/end task pairing a saga is in an unsafe state.  This is because tasks do not have to be idempotent in RollbackRecovery mode.  (This is all in the talk I linked / slides) So we have to abort if we aren't in safe state, because its not safe to just replay a task.

saga/sagalog.go
CaitieM20: Active Saga should be something that has started but not finished, but this is a perf optimization.  Implementations could just return everything and recovery will replay the log to figure out what's done vs not done.

saga/sagaRecovery.go
CaitieM20: So after recovery runs, users should check if a saga has been aborted, and then start applying compensating actions that need to be applied.  If they try to continue applying tasks to an aborted Saga, SagaState update method will catch an invalid transition and not apply it and throw an error.

sched/demo/main.go
CaitieM20: I plan to change it once we build the Node component.  The InMemoryNode will randomly generate errors so we can verify the system works correctly with dropped messages or downed nodes etc...

PR #28, opened 2016-06-16
Caitie/make dependencies

CaitieM20:
Adding local commands back to makefile.  These are useful for running & testing locally, and for quickly getting setup with scoot as defined on go/scootdev

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,8 @@ default:
 	go build ./...
 
 dependencies: 
+	# get all dependencies needed to run scoot / check if there have
+	# been updates since they were first installed
 	go get -t -u github.com/scootdev/scoot/...
 
 	# mockgen is only referenced for code gen, not imported directly
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ dependencies:
 	go get -t -u github.com/scootdev/scoot/...
 
 	# mockgen is only referenced for code gen, not imported directly
-	go get github.com/golang/mock/mockgen
+	go get -u github.com/golang/mock/mockgen
 
 generate: 
 	go generate ./...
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,9 +8,25 @@ PROJECT_URL := "https://github.com/scootdev/scoot"
 default:
 	go build ./...
 
+dependencies: 
+	# get all dependencies needed to run scoot / check if there have
+	# been updates since they were first installed
+	go get -t -u github.com/scootdev/scoot/...
+
+	# mockgen is only referenced for code gen, not imported directly
+	go get github.com/golang/mock/mockgen
```
dbentley: Might be worth having this install mockgen too?

Makefile
```diff
@@ -8,9 +8,25 @@ PROJECT_URL := "https://github.com/scootdev/scoot"
...
+
+generate:
```
dbentley: maybe this should depend on dependencies ?

dbentley:
LGTM

jasonbs10:
lgtm

Makefile
CaitieM20: it does install mockgen.  Second line.

dbentley: Sorry, I meant "go install" after "go get".

CaitieM20: ran a little experiment, deleted my whole bin directory, and then ran:

- make dependencies
-make testlocal

everything passes and bin/mockgen is created in the process so I don't think we need to explicitly install this.

Makefile
CaitieM20: yes added

CaitieM20: actually on second thought, dependencies takes a while to run because it has to check if any of the packages changed over the network too, and I often want to generate code while not connected

so for now just leaving it off, I added fullbuild target which will check dependencies, generate code, and run all tests, this is basically our checkin build, grab latest dependencies, generate code with them, then run tests.

PR #29, opened 2016-06-16
scootapi: no-op server and client

dbentley:


scootapi/client/run_job_cmd.go
```diff
@@ -0,0 +1,47 @@
+package client
+
+import (
+	"fmt"
+	"github.com/scootdev/scoot/scootapi/gen-go/scoot"
```
jasonbs10: Maybe we can adopt a [import] naming convention for generated source?

binaries/scootapi/main.go
```diff
@@ -0,0 +1,22 @@
+package main
+
+import (
+	"git.apache.org/thrift.git/lib/go/thrift"
+	"github.com/scootdev/scoot/scootapi/client"
+	"log"
+)
+
+func main() {
+	protocolFactory := thrift.NewTBinaryProtocolFactoryDefault()
```
jasonbs10: Does the scootapi user care that we're using thrift? Should we hide this?

scootapi/client/run_job_cmd.go
dbentley: OK, what would you like?

binaries/scootapi/main.go
dbentley: Probably, because you need the client and server to match. I expect you may end up with different flavors of Scoot using different protocols and transports.

proc/proc.go
```diff
@@ -0,0 +1,32 @@
+package proc
+
+import (
+	"github.com/scootdev/scoot/runner"
+)
+
+type JobType int
```
jasonbs10: Where is this file used vs the thrift def, i.e. what is the boundary.

It's a bit confusing to have jobtype as string in some places and int here.

dbentley: Thrift object should be used only for serialization and RPC, so as soon as you get a Thrift object, you should convert to a domain object.

Where do you see jobtype as a string?

binaries/scootapi/main.go
jasonbs10: It's not a huge deal, but why does pants care about any of that?

proc/proc.go
jasonbs10: I see it in messages/jobMessage.go. Am I looking at something outdated?

binaries/scootapi/main.go
dbentley: Yup, pants doesn't care about it.

So, I think we need to talk more at some point about the difference between the local scoot interface (that pants needs) and the cloud scoot api that Iron will use to run things.

This code is for the latter.

scootapi/client/run_job_cmd.go
jasonbs10: What were we doing in scala? Prefixing with 'T'?

binaries/scootapi/main.go
jasonbs10: Ah, right. Can you add a comment to that effect since I keep forgetting?

proc/proc.go
dbentley: Ah, no. Yeah, these aren't completely compatible.

jasonbs10:
lgtm

binaries/scootapi/main.go
dbentley: Done.

scootapi/client/run_job_cmd.go
dbentley: Thrift ends up putting T at the beginning of lots of Thrift types...

scootapi/api.go
```diff
@@ -0,0 +1,3 @@
+package scootapi
+
+// We should use go generate to run thrift --gen go:package_prefix=github.com/scootdev/scoot/scootapi/gen-go/ scoot.thrift
```
jasonbs10: Wait, what goes here now that you've moved everything to server?

dbentley: I think just the line to run go generate; WDYT?

jasonbs10: If it doesn't function as an actual directive then I think it's a bit unusual to add this note to source code vs something higher level.

PR #30, opened 2016-06-17
update to use github.com thrift

dbentley:


jasonbs10:
lgtm. Can you add a comment saying why you switched?

dbentley:
Switching because the github.com code is already present inside twitter, easier to read, and easier to resolve (because of how pants looks at import tags and meta fields)

PR #31, opened 2016-06-20
Property Based Testing for Sagas

CaitieM20:
I tested out property based testing in Go using [gopter](https://github.com/leanovate/gopter).  Property based testing allows us to define properties that our code should have and then generate input.  If a counterexample is found the test fails and the input that caused it to fail is returned.  This is great because it helps us explore more of the state space of possible inputs into our program.  I also personally like it because using propositional logic to define the properties of a system typically helps me uncover missing cases in my code!

While doing property based testing I found some missed cases which I have updated in the SagaState Update method.  This is also in this code review.  

I did it first for SagaState updateState method.  This is the core part of the Saga State Machine logic, and has a multitude of inputs.  Too many to enumerate in unit tests.  These are defined in sagaState_prop_test.go and run as part of our normal test suite.

I added Saga Specific Generators for property based testing to sagaGenerators.go .  These are methods that help us randomly generate a valid SagaState or Ids. 

I also removed a bunch of unit tests from sagaState_test.go that are covered in the property based tests.  The property based tests are more succinct and have the same code coverage.


saga/sagaGenerators.go
```diff
@@ -0,0 +1,140 @@
...
+// Generator for a Valid Saga State
+func GenSagaState() gopter.Gen {
+	return func(genParams *gopter.GenParameters) *gopter.GenResult {
+		state := genSagaState(genParams)
+		genResult := gopter.NewGenResult(state, gopter.NoShrinker)
+		return genResult
+	}
+}
+
+// State is a struct object and not a reference for readability
```
dbentley: Is it easy to just define a String() function so that we can print it how we want it?

dbentley:
LGTM

saga/sagaState_prop_test.go
```diff
@@ -0,0 +1,215 @@
...
+			validUpdate := validTransition && err == nil &&
+				newState != nil && newState.IsSagaAborted()
+			errorReturned := !validTransition && err != nil && newState == nil
+
+			return validUpdate || errorReturned
+		},
+		GenSagaState(),
+	))
+
+	// StartTask messages are valid unless a Saga has been Completed or Aborted
```
jasonbs10: In what situation would we want to allow starting a task after it has already been ended? And starting a comp task after it has also ended?

saga/sagaState_prop_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+import (
+	"bytes"
+	"github.com/leanovate/gopter"
+	"github.com/leanovate/gopter/gen"
+	"github.com/leanovate/gopter/prop"
+	"testing"
+)
+
+func Test_ValidSagaId(t *testing.T) {
```
jasonbs10: Aren't you basically reimplementing validateSagaId() and validateTaskID() in these tests? It seems like these would be better suited for unit tests if the implementation can't be broken down into different properties. (this is assuming we'll eventually do more than checking for empty string)

jasonbs10:
This is great! lgtm

saga/sagaState_prop_test.go
CaitieM20: Yes agree this is probably better as a unit test, will revert.  This was baby's first property based test in Go, but probably more succinct and always forces the error as a unit test.

saga/sagaState_prop_test.go
CaitieM20: After chatting, you are correct a SagaState should never allow this transition.  However the SagaLog has to interpret it correctly due to the potential for multiple writers in our system.  Will Fix!

saga/sagaGenerators.go
CaitieM20: We definitely can. I think getting all the nuance correct will take more than 5 min, lots of moving parts with SagaState will file a ticket to do this because I think it'll be really useful in future testing and debugging!

PR #32, opened 2016-06-22
scootapi: allow addr as a flag

dbentley:
In server: log requests.

### DIFF
--- a/binaries/scootapi/main.go
+++ b/binaries/scootapi/main.go
@@ -11,8 +11,7 @@ func main() {
 	protocolFactory := thrift.NewTBinaryProtocolFactoryDefault()
 	transportFactory := thrift.NewTTransportFactory()
 
-	dialer := client.NewDialer(transportFactory, protocolFactory)
-	client, err := client.NewClient(dialer)
+	client, err := client.NewClient(transportFactory, protocolFactory)
 	if err != nil {
 		log.Fatal("Cannot initialize Cloud Scoot CLI: ", err)
 	}
--- a/binaries/scootapi/main.go
+++ b/binaries/scootapi/main.go
@@ -11,7 +11,7 @@ func main() {
 	protocolFactory := thrift.NewTBinaryProtocolFactoryDefault()
 	transportFactory := thrift.NewTTransportFactory()
 
-	dialer := client.NewDialer("localhost:9090", transportFactory, protocolFactory)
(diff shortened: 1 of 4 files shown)


scootapi/client/run_job_cmd.go
```diff
@@ -8,11 +8,13 @@ import (
 )
 
 func makeRunJobCmd(c *Client) *cobra.Command {
-	return &cobra.Command{
+	r := &cobra.Command{
 		Use:   "run_job",
 		Short: "run a job",
 		RunE:  c.runJob,
 	}
+	r.Flags().StringVar(&c.dialer.addr, "addr", "localhost:9090", "address to connect to")
```
jasonbs10: Can you add a setter in Dialer to signal our intent to reset the addr externally?

binaries/scootapi/main.go
```diff
@@ -11,7 +11,7 @@ func main() {
 	protocolFactory := thrift.NewTBinaryProtocolFactoryDefault()
 	transportFactory := thrift.NewTTransportFactory()
 
-	dialer := client.NewDialer("localhost:9090", transportFactory, protocolFactory)
+	dialer := client.NewDialer(transportFactory, protocolFactory)
```
jasonbs10: I'm not sure that removing addr is necessary here, just pass a blank string like in cli.go.

dbentley: My concern is that it'd be easy to think this is the default then, but then it gets overwritten by the default in the flag. WDYT?

(BTW, fwiw, overall I think this is kinda shady all around; I'm not super happy with how this is working out. Maybe just get rid of dialer altogether and just have it baked into the client object instead of trying to separate them?)

scootapi/client/run_job_cmd.go
dbentley: I think a comment next to the field is harder to miss, so I tried that. WDYT?

jasonbs10: That's fine.

binaries/scootapi/main.go
jasonbs10: I think that would be better, however I'm willing to brainstorm about how to keep Dialer and not be shady.

dbentley:
How's this?

jasonbs10:
lgtm

PR #33, opened 2016-06-22
First pass for workerapi.

jasonbs10:


workerapi/worker.thrift
```diff
@@ -0,0 +1,38 @@
...
+  2: optional string taskId
+  3: optional string outUri
+  4: optional string errUri
+  5: optional string info
+  6: optional i32 exitCode
+}
+
+struct Task {
+  1: required list<string> argv
+  2: optional list<string> env
```
dbentley: I'd imagine a map.

workerapi/worker.thrift
```diff
@@ -0,0 +1,38 @@
...
+  3: optional string dir
```
dbentley: What's dir?

workerapi/worker.thrift
```diff
@@ -0,0 +1,38 @@
...
+  4: optional string taskId
```
dbentley: It's hard to reason about what to do in weird cases where the client sends us bogus task ids. What if it was just up to the worker to create this?

workerapi/worker.thrift
```diff
@@ -0,0 +1,38 @@
...
+  5: optional string sourceId //Allows snapshotId or patchId
+  6: optional i32 timeout_ms
```
dbentley: why underscore instead of camel-casing?

workerapi/worker.thrift
```diff
@@ -0,0 +1,38 @@
...
+  5: optional string info
+  6: optional i32 exitCode
+}
+
+struct Task {
+  1: required list<string> argv
+  2: optional list<string> env
+  3: optional string dir
+  4: optional string taskId
+  5: optional string sourceId //Allows snapshotId or patchId
```
dbentley: I'd suggest snapshot ID. I think Snapshot is the right interface. And then one way to specify an interface could be applying a patch. (Especially given the recent turmoil with the patch service)

workerapi/worker.thrift
```diff
@@ -0,0 +1,38 @@
...
+  6: optional i32 timeout_ms
+}
+
+service Worker {
+  WorkerStatus QueryStatus()
+  WorkerStatus Run(1: Task task)
+  WorkerStatus Abort(
```
dbentley: What's force?

workerapi/worker.thrift
```diff
@@ -0,0 +1,38 @@
+enum Status {
+  UNKNOWN = 0
+  READY = 1
+  PENDING = 2
+  FAILURE = 3
```
dbentley: Failure and Success seem like the same to the worker. Maybe collapse them into COMPLETED?

workerapi/worker.thrift
```diff
@@ -0,0 +1,38 @@
...
+  RUNNING = 4
+  SUCCESS = 5
+  ABORTED = 6
+  TIMEOUT = 7
+}
+
+struct WorkerStatus {
```
dbentley: Seems like this isn't the status of the worker, but of a Task.

workerapi/worker.thrift
dbentley: (to be clearer: I don't think the client of the Worker has any idea what directory the worker should run in)

workerapi/worker.thrift
jasonbs10: That would be better, I'll change it.

workerapi/worker.thrift
jasonbs10: Directory where we want to run the command. I was going to add comments after we all roughly agreed on the interface, is that all you'd want here?

dbentley: I'm concerned it's too leaky an interface. I think it's dangerous if the client of the Worker knows what directories the worker controls.

Put another way: what does this surface area of the interface buy us?

workerapi/worker.thrift
jasonbs10: This taskID is only shared between the scheduler and the worker so we know that there wasn't some cluster issue and the worker is doing what we think it's doing.

workerapi/worker.thrift
jasonbs10: No reason, I've always done underscore for units, I can change it though.

dbentley: ISTM we should use either camel casing or underscores and mixing them in one thrift seems odd.

workerapi/worker.thrift
jasonbs10: If we try to abort a worker that isn't running the taskID specified, it's a no-op. Force unconditionally aborts.

workerapi/worker.thrift
dbentley: What if somehow two schedulers are going and send the same task id?

The scheduler can maintain a map from taskID -> ID on Worker

workerapi/worker.thrift
jasonbs10: I'm not sure I follow. In that case, aborted and timeout could also be folded in? I think these are all the unique cases that we might care about upstream.

workerapi/worker.thrift
jasonbs10: I see them as the same thing, the worker is the task. If we decide we want meta info about the worker that's independent of task, then we need to decide what that info is. I can't think of anything useful for MVP.

workerapi/worker.thrift
dbentley: Oh! I see, I was thinking about it wrong. This is why my WorkerStatus comment doesn't make much sense.

This API seems to lock us into "the Worker is running at most one task at a time." Which is certainly where the implementation is going to start. But it seems like an awfully stateful protocol.

It might be nice if the protocol supported a Worker that ran a few things.

workerapi/worker.thrift
dbentley: What if you get status FAILURE but ExitCode 0? Or vice versa?

You can't tell aborted and timeout apart in that way, so that's why they should be separate states.

workerapi/worker.thrift
jasonbs10: This is a relative directory inside the source sandbox. I don't think pants tasks would need this, but an arbitrary command might?

dbentley: Oh! I see!

In that case, the command could just start by cd'ing into that directory.

workerapi/worker.thrift
jasonbs10: The taskID could be UUID and for all intents and purposes will never conflict (for example). If there's some aurora screw-up then we want to confirm consistency between the map scheduler has and what's on each worker. Right?

workerapi/worker.thrift
jasonbs10: Ok, will do.

workerapi/worker.thrift
jasonbs10: I don't think we support that if we define argv as []string. The first arg is expected to be an executable so the user would be forced to wrap everything in a 'bash -c'.

workerapi/worker.thrift
jasonbs10: Yeah, you're right.

workerapi/worker.thrift
jasonbs10: I think that's a separate discussion, if it's valuable to support multiple tasks per worker. Can we think of a motivating case?

workerapi/worker.thrift
jasonbs10: I don't think we care what the error code is but we don't want to drop it either. In that light, FAILURE and SUCCESS are clearly separate states I'd think.

workerapi/worker.thrift
dbentley: Yes, the best way to do that is to have the Worker maintain its own ID space.

Let me give you another example:
the Scheduler schedules task <id1> on worker 1.
The task fails, and the Scheduler tries to reschedule it on both worker 1 and worker 2.

Then worker1 has seen the same id twice. This is confusing.

Also confusing is if there are two schedulers.

This is confusing because we are conflating the desire (run this Task) and the action (one Run of the Task). We may want to Run one Task multiple times.

I think the way to make this simplest is to have the Worker interface assign a new ID each time we start a Run.

Yes, that means it's up to the Scheduler to maintain a mapping of Task -> List<Run>.

BUT, that's also a *good* thing: the Scheduler needs to know to not rerun the Task if it's already failed N times!

WDYT?

workerapi/worker.thrift
dbentley: Or run foo.py. I suspect that most commands we run will actually be shell scripts (as you found when you try to fit into Iron), and so won't need this.

(FWIW, I think we don't need environment variables for the same reason)

Let's not try to become a general purpose shell, let's *just* run.

workerapi/worker.thrift
dbentley: Sure, the motivating case is that it lets us increase utilization because not all processes are created equal.

I think it *also* leads to a cleaner API that is less stateful. With the current API, if a Task gets scheduled but I didn't collect the results, whoops, I'm out of luck.

My suggestion would be:
Run(TaskDescription) -> ID (or maybe the whole Status)
Query(ID) -> Status
Abort(ID) -> Status

BUT, this API means that if you don't know what's running, you can't clear the Worker. Whoops.

One approach would be to have:
AbortAll()

BUT, that can be bad because you have a scheduler canceling jobs it doesn't know about.

What would be better is:
Running() -> List<ID>

Or, to get back to what you had that's pretty nice might be:
Status() -> struct WorkerStatus {
  List<ID>
  // other info that we add
}

So, WorkerStatus just says "I'm doing X, Y, Z". If the Scheduler wants to know about X, Y, Z, it has to query each. If it wants to kill X, Y, Z, it can do that explicitly.

IF there is another Scheduler that is scheduling A, B, C so that a Scheduler thought it had cleared the worker by killing X, Y, Z, then we can see the expectation violated explicitly, instead of having this "AbortAll" just confuse us.

workerapi/worker.thrift
dbentley: OK, so, maybe I'm wrong and FAILURE means something else.

Let me make an analogy:
when you use the Postal Service to apply to college, you don't see if you got accepted or denied. You just get the letter. From the Worker's perspective, exitcode 0 and exitcode 1 are the same: they're different payloads for a completed process.

I think our API should reflect that. To us, processes don't fail or succeed, they complete.

workerapi/worker.thrift
jasonbs10: I see your point. I was initially thinking we'd keep some version of the jenkins-iron script to use with dart, and that takes env vars. I think we'll scrap most of that script anyway, so just argv should be fine.

workerapi/worker.thrift
jasonbs10: We should assume until proven otherwise that the exit code is always non-zero on failure (I thought it could be zero). If that's the case then, sure, COMPLETE plus an exit code payload is probably sufficient.

workerapi/worker.thrift
jasonbs10: Ok, in the case where we have heuristics telling us which targets are parallel vs serial, we would definitely want multiple tasks to ensure all cpu cores are used (assuming there are no memory constraints). I don't know how much is serial vs parallel with pants and that would be good to know first.  TBH I would have categorized this as premature functionality with the caveat that we'd have to do a v2 protocol when actually needed.

workerapi/worker.thrift
jasonbs10: This would mean we can't batch schedule tasks because we won't know which taskID belongs to which task. Do we want batching?

Also, the scheduler is free to generate a new id if rerunning a failed task, although I personally prefer reusing the id since that gets us a single id for the entire task lifetime, and I think that's desirable for debugging at the very least. I'm not entirely following your logic and we should discuss this further offline.

workerapi/worker.thrift
```diff
@@ -0,0 +1,42 @@
...
+  READY = 1
+  PENDING = 2
+  RUNNING = 3
+  COMPLETED = 4
+  ABORTED = 5
+  TIMEOUT = 6
+}
+
+struct WorkerStatus {
+  1: required list<string> taskIds
```
dbentley: Maybe "running"? or "running tasks"?

dbentley:
LGTM

PR #34, opened 2016-06-23
runner: make a simple runner, introduce execer interface

dbentley:
Still way too early for actual review, but shows direction.


dbentley:
This now does simple running, so I think it's ready for review to begin (though I'm not sure it's completely there)

jasonbs10:
lgtm, although I'm not sure why the tests are failing. Do they work locally?

PR #35, opened 2016-06-24
Moved all direct dependencies into the vendor dir.

jasonbs10:


dbentley:
How will this work for using Scoot as a library? Won't it mean that Scoot will compile with a different version of dependencies than the binaries that you use when you import Scoot into a monorepo?

Oh, maybe this won't have that because if you just get a tar, you won't get this code because you won't use submodules?

dbentley:
LGTM

Yeah, as long as this is going to let us import it into a monorepo this is great.

(I have a vague discomfort with git submodules, only because I've heard they're never quite right, but it seems like this use is good)

PR #36, opened 2016-06-24
sched: create queue abstraction and simple implementation

dbentley:
Move messages into sched package.

Change some capitalization/names in messages

### DIFF
--- a/sched/queue/memory/simple.go
+++ b/sched/queue/memory/simple.go
@@ -1,0 +1,36 @@
+package memory
+
+// In-memory Scheduler Queue
+
+import (
+	"github.com/scootdev/scoot/sched"
+)
+
+type simpleQueue struct {
+	nextID  int
+	nextJob *sched.Job
+	readCh  chan sched.Job
+	inCh    chan enqueueReq
+}
+
+type enqueueReq struct {
+	job      sched.Job
+	resultCh chan enqueueResult
+}
+
+type enqueueResult struct {
+	jobID sched.JobID
+	err   error
+}
+
+func (q *simpleQueue) Enqueue(job sched.Job) (sched.JobID, error) {
+	return sched.JobID(""), fmt.Errorf("Not yet implemented")
+}
+
+func (q *simpleQueue) Read() (chan sched.Job, error) {
+	return nil, fmt.Errorf("Not yet implemented")
+}
+
+func (q *simpleQueue) Close() error {
+	return nil
+}
--- a/sched/queue/memory/simple.go
+++ b/sched/queue/memory/simple.go
@@ -55,6 +55,8 @@ func (q *simpleQueue) Enqueue(job sched.Job) (string, error) {
 }
 
 func (q *simpleQueue) Close() error {
+	close(q.inCh)
+	close(q.outCh)
 	return nil
 }
 
--- a/sched/queue/memory/simple.go
+++ b/sched/queue/memory/simple.go
@@ -3,14 +3,24 @@ package memory
 // In-memory Scheduler Queue
 
 import (
+	"fmt"
 	"github.com/scootdev/scoot/sched"
+	"github.com/scootdev/scoot/sched/queue"
+	"time"
 )
 
+func NewSimpleQueue() (queue.Queue, chan sched.Job) {
+	q := &simpleQueue{}
+	q.outCh = make(chan sched.Job, 1)
+	q.inCh = make(chan enqueueReq)
+	go q.loop()
+	return q, q.outCh
+}
+
 type simpleQueue struct {
-	nextID  int
-	nextJob *sched.Job
-	readCh  chan sched.Job
-	inCh    chan enqueueReq
+	nextID int
+	inCh   chan enqueueReq
+	outCh  chan sched.Job
 }
 
 type enqueueReq struct {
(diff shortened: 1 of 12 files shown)


sched/queue/queue.go
```diff
@@ -0,0 +1,40 @@
+package queue
+
+import (
+	"github.com/scootdev/scoot/sched"
+	"time"
+)
+
+// Note: queued jobs can be read straight out of a channel that should
+// be returned along with the creator of a Queue
+type Queue interface {
```
CaitieM20: Looks Good!  Can you also add a ticket to add Removing Work from the Queue to the interface

sched/queue/memory/simple.go
```diff
@@ -0,0 +1,81 @@
...
+	resultCh chan enqueueResult
+}
+
+type enqueueResult struct {
+	jobID string
+	err   error
+}
+
+func (q *simpleQueue) Enqueue(job sched.Job) (string, error) {
+	if job.Id != "" {
```
CaitieM20: would prefer to have one method associated with messages that is IsValidJobId() like in Sagas.  We may end up with stricter rules as this system evolves

sched/queue/memory/simple.go
```diff
@@ -0,0 +1,81 @@
...
+		return "", queue.NewInvalidJobRequest(fmt.Sprintf("job.Id must be unset; was %v", job.Id))
```
CaitieM20: what does job.Id must be unset mean?  Would Invalid Job Id be a better more readable error

sched/queue/memory/simple.go
```diff
@@ -0,0 +1,81 @@
...
+	job      sched.Job
+	resultCh chan enqueueResult
+}
+
+type enqueueResult struct {
+	jobID string
+	err   error
+}
+
+func (q *simpleQueue) Enqueue(job sched.Job) (string, error) {
```
CaitieM20: Thought is the validation of a Job something we could provide as a function associated with Queue interface?  Seems like all Queue implementations will have to implement this

sched/queue/memory/simple.go
```diff
@@ -0,0 +1,81 @@
...
+	if job.Id != "" {
+		return "", queue.NewInvalidJobRequest(fmt.Sprintf("job.Id must be unset; was %v", job.Id))
+	}
+	if len(job.Tasks) == 0 {
+		return "", queue.NewInvalidJobRequest("job must have at least 1 task")
+	}
+	for _, task := range job.Tasks {
+		if task.Id == "" {
```
CaitieM20: would prefer IsValidTaskId() method

sched/queue/memory/simple.go
```diff
@@ -0,0 +1,81 @@
...
+			return "", queue.NewInvalidJobRequest("task.Id must be set")
```
CaitieM20: taskId is Not Valid is a more understandable error message.

sched/queue/memory/simple.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+		if len(task.Command) == 0 {
+			return "", queue.NewInvalidJobRequest("task.Commands must be set")
```
CaitieM20: A more specific error message like task.Commands must have at least one entry or something.  Set is very implementation specific to Golang, less the semantics of what we want.

sched/queue/memory/simple.go
```diff
@@ -0,0 +1,81 @@
...
+// In-memory Scheduler Queue
+
+import (
+	"fmt"
+	"github.com/scootdev/scoot/sched"
+	"github.com/scootdev/scoot/sched/queue"
+	"time"
+)
+
+func NewSimpleQueue() (queue.Queue, chan sched.Job) {
```
CaitieM20: Nit: using this to demo, maybe allowing the caller to specify the queue capacity would be nice.

sched/queue/queue.go
CaitieM20: I'm also super interested in the Dequeue part Peek() and Dequeue() because there has to be a two way handshake there, to prevent data loss during crashes
- scheduler give me work, 
- once's its scheduled work and started the saga return i successfully started this
- then its safe to dequeue

sched/queue/memory/simple.go
jasonbs10: TBH, I find this whole thing confusing. Do we need all these channels? Also, I think we'll need a peek() method so we don't drop jobs, and more detail about dequeu would be helpful.

sched/queue/queue.go
dbentley: Yeah, that makes sense.

Now the channel of work is of type queue.WorkItem, which includes a Job and a Dequeue method.

How's that?

sched/queue/memory/simple.go
dbentley: Channels look weird at first, but they replace a lot of more confusing code involving sync primitives. They also let client code look more familiary.

For instance, how would you know when to peek? Maybe peek waits, but then peek wouldn't be able to be used in a select statement.

Now Dequeue is a method on the item that's returned.

sched/queue/memory/simple.go
dbentley: Agreed. Migrated there (and also redid a lot of the error messages)

sched/queue/memory/simple.go
dbentley: Encapsulate in Queue package. (Or should this even just be moved to the sched package, where Job is defined?)

sched/queue/memory/simple.go
dbentley: WDYT of this?

sched/queue/memory/simple.go
dbentley: How's that?

sched/queue/memory/simple.go
dbentley: Sure, what were you thinking?

Maybe sched should define JobDefinition and TaskDefinition structs, that are different from Job and Task?

The downside is that they'll be almost the same, except for ID?

CaitieM20:
LGTM thanks for chatting through this, this morning!

PR #37, opened 2016-06-27
Jschiller/vendoring

jasonbs10:


Makefile
```diff
@@ -9,21 +9,22 @@ GO15VENDOREXPERIMENT := 1
...
+	# Checkout our vendored dependencies.
+	# Note: The submodule dependencies must be initialized prior to running scoot binaries.
+	#       When used as a library, the vendor folder will be empty by default (if 'go get'd). 
 	git submodule update --init --recursive
 
-    # Install mockgen binary (it's only referenced for code gen, not imported directly.)
-    # Both the binary and a mock checkout will be placed in $GOPATH (duplicating the vendor checkout.)
+	# Install mockgen binary (it's only referenced for code gen, not imported directly.)
+	# Both the binary and a mock checkout will be placed in $GOPATH (duplicating the vendor checkout.)
 	go get github.com/golang/mock/mockgen
```
dbentley: Do we want to vendor this, too?

vendor/github.com/BurntSushi/toml
```diff
@@ -0,0 +1 @@
+Subproject commit f0aeabca5a127c4078abb8c8d64298b147264b55
```
dbentley: Maybe I'm missing something... What determines these commits? Is it deps.sh?

.gitmodules
```diff
@@ -28,3 +28,72 @@
 [submodule "vendor/golang.org/x/sys"]
 	path = vendor/golang.org/x/sys
 	url = https://go.googlesource.com/sys
+[submodule "vendor/github.com/BurntSushi/toml"]
```
dbentley: Can we put comments in these files describing the annoyance?

jasonbs10: Yeah, it was auto generated but I believe I can add comments.

Makefile
jasonbs10: It is vendored, this is just to install the binary in $GOPATH/bin

vendor/github.com/BurntSushi/toml
jasonbs10: deps.sh creates a detached branch based on HEAD when it first adds the submodule.

dbentley: What updates this? (Or will?)

Makefile
dbentley: Then why do we need the go get? Can this just be go install?

vendor/github.com/BurntSushi/toml
jasonbs10: I intended to have updates happen manually once committed. It would be easy to script that though.

Makefile
jasonbs10: I think that's effectively the same thing, but perhaps signals intent better?

dbentley: I think we explicitly want this to *not* do any network access, which go get can do.

.gitmodules
```diff
@@ -28,3 +28,72 @@
...
+[submodule "vendor/github.com/magiconair/properties"]
+	path = vendor/github.com/magiconair/properties
+	url = https://github.com/magiconair/properties
+[submodule "vendor/github.com/mitchellh/mapstructure"]
+	path = vendor/github.com/mitchellh/mapstructure
+	url = https://github.com/mitchellh/mapstructure
+[submodule "vendor/github.com/russross/blackfriday"]
+	path = vendor/github.com/russross/blackfriday
+	url = https://github.com/russross/blackfriday
+[submodule "vendor/github.com/scootdev/scoot"]
```
dbentley: Oh, hey, we probably shouldn't vendor scoot, eh?

dbentley:
LGTM

(once you can get the build working)

Makefile
```diff
@@ -21,13 +21,14 @@ dependencies:
 
 	# Install mockgen binary (it's only referenced for code gen, not imported directly.)
 	# Both the binary and a mock checkout will be placed in $GOPATH (duplicating the vendor checkout.)
-	go install github.com/golang/mock/mockgen
+	# We use 'go get' here because 'go install' will not build out of our vendored mock repo.
+	go get github.com/golang/mock/mockgen
```
dbentley: So does this actually download it separately from our vendored mock repo?

Makefile
```diff
@@ -12,20 +12,22 @@ default:
...
+
 	# Checkout our vendored dependencies.
 	# Note: The submodule dependencies must be initialized prior to running scoot binaries.
 	#       When used as a library, the vendor folder will be empty by default (if 'go get'd). 
 	git submodule update --init --recursive
 
 	# Install mockgen binary (it's only referenced for code gen, not imported directly.)
 	# Both the binary and a mock checkout will be placed in $GOPATH (duplicating the vendor checkout.)
-	go get github.com/golang/mock/mockgen
+	go install github.com/golang/mock/mockgen
```
dbentley: Oh, nm, this is great.

PR #39, opened 2016-06-28
adding specific errors to saga log

CaitieM20:
Callers need to be able to distinguish between error types, so they know if its transient error i.e. log writes temporarily unavailable or illegal state update.  Test cases updated to check for specific error

Also practicing what I preach In [Simple Test Cases Can Prevent Most Critical Failures](https://www.usenix.org/system/files/conference/osdi14/osdi14-paper-yuan.pdf) one of the big discoveries that aborting on overly generic errors / poor error handling resulted in many catastrophic failures in distributed systems.


dbentley:
LGTM

PR #40, opened 2016-06-28
Ignore fuse when doing test coverage (may be added back later.)

jasonbs10:


CaitieM20:
LGTM!  Yay! was just going to ask if we could do this Thanks!

PR #43, opened 2016-06-30
sched/queue: add functionality to make clearer how channels would look

dbentley:


sched/definitions.go
```diff
@@ -17,3 +17,33 @@ type Task struct {
 	Command    []string // Argv to run
 	SnapshotId string
 }
+
+type JobState int
+
+const (
+	// An unambiguous 0-value.
+	UNKNOWN JobState = iota
```
CaitieM20: I don't think the way the system is designed we can tell between UNKNOWN and PENDING.  Sagas are only created and queryable once running

sched/definitions.go
```diff
@@ -17,3 +17,33 @@ type Task struct {
...
+	State JobState
+}
+
+func PendingStatus(jobID string) (s JobStatus) {
+	s.ID = jobID
+	s.State = PENDING
+	return s
+}
+
+func UnknownStatus(jobID string) (s JobStatus) {
```
CaitieM20: Should there be RunningStatus & DoneStatus functions.  
Should we prefix these methods with Make since they are Factory Methods?

sched/definitions.go
```diff
@@ -17,3 +17,33 @@ type Task struct {
...
+	UNKNOWN JobState = iota
+	// Waiting to run.
+	PENDING
+	// Running
+	RUNNING
+	// Done
+	DONE
+)
+
+type JobStatus struct {
```
CaitieM20: If the goal is to only allow folks to use the factory methods, should JobStatus be internal (jobStatus)

sched/queue/queue.go
```diff
@@ -12,6 +12,11 @@ type Queue interface {
 	// Enqueue enqueues a job and assigns it an ID. Errors may be
 	// an *InvalidJobRequest or a *CanNotScheduleNow
 	Enqueue(job sched.Job) (string, error)
+
+	Chan() chan WorkItem
+
+	Status(jobId string) (sched.JobStatus, error)
```
CaitieM20: Is this something we plan to implement on the Queue?  
Getting status from a durable queue is going to be expensive, requires reading the whole queue every time.

sched/definitions.go
jasonbs10: Wouldn't we just iterate over an array of jobs to determine if something is pending but not yet created? I guess, which part of the system design are you referring to?

sched/queue/memory/simple.go
```diff
@@ -48,33 +142,28 @@ func (q *simpleQueue) Enqueue(job sched.Job) (string, error) {
...
-		case q.outCh <- simpleWorkItem(job):
-			req.resultCh <- enqueueResult{id, nil}
-		default:
-			req.resultCh <- enqueueResult{"", queue.NewCanNotScheduleNow(1*time.Second, "queue full")}
-		}
-	}
+func (q *simpleQueue) Status(jobId string) (sched.JobStatus, error) {
+	resultCh := make(chan statusResult)
+	q.reqCh <- statusReq{jobId, resultCh}
+	result := <-resultCh
```
CaitieM20: why is the status request added to the queue?  You have enough information in the queueState to answer this immediately.  Either its already been queued (PENDING) or its not (UNKNOWN)

What benefit is their of the increased latency for this read being answered in order of requests being processed?

sched/queue/memory/simple.go
```diff
@@ -48,33 +142,28 @@ func (q *simpleQueue) Enqueue(job sched.Job) (string, error) {
...
-		q.nextID++
-		job.Id = id
-		select {
-		case q.outCh <- simpleWorkItem(job):
-			req.resultCh <- enqueueResult{id, nil}
-		default:
-			req.resultCh <- enqueueResult{"", queue.NewCanNotScheduleNow(1*time.Second, "queue full")}
-		}
-	}
+func (q *simpleQueue) Status(jobId string) (sched.JobStatus, error) {
```
jasonbs10: Just a thought, we can pull out the send request logic. This is uglier but pulls out doRequest() rather than copy-pasting core logic for each additional request type.
```
type request interface {
  resultCh() *(chan enqueueResult)
}

type req struct {
	resultCh chan enqueueResult
}
func (req *request) resultCh() *(chan enqueueResult) {
   return &resultCh
}

type enqueueReq struct {
        req
	job      sched.Job
}
type statusReq struct {
        req
	jobId    string
}

func (q * simpleQueue) doRequest(req request) interface{} {
	*req.resultCh := make(chan interface{})
	q.reqCh <- req
	return <-req.resultCh
}

func (q *simpleQueue) Enqueue(job sched.Job) (string, error) {
	err := queue.ValidateJob(job)
	if err != nil {
		return "", err
	}
        result := q.doRequest(enqueueReq{job, resultCh}).(*enqueueResult)
        return result.jobID, result.err
}
```

sched/queue/memory/simple.go
```diff
@@ -3,40 +3,134 @@ package memory
...
+	close(s.outCh)
+}
+
+func (s *simpleQueueState) done() bool {
+	// We are done once our inputs are closed and we have no items left to send
+	return s.reqCh == nil && len(s.items) == 0
 }
 
-type simpleWorkItem sched.Job
+func (s *simpleQueueState) iter() {
```
jasonbs10: All the import stuff is in here and everything else is pretty simple. That's good.

I'm not quite getting how this changes the lack of back pressure when something slow like durability is added. Also, it's a bit hard to reason about sending to a possibly nil outCh and why we set outCh whenever something is enqueued.

sched/queue/memory/simple.go
jasonbs10: I've been talking to Dan about the general design here and he's following the ''share by communicating' model whereas I worked up an alternative implementation which relies on locking shared data. So the answer to your question is that it's avoiding locking.

sched/definitions.go
dbentley: I did a poor job of commenting and that made the design unexplained.

I think that once a client has submitted to the Scoot Cloud Exec API, it needs the Cloud Exec API to be able to say what the status is. If jobs are invisible, that's an unworkable design, because clients can't make sense of it.

Fortunately, I think it's easy for to support this API!

sched/definitions.go
dbentley: There should be at some point, but the Queue will never deal with jobs in those status (because the Queue never runs or knows about done jobs). I can remove them.

That's a good point about should they be "MakeFoo". I guess I think it's kind of different because it's not JobStatus, it's RunningStatus, so we don't need the Make to differentiate the name from the struct type. I think it's nice to have the shorter method names, but I don't feel at all strongly.

sched/definitions.go
dbentley: I think people should use the factory methods to create, but want to just read the data. I wish there were immutable types, so that we could ensure clients don't modify the fields, but, well, go.

sched/queue/queue.go
dbentley: I think we can take advantage of our (small) size to know that we can keep the Queue in-memory. So the durable log is for durability, and we can recover, but we know that our server will have the queue in-memory. So this should still be an in-memory operation for all reasonable implementations, and therefore very fast.

Does that make sense?

sched/queue/memory/simple.go
dbentley: Hah! Yeah, this is a lot like one iteration I had (but actually saves a few more lines, at the cost of less type safety). I think there are other points on the spectrum, too.

But, ultimately, it felt like it saved a few lines of boilerplate, but led to less type-safety (with the casting). WDYT?

sched/queue/memory/simple.go
dbentley: Wait, this has back-pressure, right? Because it checks it against capacity? Or am I misunderstanding?

As for possibly-nil outCh: I found Dave Cheney's explanation of "Channel Axioms" very useful: http://dave.cheney.net/2014/03/19/channel-axioms

In this case, "A send to a nil channel blocks forever"

So, sometimes we don't want to send to outCh (if we have nothing to send, or if we're still waiting for a dequeue ping to come in). How do we do this?

One way: use if to pick which of several select statements to call. BUT, then the select statements have a lot of repeated code. So that's bad.

We could use reflect.Select, which takes a slice of SelectCase. But that leads to code that looks weird.

So, AIUI (and maybe I'm missing a simpler way), if there's some case in a select statement you want to do conditionally, the idiomatic way is to set it to nil (in which case the operation will block forever, and therefore never be selected).

sched/queue/memory/simple.go
jasonbs10: The more complex or long something is, the more sense it makes to pull out a common function. I don't know if this qualifies, but there is a transfer of control taking place and a single function to do that would be nice. Up to you really.

sched/queue/memory/simple.go
jasonbs10: If you have a bunch of requests waiting in L62: 'case req, ok := <-s.reqCh:' due to slow durability or such, then the 'items' array will not change responsively and will not accurately reflect the current load.

In terms of the nil channel, I figured it out but I had to convince myself that the logic was ok and ended up having to look up this particular behavior (well actually I just tried it out on the playground). This is the perfect place to comment that we continually reassign outCh to avoid an 'if' statement (or maybe check if null before assigning?) and that nil channel is ok initially because we have nothing to send on the first iteration and want it to fail/block.

sched/definitions.go
CaitieM20: So in the API design as I understand it,

If you have a JobId it means your Job was successfully scheduled, since we return this as part of accepting the work.  So if they have a JobId you know your job was scheduled successfully.   Why do we need this extra status?

sched/definitions.go
CaitieM20: I just meant make the struct type lower case.  that way they can't create on their own without using the factory method.  If you want to enforce immutability in the API you can make the struct components protected (lowercase) and expose accessor read only methods.

sched/queue/queue.go
CaitieM20: I think this introduces added complexity and consistency issues if we have to write to a durable queue and an in memory one.  The only reason to do this is if we have latency requirements aka use it as a cache.  I would prefer to just read off the durable log and not have both, less code and less potential bugs.  

I was under the impression that the in memory queue is just for testing

sched/definitions.go
dbentley: I imagine that I'd send the job, get back its ID (knowing it made it onto the queue), and then I'll have this ID. I'll want to sit there and poll its status, seeing it progress. It would be weird if I got the ID back, then as soon as I poll it's invisible.

An alternative would be to have the initial submission not return at all until the job has been pulled from the queue into the distributor and been persisted to a saga log. That would make this code unnecessary, but in order to give clients a good experience we'd need to make sure a job doesn't stay in the queue for more than... 200 milliseconds?

sched/queue/queue.go
dbentley: Absolutely the in-memory queue is not for production use (beyond the alpha).

sched/definitions.go
CaitieM20: I'm not suggesting returning in progress as soon as its placed on the Queue.  My suggestion is to remove UNKNOWN.  

PENDING - on the work queue not scheduled
RUNNING - scheduled has a Start Saga Log Entry
DONE - has a Start Saga Log Entry and an End Saga Log Entry

This eliminates the need for status on the work queue.  

I don't like the proposed alternative, we are implementing an async work model and this goes against this paradigm.   Also doing this would result in a lot of false negatives (failed to enqueue) if connections are broken or take longer than expected which can happen in failure modes.  We should not design the system to make assumptions that aren't based on worst case scenarios.  This will lead to every fragile solutions

sched/definitions.go
CaitieM20: I'm ok with either one.

sched/queue/memory/simple.go
CaitieM20: ok, see previous comment I still think we don't need a Status on the Queue

PR #45, opened 2016-07-01
Distributor V1

CaitieM20:
- PoolDistributor that schedules work on any free node available (essentially a node pool), current implementation assumes static cluster, will support dynamic cluster in next pull request.
- Removed Distributor Interface, I was making a bunch of modifications and I don't have a good common abstraction the more I think about it.  So deleting that code for now.
- Updated Demo to use PoolDistributor, and increased the number of Jobs we can run.  Prior to this scheduler around 1000+ jobs the demo would blow up because we would over schedule nodes.  
- Renamed ClusterMembership file to Cluster to match interface inside


sched/distributor/poolDistributor_test.go
```diff
@@ -0,0 +1,68 @@
...
+		t.Error("Expected dist to be nil when no members in Cluster")
+	}
+}
+
+func TestPoolDistributor_OneNodeCluster(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	nodes := cm.GenerateTestNodes(1)
+	clusterMock := cm.NewMockCluster(mockCtrl)
```
dbentley: I think this test would be cleaner with a fake implementation instead of a mock. It's not important that Members() is called exactly once.

Not a big deal, and I know we disagree, but just wanted to mention how we could save lines in tests by replacing these 6 lines with:

dist :=NewPoolDistributor(fakecluster.Make(1))

dbentley:
LGTM

sched/distributor/poolDistributor.go
```diff
@@ -0,0 +1,63 @@
...
+	freeCh chan cm.Node
+}
+
+//
+// Creates a Pool Distributor, assumes
+// all nodes in the cluster are free upon creation
+//
+// If the cluster has 0 nodes in it returns nil
+//
+func NewPoolDistributor(cluster cm.Cluster) *PoolDistributor {
```
jasonbs10: I should've asked earlier, but what do we gain by having the interface cm.Cluster vs just passing around []cm.Node? Are we planning to enhance cm.Cluster in any way?

sched/distributor/poolDistributor.go
```diff
@@ -0,0 +1,63 @@
...
+
+	return &PoolDistributor{
+		freeCh: freeCh,
+	}
+}
+
+//
+// Retrieve a Node to Do Work for this Particular Job
+//
+func (dist *PoolDistributor) ReserveNode(work sched.Job) cm.Node {
```
jasonbs10: Is this Reserve/Release interface sufficient past alpha? I imagine we'll eventually , want to query for a node with the best data locality, not sure that matters much now though.

sched/distributor/poolDistributor_test.go
jasonbs10: fwiw I think we should be judicious with our use of mocks. They're ok to have but I wouldn't want them to be our go-to-test-type.

jasonbs10:
lgtm

sched/distributor/poolDistributor.go
CaitieM20: Yes to some extent we will always need to say this node is no longer doing work.  Even if there are multiple tasks allowed to be scheduled on a node.  This code will get smarter over time, and I don't have the interface any more because this may evolve slightly and is pretty specific to individual schedulers.  I don't feel confident that I have the right abstraction yet

sched/distributor/poolDistributor.go
CaitieM20: Yes next code review cm.Cluster provides notifications of changes Add and Remove.  Easier to do incremental operations than having complex logic in every client to detect what's changed

sched/distributor/poolDistributor_test.go
CaitieM20: I don't really understand the pushback on Mocks.  

I personally would rather use generated code instead of maintaining hand written fakes.  Its less work and less code for us to maintain.  If the interface changes I just run generate and a new mock is generated, vs having to go modify a bunch of hand written code that is untested.  

What drawbacks do you see to using mocks? @dbentley @jasonbs10

jasonbs10: AIUI, mocks are used to test interaction with dependencies and stubs/fakes are used when we only care about the result. Here we're testing two things: that Members() is called and that that NewPoolDistributor() succeeds with >1 node in the cluster. If we want to refactor so that NewPoolDistributor() calls a new function AllMembers(), all the tests here have to change. With a fake, we only have to update the MakeFake() implementation, and the tests remain unchanged since they don't care what underlying method actually gets called.

If this isn't a convincing example, I'd like to know why so I can either change my views are explain it differently. Dan, what's the basis of your opinion?

The tests in this PR are simple enough that I don't think the type of test makes a huge difference, but that can change as our code grows.

CaitieM20: I agree, if the Interface fundamentally changes a lot of the test cases will break or need to be updated,  but I think this is the case even with a fake.  The calling code is only interacting with the interface so if the method it calls changes, the tests will need to be updated, which would also involve updating the fake.  I'd have to add the new method and have it return what I want.  

Also I'm fine with test cases being deleted and re-written if the implementation changes significantly.  I think this is healthy in an evolving code base, instead of trying to retrofit old test cases to support the modified code.  

I think you will see less Mock abuse in Go than other languages since you only can Mock interface, and not objects themselves.  Making a breaking change to an interface is going to cause a rather large refactor anyway including the Mocks.

PR #46, opened 2016-07-05
Added workerserver and workercl

jasonbs10:


binaries/workercl/main.go
```diff
@@ -0,0 +1,23 @@
...
+
+	client, err := client.NewClient(transportFactory, protocolFactory)
+	if err != nil {
+		log.Fatal("Cannot initialize Worker CL: ", err)
+	}
+	err = client.Exec()
+	if err != nil {
+		log.Fatal("error running workercl ", err)
+	}
+}
```
dbentley: (Leaving comment here because I can't figure out how to comment on a binary file)

What's binaries/workerserver/main?

scootapi/api.go
```diff
@@ -1,3 +1,5 @@
 package scootapi
 
-// We should use go generate to run thrift --gen go:package_prefix=github.com/scootdev/scoot/scootapi/gen-go/ scoot.thrift
+// We should use go generate to run:
+//   thrift --gen go:package_prefix=github.com/scootdev/scoot/scootapi/gen-go/ scoot.thrift
```
dbentley: I'm sorry, I think I may have forgotten to update this. Thrift has an option so I think you can just do:

//   thrift --gen go:package_prefix=github.com/scootdev/scoot/scootapi/gen-go/,thrift_import=github.com/apache/thrift/lib/go/thrift scoot.thrift

binaries/workercl/main.go
jasonbs10: That shouldn't be there, I almost made the same mistake in my phab review.
I'll delete it.

workerapi/server/server.go
```diff
@@ -0,0 +1,102 @@
...
+	processor := worker.NewWorkerProcessor(handler)
+	server := thrift.NewTSimpleServer4(processor, transport, transportFactory, protocolFactory)
+
+	fmt.Println("About to serve")
+
+	return server.Serve()
+}
+
+type Handler struct {
+	runs map[string]*worker.RunStatus
```
dbentley: Would it be easier to use the no-op runner that currently exists to do this?

(That would make this file just a translator between the Thrift objects and the domain objects, and then call the Runner for the business logic)

workerapi/server/server.go
```diff
@@ -0,0 +1,102 @@
...
+
+func makeStatus(status worker.Status, exitCode int32, info string) *worker.RunStatus {
+	rs := worker.NewRunStatus()
+	rs.Status = status
+	rs.Info = &info
+	rs.ExitCode = &exitCode
+	return rs
+}
+
+//TODO: integrate runner lib and remove all this temp test code.
```
dbentley: Oh, I see. Yeah, I'd think it might be easier to just integrate it now, but I don't want to block you.

workerapi/client/commands.go
```diff
@@ -0,0 +1,115 @@
+package client
+
+import (
+	"log"
+	"strings"
+
+	"github.com/luci/go-render/render"
```
dbentley: Hadn't seen this before; looks kinda nifty.

dbentley:
LGTM

CaitieM20:
Test cases would be nice :)

binaries/workerserver/main.go
```diff
@@ -0,0 +1,40 @@
...
+	protocolFactory := thrift.NewTBinaryProtocolFactoryDefault()
+	transportFactory := thrift.NewTTransportFactory()
+
+	err := server.Serve(server.NewHandler(), fmt.Sprintf(":%d", *thriftPort), transportFactory, protocolFactory)
+	if err != nil {
+		log.Fatal("Error serving Worker Server: ", err)
+	}
+}
+
+func handler(w http.ResponseWriter, r *http.Request) {
```
CaitieM20: nit if this handler is meant to be the healthcheck should we name it that?

jasonbs10:
Maybe you're right, but I was only planning to do them once I transitioned
from dummy impl to something more fleshed out.

CaitieM20:
Cool LGTM didn't realize this was more of a toy

dbentley:
Also, I think these packages are just the translation layers (thrift -> domain objects), and then the code in runner/ already has tests. Yay tests.

PR #47, opened 2016-07-05
Caitie/dynamic cluster distributor

CaitieM20:
Pool Distributor Supports a Dynamic Cluster

* Adding a Dynamic Cluster Interface that extends Cluster Interface. New DynamicCluster methods are expected to return a ClusterState object which provides initial nodes and a channel for updates.  This simplifies distributor logic, and avoids rebuilding the state of the cluster on every request, updated dynamicCluster implementation to support these new features.  
* Made LocalNode accessible outside the package
* Updated Demo to Use DynamicCluster, and remove/add nodes. The current dummy function is not all that random, it can definitely be improved
* PoolDistributor updates if initialized with a DynamicCluster.  Maintains a toRemove list.  On Retrieve or Release, ensures nodes on the toRemove list are not returned or re-added to the free list.  Once a node has been discarded its removed from the toRemove List.


sched/clustermembership/cluster.go
```diff
@@ -3,8 +3,37 @@ package cluster_membership
...
-	/*
-	 * Returns a Slice of Node Ids in the Cluster
-	 */
+	//
+	// Returns a Slice of Node in the Cluster
+	//
 	Members() []Node
 }
+
+type DynamicCluster interface {
```
dbentley: I'd recommend looking at this through the lens of the Interface Segregation Principle, which is the I in SOLID.

I think this is really two interfaces.

The Scheduler wants the updates.

Some other system is what will drive the AddNode and RemoveNode, so that can be a separate interface.

sched/clustermembership/cluster.go
```diff
@@ -3,8 +3,37 @@ package cluster_membership
...
+	Cluster
+
+	// Returns a Channel which contains Node Updates
+	NodeUpdates() <-chan NodeUpdate
```
dbentley: Why push this to the client? I think if I'm a client, I'd just like to get the nodes at the time I ask.

sched/clustermembership/cluster.go
CaitieM20: I mean sure but any implementing class has to implement all of them.  I'll split them out, because I don't want to go back and forth, but I think this is a rather picky comment.  Sometimes SOLID gets in the way of clean simple interfaces.

sched/clustermembership/cluster.go
CaitieM20: I want a Pub/Sub model of node updates for the client.  Also simple implementation

sched/clustermembership/dynamicCluster.go
```diff
@@ -1,49 +1,77 @@
...
- * Keeps a Map of the nodeId to Node for fast lookup of nodes.
- */
+//
+// Represents a Cluster of nodes that can change over time.
+// Keeps a Map of the nodeId to Node for fast lookup of nodes.
+//
 type dynamicCluster struct {
 	memberSet   map[string]bool
 	membersList []Node
+	updateCh    chan NodeUpdate
```
dbentley: Maybe have it be a slice of channels, and then iterate over them? Then you could have multiple watchers, and not have to have also the watcher variable?

sched/clustermembership/dynamicCluster.go
```diff
@@ -1,49 +1,77 @@
...
 func (c *dynamicCluster) AddNode(n Node) {
 
 	_, ok := c.memberSet[n.Id()]
 	if !ok {
 		c.membersList = append(c.membersList, n)
 		c.memberSet[n.Id()] = true
-	}
 
+		// don't add to channel and block unless someone is subscribed to updates
+		if c.watcher {
```
dbentley: Then this could just be a range over the slice of channels to send to

sched/clustermembership/cluster.go
dbentley: Yep, it is a picky comment, because I like your code and want it to be the best it can be.

If you think it's a bad suggestion, then skip it, don't block on it, and we can talk about it off the critical path. Or not.

You're right: every implementation will have to implement both sides.

I just want the consumer of the interface to not be confused by extra methods. But maybe there's only one consumer, and it's our core logic, so it's not concerning.

sched/clustermembership/dynamicCluster.go
CaitieM20: You mean an Add & Remove Channel?

I decided against that because we want Adds & Removes to be in order that they are received.  Otherwise you run into problems with the state of a flapping node. Add Node, Remove Node, Add Node, etc... You have races and a node that is flapping for a while could get lost unless the updates are in order

sched/clustermembership/cluster.go
dbentley: Oh, I think I see now, because you're maintaining the nodes in the scheduler as elements in a channel, so you can't just get all the nodes afresh each time.

sched/clustermembership/dynamicCluster.go
dbentley: No, I think you're right to have one channel handle all updates.

If you have multiple watchers, then each update would only be seen by one watcher.

Also, there might be a race condition where you ask for the members and then start listening, but miss updates? Maybe there should be one method that, atomically, gives you the current state and  a new channel?

dbentley:
LGTM to unblock

(But I think the watcher channel issues and race condition are worth addressing, even if only commenting)

sched/clustermembership/dynamicCluster.go
CaitieM20: Oh yes I understand, the idea is that there should only be one watcher per dynamic cluster.  I can update and add a comment.  Will combine the current state with updates as a return.  I agree unless the caller knows the implementation could result in missed updates

CaitieM20: Thanks for the suggestion, I like this much cleaner with the new refactor, and harder to misuse the API.

PR #48, opened 2016-07-06
runner/execer/os: an execer that actually runs an OS process

dbentley:


dbentley:
This isn't high-quality code yet, and certainly misses lots, but it's a nice first step. Next I'm going to go beef up the test suite and what not, and hook it up in daemon server.

jasonbs10:
lgtm

PR #49, opened 2016-07-06
runner: extend sim execer to support pausing, test runner w/it

dbentley:
Our simple runner rejects requests when a process is already running. Yay.


jasonbs10:
lgtm

PR #50, opened 2016-07-07
scheduler startup

CaitieM20:
Very Basic Scheduler, lots of TODOs on where we can improve etc... Wanted to share without test cases for now (coming tomorrow) since folks seemed interested in how this was progressing in sprint planning.  

* Scheduler object and methods is in charge of scheduling a Job.   Will return that it has successfully scheduled a job once a Start Saga Log entry has been made, and nodes reserved.  This gives us automatic backpressure (can't overschedule the cluster)
* Job Runner is in charge of running a job to completion and adding the Saga Log Entries.  
* Demo is updated to use the actual scheduler now, instead of the dumb demo code.  
* Removed Job parameter from ReserveNode in PoolDistributor not needed

The updateSagaState function in JobRunner definitely needs a refactor to share common error handling code.  Any Suggestions?  Go doesn't really support lambdas right?  Going to play with this tomorrow as well.


sched/scheduler/scheduler.go
```diff
@@ -0,0 +1,121 @@
...
+		numNodes := getNumNodes(job)
+		nodes := make([]cm.Node, 0, numNodes)
+		for i := 0; i < numNodes; i++ {
+			n := s.distributor.ReserveNode()
+			nodes = append(nodes, n)
+		}
+
+		// Start Running Job
+		s.wg.Add(1)
+		go func(job sched.Job, sagaState *saga.SagaState, nodes []cm.Node) {
```
jasonbs10: Might as well drop all the args and obtain them via closure since you're already using the closure for s.distributor.

sched/scheduler/scheduler.go
```diff
@@ -0,0 +1,121 @@
...
+	sagaState, err := s.saga.StartSaga(job.Id, nil)
+
+	// If we succssfully started the Saga, ProcssJob
+	if err == nil {
+
+		// Reserve Nodes to Schedule Job On
+		numNodes := getNumNodes(job)
+		nodes := make([]cm.Node, 0, numNodes)
+		for i := 0; i < numNodes; i++ {
+			n := s.distributor.ReserveNode()
```
jasonbs10: This blocks if we run out of nodes, shouldn't we error out and provide back pressure? (ReserveNode() impl could select{} on freeCh)

sched/server/main.go
```diff
@@ -0,0 +1,27 @@
+package main
```
jasonbs10: Let's move this (and sched demo?) to the top level binaries directory.

sched/scheduler/jobRunner.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+type stateUpdate struct {
+	msgType saga.SagaMessageType
+	taskId  string
+	data    []byte
+}
+
+// handles all the sagaState updates
+// TODO: lots of shared retry code, need either Lambdas(go support?) or some refactor to re-use shared code.
```
jasonbs10: You can pass around anonymous functions, they're the same as lambdas.

sched/scheduler/jobRunner.go
```diff
@@ -0,0 +1,267 @@
...
+	wg         sync.WaitGroup
+}
+
+func NewJobRunner(job sched.Job, saga saga.Saga, sagaState *saga.SagaState, nodes []cm.Node) jobRunner {
+	jr := jobRunner{
+		job:        job,
+		saga:       saga,
+		sagaState:  sagaState,
+		dist:       dist.NewPoolDistributor(cm.StaticClusterFactory(nodes)),
+		updateChan: make(chan stateUpdate),
```
jasonbs10: We probably want this to be buffered.

sched/scheduler/jobRunner.go
```diff
@@ -0,0 +1,267 @@
...
+			}
+
+			// if state transition was successful update internal state
+			if state != nil {
+				jr.sagaState = state
+
+				// state transition was not successful maxed out retries
+				// TODO: for now just panic eventually implement dead letter queue.
+			} else {
+				panic(fmt.Sprintf("Failed to succeesfully log StartTask. sagaId: %v, taskId: %v", jr.sagaState.SagaId(), update.taskId))
```
jasonbs10: We don't actually want to panic do we? If we exit the loop then runJob could fill up updateChan and then block forever. We'll want a different mechanism I think.

sched/scheduler/jobRunner.go
```diff
@@ -0,0 +1,267 @@
...
+						taskId:  task.Id,
+					}
+					jr.dist.ReleaseNode(node)
+					jr.wg.Done()
+				}(node, task)
+			}
+		}
+	}
+
+	if jr.sagaState.IsSagaAborted() {
```
jasonbs10: I'd make this an 'else' otherwise the intent is confusing. Like we start these task goroutines and then before any of them make progress/changes we check for IsSagaAborted() again?

sched/scheduler/jobRunner.go
```diff
@@ -0,0 +1,267 @@
...
+
+				jr.wg.Add(1)
+				go func(node cm.Node, task sched.Task) {
+
+					// TODO: After a number of attempts we should stop
+					// Trying to run a task, could be a poison pill
+					// Implement deadletter queue
+					taskExecuted := false
+					for !taskExecuted {
+						err := node.SendMessage(task)
```
jasonbs10: I'd add a TODO about actually waiting for the task to complete. Or are we intending for Node.SendMessage() to block until worker communicates that it's finished?

sched/scheduler/jobRunner.go
```diff
@@ -0,0 +1,267 @@
...
+	if jr.sagaState.IsSagaAborted() {
+		// TODO: we don't have a way to specify comp tasks yet
+		// Once we do they should be ran here.  Currently the
+		// scheduler only supports ForwardRecovery in Sagas so Panic!
+		panic("Rollback Recovery Not Supported Yet!")
+	}
+
+	// wait for all tasks to complete
+	jr.wg.Wait()
+	jr.updateChan <- stateUpdate{
```
jasonbs10: Should we wait for this to be committed before returning?

sched/scheduler/jobRunner.go
```diff
@@ -0,0 +1,267 @@
...
+						if err == nil {
+							taskExecuted = true
+						}
+					}
+
+					jr.updateChan <- stateUpdate{
+						msgType: saga.EndTask,
+						taskId:  task.Id,
+					}
+					jr.dist.ReleaseNode(node)
```
jasonbs10: nit: might make sense to defer these at the top of this goroutine.

sched/scheduler/jobRunner.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+	}
+}
+
+// checks the error returned my updating saga state.
+func retryableErr(err error) bool {
+
+	// InvalidSagaState is an unrecoverable error. This indicates a fatal bug in the code
+	// which is asking for an impossible transition.
+	_, ok := err.(saga.InvalidSagaStateError)
```
dbentley: A type switch might be shorter

sched/scheduler/jobRunner.go
```diff
@@ -0,0 +1,267 @@
...
+
+				// state transition was not successful maxed out retries
+				// TODO: for now just panic eventually implement dead letter queue.
+			} else {
+				panic(fmt.Sprintf("Failed to succeesfully log EndSaga. sagaId: %v", jr.sagaState.SagaId()))
+			}
+
+		case saga.AbortSaga:
+			attempts := 0
+			state, err := jr.saga.AbortSaga(jr.sagaState)
```
dbentley: Could saga just take the message to append to the saga log, and so you don't need a switch at all here? You can just send it directly to the saga log?

sched/scheduler/scheduler.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Blocks until all scheduled jobs are compeleted
+func (s *scheduler) BlockUnitlAllJobsCompleted() {
+	s.wg.Wait()
+}
+
+// Starts the scheduler, must be called before any other
+// methods on the scheduler can be called
+func (s *scheduler) Start() {
```
dbentley: Could this just be a private method that's called as part of NewScheduler?

sched/scheduler/jobRunner.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	// if the saga has not been aborted, just run each task in the saga
+	// TODO: this can be made smarter to run the next Best Task instead of just
+	// the next one in order etc....
+	if !jr.sagaState.IsSagaAborted() {
+		for _, task := range jr.job.Tasks {
+
+			if !jr.sagaState.IsTaskCompleted(task.Id) {
+				node := jr.dist.ReserveNode()
```
dbentley: Will this block until a node is available?

sched/scheduler/jobRunner.go
```diff
@@ -0,0 +1,267 @@
...
+		dist:       dist.NewPoolDistributor(cm.StaticClusterFactory(nodes)),
+		updateChan: make(chan stateUpdate),
+	}
+
+	go jr.updateSagaState()
+	return jr
+}
+
+// Runs the Job associated with this JobRunner to completion
+func (jr jobRunner) runJob() {
```
dbentley: This is great to get going.

This is the loop that I've been saying could be decomposed to be an explicit function at each tick. (It's easier to debug and test a function with explicit input and output than a long-running function that accretes state and writes updates).

Would it make sense for me to try that reorientation after you merge this, and you can see if you think it makes more sense that way?

sched/scheduler/scheduler.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// Log StartSaga Message
+	// TODO: need to serialize job into binary and pass in here
+	// so we can recover the job in case of failure
+	sagaState, err := s.saga.StartSaga(job.Id, nil)
+
+	// If we succssfully started the Saga, ProcssJob
+	if err == nil {
+
+		// Reserve Nodes to Schedule Job On
```
dbentley: Why reserve a priori? This seems like more work than just passing the entire pool down to the job runner.

sched/scheduler/jobRunner.go
```diff
@@ -0,0 +1,267 @@
...
+import (
+	"fmt"
+	"github.com/scootdev/scoot/saga"
+	"github.com/scootdev/scoot/sched"
+	cm "github.com/scootdev/scoot/sched/clustermembership"
+	dist "github.com/scootdev/scoot/sched/distributor"
+	"sync"
+)
+
+const MAX_RETRY = 3
```
dbentley: The retry is for writing to the saga log, right? (not rerunning the job if it fails)

I think we could leave this off here, and if we want retries, wrap the saga log in a RetryingSagaLog that retries as part of the saga log interface, which would make this code clearer because it's only doing one thing.

dbentley:
LGTM

sched/scheduler/scheduler.go
jasonbs10: I don't understand this comment. It makes sense to me to carve out some nodes so the job runner can work with a subpool (although it may make sense to move this into the goroutine below).

dbentley: If it's to limit the number of nodes it runs on, then we should just have the scheduler only reserve n nodes.

We should be able to reserve the node for only as long as we're running on it. Why have this extra step of "reserve it for job" as opposed to just "reserve it for task"? It's also an issue if, e.g., we reserve 5 nodes and then all 5 of those nodes disappear.

It seems easier to just hand the underlying job runner all of the nodes we have available, and each time it wants a node, it reserves it right then.

jasonbs10: This is too removed from real issues, and no matter the approach, this will be 'easy'.

For a motivating case, how about a job with 1 long running task and a bunch of short lived tasks. We'll be tying up multiple nodes doing nothing because we can't free our pool until everything is done. In that case, I would want to pass the pool down to job runner. The only issue with that is it becomes harder to provide back pressure.

dbentley: OHHHHH.

I see, the reason to reserve early is to provide back pressure. OK, cool, thanks. I totally wasn't getting that.

jasonbs10:
lgtm, but the requires a couple changes first kind.

sched/scheduler/scheduler.go
CaitieM20: Yes you are right, no loop so no scoping issues, I'm actually splitting this out into its own function for testability.  Easier to test a function does the right thing than an anonymous go routine

sched/scheduler/scheduler.go
CaitieM20: So the idea here was we want to block until a free node becomes available.  It won't return until it can schedule this job.  

The idea here is the code that glues the WorkQueue and the Scheduler together will only call ScheduleJob in a single thread, like in the demo.  If you can't successfully schedule a job because there aren't enough nodes we pause on accepting jobs.

sched/scheduler/jobRunner.go
CaitieM20: I want this to be done in order to ensure saga log messages are logged before I start a task.

i.e. StartTask must be logged before task is executed to ensure correctness of the protocol in recovery.  otherwise you could pass the update (it doesn't get logged) then StartTask and then a failure occurs so the message never gets logged.  I made the buffer declaration explicit

sched/scheduler/jobRunner.go
CaitieM20: Yes for now, eventually we will have the deadletter queue for updates that fail to log to the SagaLog for MaxRetry number of times.  I made the code more explicit with a TODO where this will fit in once we have durable storage.

If its an unrecoverable error something is very wrong (a really bad bug in our code most likely) Its not recoverable.  So i think Panic is the right thing to do here.

sched/scheduler/jobRunner.go
CaitieM20: sure fixed!

sched/scheduler/jobRunner.go
CaitieM20: The current API and localNode implementation block until the message has completed processing.    The idea here is it would just block, once it returns a response (right now success or err) we know its completed and can log EndTask

sched/scheduler/jobRunner.go
CaitieM20: Yes actually all messages should be acked as durably logged before proceeding.  Will Fix!

sched/scheduler/jobRunner.go
CaitieM20: Done

sched/scheduler/jobRunner.go
CaitieM20: Done! didn't realize Go supported this, awesome

sched/scheduler/jobRunner.go
CaitieM20: yes that's the idea, that way we don't overschedule the cluster

sched/scheduler/scheduler.go
CaitieM20: yup done

sched/scheduler/scheduler.go
CaitieM20: The reason to reserve early is to provide BackPressure.  Also by having a pool of nodes that we run all tasks in a Job on we easily get data locality for all those tasks.

I suspect this code will undergo several revisions to optimize scheduling over time.  But the first goals of 
* Backpressure
* Provide DataLocality within a job.

sched/scheduler/jobRunner.go
CaitieM20: great thanks refactored!

sched/scheduler/jobRunner.go
CaitieM20: sure removed, probably overkill and easier to control retires closer to the actual point of writing.

sched/server/main.go
CaitieM20: cool done

sched/scheduler/jobRunner.go
CaitieM20: sure

sched/scheduler/jobRunner.go
CaitieM20: So after more thought I think this logic should actually be moved into the Saga class, each consumer will basically have to write this code so it should just be hidden behind the main API.  I'm working on a refactor in another branch that will come next.

dbentley: SGTM.

PR #51, opened 2016-07-08
sched: connect work queue to apiserver

dbentley:


dbentley:
This is starting to look weird because of out-of-date names.

Here are a few things I could do to make it clearer, but it would make merging hard if others have in-flight changes:
1) make a cloud/ directory
2) scootapi/ -> cloud/api/
3) s/Proc/CloudScoot/ in thrift
4) What else?

I want to clean up, but I also don't want to needlessly disrupt others. Should I do this now, or do (any of) you want to do it?

jasonbs10:
I created a branch for it but got distracted, you should definitely do it since you're currently working on the things to be renamed (I'm working in workerserver/workerapi which should remain the same).

binaries/apiserver/main.go
```diff
@@ -10,7 +11,9 @@ func main() {
 	protocolFactory := thrift.NewTBinaryProtocolFactoryDefault()
 	transportFactory := thrift.NewTTransportFactory()
 
-	err := server.Serve(server.NewHandler(), "localhost:9090", transportFactory, protocolFactory)
+	queue, _ := memory.NewSimpleQueue()
```
CaitieM20: recommend adding a TODO here to swap out with durable implementation.  I've been doing that in the code, so it'll be easy to ensure we don't miss anything when we move to MVP

binaries/apiserver/main.go
```diff
@@ -10,7 +11,9 @@ func main() {
...
+	handler := server.NewHandler(queue)
+	err := server.Serve(handler, "localhost:9090", transportFactory, protocolFactory)
```
CaitieM20: Should loalhost:9090 be hardcoded.  For instance when this is actually ran in aurora, we don't get to pick the port.  Suggest maybe having a variable for now with a TODO to read from config.

scootapi/scoot.thrift
```diff
@@ -31,8 +35,9 @@ struct JobDefinition {
   2: optional JobType jobType,
 }
 
+// TODO(dbentley): rename this to... Exec? CloudExec? CloudScoot?
 service Proc {
-  Job RunJob(1: JobDefinition job) throws (
+   JobId RunJob(1: JobDefinition job) throws (
```
CaitieM20: why are we not returning the whole job anymore? I thought the idea was that we'd send back the Job with assigned JobIds and TaskIds.

scootapi/server/serve.go
```diff
@@ -20,16 +23,70 @@ func Serve(handler scoot.Proc, addr string, transportFactory thrift.TTransportFa
...
+		return nil, err
+	}
+
+	r := scoot.NewJobId()
+	r.ID = id
+	return r, nil
+}
+
+// Translates thrift job definition message to scoot domain object
+func thriftJobToScoot(def *scoot.JobDefinition) (result sched.Job, err error) {
```
CaitieM20: this method would be really good to use property based testing on.  To ensure a variety of supplied jobs get converted correctly.

scootapi/server/serve.go
```diff
@@ -20,16 +23,70 @@ func Serve(handler scoot.Proc, addr string, transportFactory thrift.TTransportFa
...
+}
 
-func NewHandler() scoot.Proc {
-	return &Handler{}
+func NewHandler(queue queue.Queue) scoot.Proc {
+	return &Handler{queue}
 }
 
-func (h *Handler) RunJob(def *scoot.JobDefinition) (*scoot.Job, error) {
+func (h *Handler) RunJob(def *scoot.JobDefinition) (*scoot.JobId, error) {
```
CaitieM20: Test cases please.  Using a Mock of the Queue interface makes testing the error cases really easy.

scootapi/scoot.thrift
dbentley: All of the job info is about to disappear from the API (as it sits in the queue). I thought it would be more consistent to return just the ID.

TaskIds: I don't understand why task IDs are optional in job definition. I see the value of them being:
mandatory (the client has to fill them out)
not allowed in a definition (the server fills them out)
absent entirely (just use the index into the list of tasks).

binaries/apiserver/main.go
dbentley: Probably not, but note that this isn't the binary we run in aurora. For that, we have a binary that mixes in twitter-specific logic.

binaries/apiserver/main.go
dbentley: Done.

scootapi/server/serve.go
dbentley: Added test cases using the in-memory implementation.

scootapi/server/serve.go
dbentley: Done.

scootapi/scoot.thrift
CaitieM20: We either need the clients to assign unique taskIds per job.  Which is fine, but it needs to be part of the API, so please update the Thrift.

We should not rely on order in the list.  That makes life much more difficult because then we have to preserve the order throughout the entire codebase.  From experience this is a pain.  A map of Id to Task info is what we want.

scootapi/server/serve_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// Empty queue
+	item := <-itemCh
+	item.Dequeue()
+
+	_, err = handler.RunJob(jobDef)
+	if err != nil {
+		t.Fatalf("can't enqueue after emptying: %v", err)
+	}
+}
```
CaitieM20: Need a test case for Queue returning an error because its unavailable.  

Obviously this will never be the case with the in-memory one, but requests to the durable queue could take too long (timeout) or be rejected because the queue is unavailable

scootapi/server/serve.go
CaitieM20: Awesome!

dbentley:
OK, I refactored the Thrift and the corresponding objects in sched. How does it look?

(This rendered the property test totally broken, so I'm fixing it now)

scootapi/server/serve_test.go
dbentley: Done.

scootapi/scoot.thrift
dbentley: OK, updated. WDYT?

CaitieM20:
LGTM

PR #52, opened 2016-07-09
Refactor Saga to allow for thread safe updates

CaitieM20:
Refactor Saga into SagaCoordinator and Saga.  

Saga controls access to a single SagaState object.  It provides the same API but allows for concurrent updates, i.e. when two tasks start or finish at the same time.  This class provides Read Your Write Consistency.  Once a write successfully completes and returns calling saga.GetState() will reflect the updates just made by the write.

SagaCoordinator class contains basic functions for creating, recovering Sagas that may be in progress.  This also disambiguates a single Saga from general functions on the log.   

Updates Scheduler code to use the new thread-safe saga instead of requiring all consumers of sagas to implement their own thread safe guarantees.

### DIFF
--- a/saga/saga.go
+++ b/saga/saga.go
@@ -12,7 +12,7 @@ type Saga struct {
 	log      SagaLog
 	state    *SagaState
 	updateCh chan sagaUpdate
-	mutex    *sync.RWMutex
+	mutex    sync.RWMutex
 }
 
 // Start a New Saga.  Logs a Start Saga Message to the SagaLog
@@ -36,7 +36,7 @@ func newSaga(sagaId string, job []byte, log SagaLog) (*Saga, error) {
 		log:      log,
 		state:    state,
 		updateCh: updateCh,
-		mutex:    &sync.RWMutex{},
+		mutex:    sync.RWMutex{},
 	}
 
 	go s.updateSagaStateLoop()
@@ -53,7 +53,7 @@ func rehydrateSaga(sagaId string, state *SagaState, log SagaLog) *Saga {
 		log:      log,
 		state:    state,
 		updateCh: updateCh,
-		mutex:    &sync.RWMutex{},
+		mutex:    sync.RWMutex{},
 	}
 
 	if !state.IsSagaCompleted() {
@@ -144,28 +144,6 @@ func (s *Saga) EndCompensatingTask(taskId string, results []byte) error {
 	return s.updateSagaState(MakeEndCompTaskMessage(s.id, taskId, results))
 }
 
-// updateSagaStateLoop that is executed inside of a single go routine.  There
-// is one per saga currently executing.  This ensures all updates are applied
-// in order to a saga.  Also controls access to the SagaState so its is
-// updated in a thread safe manner
-func (s *Saga) updateSagaStateLoop() {
-	for update := range s.updateCh {
-		s.mutex.Lock()
-		currState := s.state
-		newState, err := logMessage(currState, update.msg, s.log)
-
-		if err == nil {
-			s.state = newState
-
-			update.resultCh <- nil
-
-		} else {
-			update.resultCh <- err
-		}
-		s.mutex.Unlock()
-	}
-}
-
 // adds a message for updateSagaStateLoop to execute to the channel for the
 // specified saga.  blocks until the message has been applied
 func (s *Saga) updateSagaState(msg sagaMessage) error {
@@ -184,6 +162,32 @@ func (s *Saga) updateSagaState(msg sagaMessage) error {
 	}
 
 	return result
+}
+
+// updateSagaStateLoop that is executed inside of a single go routine.  There
+// is one per saga currently executing.  This ensures all updates are applied
+// in order to a saga.  Also controls access to the SagaState so its is
+// updated in a thread safe manner
+func (s *Saga) updateSagaStateLoop() {
+	for update := range s.updateCh {
+		s.itr(update)
+	}
+}
+
+func (s *Saga) itr(update sagaUpdate) {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	currState := s.state
+	newState, err := logMessage(currState, update.msg, s.log)
+
+	if err == nil {
+		s.state = newState
+
+		update.resultCh <- nil
+
+	} else {
+		update.resultCh <- err
+	}
 }
 
 type sagaUpdate struct {
--- a/saga/saga.go
+++ b/saga/saga.go
@@ -1,146 +1,201 @@
 package saga
 
-/*
- * Saga Object which provides all Saga Functionality
- * Implementations of SagaLog should provide a factory method
- * which returns a saga based on its implementation.
- */
+import (
+	"sync"
+)
+
+// Concurrent Object Representing a Saga
+// Methods update the state of the saga or
+// Provide access to the Current State
 type Saga struct {
-	log SagaLog
+	id       string
+	log      SagaLog
+	state    *SagaState
+	updateCh chan sagaUpdate
+	mutex    *sync.RWMutex
 }
 
-/*
- * Make a Saga which uses the specied SagaLog interface for durable storage
- */
-func MakeSaga(log SagaLog) Saga {
-	return Saga{
-		log: log,
-	}
-}
+// Start a New Saga.  Logs a Start Saga Message to the SagaLog
+// returns a Saga, or an error if one occurs
+func newSaga(sagaId string, job []byte, log SagaLog) (*Saga, error) {
 
-/*
- * Start Saga. Logs Message message to the log.
- * Returns the resulting SagaState or an error if it fails.
- */
-func (s Saga) StartSaga(sagaId string, job []byte) (*SagaState, error) {
-
-	//Create new SagaState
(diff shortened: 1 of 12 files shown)


saga/saga.go
```diff
@@ -1,146 +1,197 @@
...
+			if update.msg.msgType == EndSaga {
+				close(s.updateCh)
+			}
+
+			update.resultCh <- nil
+
+		} else {
+			update.resultCh <- err
+		}
+		s.mutex.Unlock()
```
dbentley: defer the unlock from above?

sched/scheduler/jobRunner.go
```diff
@@ -81,111 +71,26 @@ func (jr jobRunner) runJob() {
...
-	// InternalLogError is a transient error experienced by the log.  It was not
-	// able to durably store the message but it is ok to retry.
-	case saga.InternalLogError:
-		return true
-
-	// unknown error, default to retryable.
-	default:
-		return true
+func handleSagaLogErrors(err error) {
+	if s.FatalErr(err) {
```
dbentley: should this test be reversed?

saga/saga.go
```diff
@@ -1,146 +1,197 @@
...
+func (s *Saga) updateSagaStateLoop() {
+	for update := range s.updateCh {
+		s.mutex.Lock()
+		currState := s.state
+		newState, err := logMessage(currState, update.msg, s.log)
+
+		if err == nil {
+			s.state = newState
+
+			if update.msg.msgType == EndSaga {
```
dbentley: It's more idiomatic to have the sender close, instead of the receiver. Maybe move to EndSaga, above?

saga/saga.go
```diff
@@ -1,146 +1,197 @@
...
 
-	return s.logMessage(state, MakeAbortSagaMessage(state.sagaId))
+//
+// Log an AbortSaga message.  This indicates that the
+// Saga has failed and all execution should be stopped
+// and compensating transactions should be applied.
+//
+// Returns an error if it fails
+//
+func (s *Saga) AbortSaga() error {
```
dbentley: Should this close, too? Are you allowed to log more messages after abort'ing?

dbentley:
LGTM

This looks good for now, and I think makes this stage simpler. I worry about it long-term, because it seems like it could lead to race conditions.

start task on node A
cluster tells us node A died
start task on node B

And then node A and node B return concurrently.

(To add to the complexity, imagine that this is the last task, so when it finishes, we close the saga, which makes the slower result into a panic)

saga/saga.go
CaitieM20: I tried that, and all the test cases and demo code deadlock.  Does defer release after each loop iteration.  Doesn't appear to

dbentley: Oh, duh. Yeah.

I'd probably refactor to make each iteration just call a function, that then is able to defer the unlock. This will save us a deadlock if, e.g., logMessage panics.

But it's also more code and functions, so I understand if you'd prefer not to.

sched/scheduler/jobRunner.go
CaitieM20: yup good catch fixed!

saga/saga.go
dbentley: (I was reading this code and answered my own question: this should not close. After you write an abort message, you have to start writing begin comp task messages, and then end comp task messages, and only after all that the close. So yes, you have to log more messages after abort'ing.)

saga/saga.go
CaitieM20: cool done!

CaitieM20:
I dont think there is a race condition here.  Logging two EndTask messages is fine.  The last writer will win.  If the scheduler logs an EndTask message after EndSaga that's a bug in the scheduler.  Making Saga Messages Thread is to allow tasks to be ran in parallel.  StartSaga and EndSaga still need to be the first message and the last message.

The scenario you described above is handled and does not cause a panic in the current scheduler code.

Log StartSaga
Log StartTask1 
Assign Task1 to NodeA
Node A fails in the takes to long way ( currently the code panics here since everything is in memory )

// We don't currently do this but can
Reschedule Task1 on NodeB
Node B Completes task 
Log EndTask1 message
Log EndSaga message

Now NodeA responds that it finished Task1.  Nothing is waiting for this response so it'll just get dropped on the floor.  Ideally using leases would allow NodeA to determine that no one is waiting and just stop running the job.

sched/scheduler/scheduler.go
```diff
@@ -65,24 +65,20 @@ func (s *scheduler) ScheduleJob(job sched.Job) error {
 
 		// Start Running Job
 		s.wg.Add(1)
-		go s.runJob(job, sagaState, nodes)
+		go func() {
+			defer s.wg.Done()
```
jasonbs10: What is the intended behavior for BlockUnitlAllJobsCompleted()? We currently allow jobs to start even after that function returns and I'm not sure that's useful.

saga/saga.go
```diff
@@ -1,146 +1,200 @@
...
-	}
+// Concurrent Object Representing a Saga
+// Methods update the state of the saga or
+// Provide access to the Current State
+type Saga struct {
+	id       string
+	log      SagaLog
+	state    *SagaState
+	updateCh chan sagaUpdate
+	mutex    *sync.RWMutex
```
jasonbs10: Does this need to be a pointer?

saga/sagaCoordinator.go
```diff
@@ -0,0 +1,79 @@
...
+// Make a Saga which uses the specied SagaLog interface for durable storage
+//
+func MakeSagaCoordinator(log SagaLog) SagaCoordinator {
+	return SagaCoordinator{
+		log: log,
+	}
+}
+
+// Make a Saga add it to the SagaCoordinator, if a Saga Already exists
+// with the same id, it will overwrite the already existing one.
```
jasonbs10: Can we flesh out the overwriting part of this and why that's ok vs just panicking? Doesn't that mean there could be multiple goroutines running different jobs one succeeds, the other fails and can't do rollback because the saga is already done?

saga/sagaCoordinator.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+//
+// Recovers SagaState by reading all logged messages from the log.
+// Utilizes the specified recoveryType to determine if Saga needs to be
+// Aborted or can proceed safely.
+//
+// Returns the current SagaState
+//
+func (sc SagaCoordinator) RecoverSagaState(sagaId string, recoveryType SagaRecoveryType) (*Saga, error) {
```
jasonbs10: Does it make sense to say that Recover can only be called once per sagaId? Or is that too defensive-coding/

jasonbs10:
lgtm. I had a question similar to Dan's race condition question but without the race component (I'm not sure what level of defensive coding we want to practice so some of my questions may be irrelevant.)

CaitieM20:
Right the Saga Library will not let you do something incorrect to get in a
bad state (aka return an invalid error or panic), but you do have to use it
correctly.

I agree the panic is kind of ugly as part of the API, an error would be
better since it wouldn't crash someone's programs.

I think we could do some cleanup on the Saga to return an error which is
nicer, but involves making the code a bit more complicated.

saga/saga.go
CaitieM20: done

sched/scheduler/scheduler.go
CaitieM20: its just for testing / demo.  Since there is async work being done we want to wait before we verify state.  I updated the comment saying this is for testing only

saga/saga.go
CaitieM20: nope updated, I think I was following an example from somewhere, but can't remember.  Go Sync doesn't specify how to use it so change to non pointer.  

The Go Programming book shows it as not a pointer so looks like that's more idiomatic.  Changed

saga/sagaCoordinator.go
CaitieM20: Oh sorry this comment is actually out of date.  I was originally exploring the idea of keeping a map of all the inprogress Sagas in the SagaCoordinator.  But that made the code more complicated and I couldn't think of a good reason why the SagaCoordinator needs to keep track of each running Saga right now (obviously we can add this later)  So the comment is form that context.  Fixed comment

What will actually happen is each Saga object will have its own channel to do updates. The SagaLog will have multiple entries

StartSaga
StartTask1
StartSaga
StartTask1
EndTask1
EndSaga
EndTask1
EndSaga

But that's ok, Recovery and reading state from the Saga Log handles this and will handle this correctly.  We can have this scenario if two schedulers run the same job (Aurora can sometimes have ghost instances during leader election).

saga/sagaCoordinator.go
CaitieM20: It should only be called once logically, but I don't think anything bad will happen, besides potentially running a Job twice.  Like explained in the MakeSaga comment above

CaitieM20:
Yea I kind of view the EndSaga then log a message and Panic akin to closing a channel and then adding a message causes a Panic.  You aren't supposed to do it, and it puts the Saga in a bad state, so  we Panic so that you catch it very quickly in development.

PR #53, opened 2016-07-11
Initial impl of servable stats.

jasonbs10:
Initial impl of servable stats. Adds StatsReceiver type, using go-metrics under the hood, with the flexibility to control stat semantics/rendering. Currently produces the following output:

{
  "workerserver/handler/run#": 2,
  "workerserver/handler/run%3": 2,
  "workerserver/handler/runLatency_ms.avg": 105.9839915,
  "workerserver/handler/runLatency_ms.count": 2,
  "workerserver/handler/runLatency_ms.max": 110,
  "workerserver/handler/runLatency_ms.min": 101,
  "workerserver/handler/runLatency_ms.p50": 105.9839915,
  "workerserver/handler/runLatency_ms.p90": 110.317428,
  "workerserver/handler/runLatency_ms.p95": 110.317428,
  "workerserver/handler/runLatency_ms.p99": 110.317428,
  "workerserver/handler/runLatency_ms.p999": 110.317428,
  "workerserver/handler/runLatency_ms.p9999": 110.317428,
  "workerserver/handler/runLatency_ms.sum": 211
}

### DIFF
--- a/binaries/workerserver/main.go
+++ b/binaries/workerserver/main.go
@@ -4,9 +4,11 @@ import (
 	"flag"
 	"fmt"
 	"log"
-	"net/http"
+	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/scootdev/scoot/common/endpoints"
+	"github.com/scootdev/scoot/common/stats"
 	"github.com/scootdev/scoot/workerapi/server"
 )
 
@@ -15,26 +17,18 @@ var httpPort = flag.Int("http_port", 9091, "port to serve http on")
 
 func main() {
 	flag.Parse()
-	go serveHTTP()
+	stat, _ := stats.NewCustomStatsReceiver(stats.NewFinagleStatsRegistry, 15*time.Second)
+	stat = stat.Precision(time.Millisecond)
+	endpoints.RegisterStats("/admin/metrics.json", stat)
+	endpoints.RegisterHealthCheck("/")
+	go endpoints.Serve(fmt.Sprintf(":%d", *httpPort))
 
 	protocolFactory := thrift.NewTBinaryProtocolFactoryDefault()
 	transportFactory := thrift.NewTTransportFactory()
 
-	err := server.Serve(server.NewHandler(), fmt.Sprintf(":%d", *thriftPort), transportFactory, protocolFactory)
+	handler := server.NewHandler(stat.Scope("workerserver"))
+	err := server.Serve(handler, fmt.Sprintf(":%d", *thriftPort), transportFactory, protocolFactory)
 	if err != nil {
 		log.Fatal("Error serving Worker Server: ", err)
 	}
-}
-
-func handler(w http.ResponseWriter, r *http.Request) {
-	log.Println("Handling!", r.URL.Path)
-	// TODO(dbentley): implement the actual health-checking protocol.
-	fmt.Fprintf(w, "OK (health check in golang for %v )", r.URL.Path)
-}
-
-func serveHTTP() {
-	log.Println("Serving http on", *httpPort)
-	http.HandleFunc("/", handler)
-	err := http.ListenAndServe(fmt.Sprintf(":%d", *httpPort), nil)
-	log.Println("Done serving http", err)
 }
--- a/common/stats/stats.go
+++ b/common/stats/stats.go
@@ -131,6 +131,7 @@ func DefaultStatsReceiver() StatsReceiver {
 // Like DefaultStatsReceiver() but latched interval is made explicit.
 // Starts a goroutine that periodically captures all and clears select instruments.
 // Note: setting latched to <=0 will disable latching so rendering/resetting is on demand.
+// Note: it is up the main app to prevent calls to Render() after canceling the latched receiver.
 func NewLatchedStatsReceiver(latched time.Duration) (stat StatsReceiver, cancelFn func()) {
 	return NewCustomStatsReceiver(nil, latched)
 }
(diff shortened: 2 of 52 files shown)


common/stats/stats.go
```diff
@@ -0,0 +1,479 @@
+// This package borrows from go-metrics so we can avoid exposing dependencies,
+// and to give callers more flexibility in recording and formatting scoped stats.
```
dbentley: What are the changes?

Do we still need to import go-metrics if we use this?

workerapi/server/server.go
```diff
@@ -51,13 +59,22 @@ func makeStatus(status worker.Status, exitCode int32, info string) *worker.RunSt
...
+	time.Sleep(time.Millisecond * time.Duration(100+(rand.Int()%50)))
+	dummyGaugeVal++
+
 	numRuns := len(h.runs)
 	runId := strconv.Itoa(numRuns)
 	prevRunId := strconv.Itoa(numRuns - 1)
 
 	log.Println("Running", render.Render(cmd), runId)
 	if numRuns > 0 && h.runs[prevRunId].Status == worker.Status_RUNNING {
+		h.stat.Counter("alreadyRunning").Inc(1)
```
dbentley: Sweet.

dbentley:
LGTM

common/stats/stats.go
jasonbs10: carealot had 2k+ loc to mimic twitter stats, this is about 1/4 of that and basically a full rewrite.

We still use go-metrics for the default registry and instruments, but that isn't exposed to the user. User can go with the defaults or override the registry impl and/or individual instrument impls.

binaries/workerserver/main.go
```diff
@@ -15,26 +17,18 @@ var httpPort = flag.Int("http_port", 9091, "port to serve http on")
...
+	stat = stat.Precision(time.Millisecond)
+	endpoints.RegisterStats("/admin/metrics.json", stat)
+	endpoints.RegisterHealthCheck("/")
+	go endpoints.Serve(fmt.Sprintf(":%d", *httpPort))
 
 	protocolFactory := thrift.NewTBinaryProtocolFactoryDefault()
 	transportFactory := thrift.NewTTransportFactory()
 
-	err := server.Serve(server.NewHandler(), fmt.Sprintf(":%d", *thriftPort), transportFactory, protocolFactory)
+	handler := server.NewHandler(stat.Scope("workerserver"))
```
CaitieM20: You don't have to prefix stats with the service name, some services do some don't.  But just fyi absorber exports (service, source(host), metric name) so we already have this info.

common/stats/stats.go
```diff
@@ -0,0 +1,479 @@
...
+// For testing.
+var Time StatsTime = DefaultStatsTime()
+
+// Stats users can either reference this global receiver or construct their own.
+var CurrentStatsReceiver StatsReceiver = NilStatsReceiver()
+
+// Overridable instrument creation.
+var NewCounter func() Counter = newMetricCounter
+var NewGauge func() Gauge = newMetricGauge
+var NewGaugeFloat func() GaugeFloat = newMetricGaugeFloat
```
CaitieM20: what's the difference between Gauge & GaugeFloat?  The Twitter Observability Stack interprets everything as a double just FYI

CaitieM20: Do we need both?

binaries/workerserver/main.go
jasonbs10: Good to know. I like having it to copy and paste the json output, but
that's not so useful when we switch to viz.

On Mon, Jul 11, 2016 at 12:32 PM, Caitie McCaffrey <notifications@github.com

common/stats/stats.go
jasonbs10: I only did both since go-metrics provides both. In terms of usefulness, I
guess it mostly signals intent (I don't think we care about the
magnitude/precision tradeoff and we're not doing math with it). I'd keep it
by default but can remove it if Dan agrees.

On Mon, Jul 11, 2016 at 12:37 PM, Caitie McCaffrey <notifications@github.com

common/stats/stats.go
```diff
@@ -0,0 +1,479 @@
...
+}
+
+//
+// DefaultStats is a small wrapper around a go-metrics like registry.
+// Uses defaultStatsRegistry and sets latched duration to zero.
+// Note: a <=0 latch means that the stats are reset on every call to Render().
+func DefaultStatsReceiver() StatsReceiver {
+	stat, _ := NewCustomStatsReceiver(nil, 0)
+	return stat
+}
```
CaitieM20: So one of the nice things in the Twitter Metrics lib is NullStatsReceiver, which is basically like a write to dev/null, nice for testing, because you end up passing stats receivers around in methods and constructors.  

Might be worth adding if this implementation requires additional setup or mocks to work correctly

jasonbs10: Wait, isn't that taken care of by NilStatsReceiver below?

common/stats/stats.go
dbentley: You two are way ahead of me in this conversation. I'm reading it but have frightfully little to add (other than "can you explain this to someone who hasn't thought about it much?")

common/stats/stats.go
CaitieM20: And you already did this :)

common/stats/stats_time.go
```diff
@@ -0,0 +1,39 @@
+package stats
+
+import (
+	"time"
+)
+
+// wraps the stdlib time.Ticker struct, allows for mocking in tests
```
CaitieM20: yay!

common/stats/stats.go
CaitieM20: Doesn't really matter to me either way, we can totally keep it, just wanted to note that the service ingesting it can't tell the difference :)

CaitieM20:
LGTM

dbentley:
LGTM

CaitieM20:
LGTM!

PR #54, opened 2016-07-13
Scheduler Test Cases

CaitieM20:
Adding Test Cases to the Scheduler

- Property based testing to ensure a variety of jobs will be ran and executed successfully
- Unit tests to validate errors are handled appropriately and tasks not executed when failure happens
- refactored the RunTask part of RunJob into its own function


sched/generators.go
```diff
@@ -0,0 +1,51 @@
...
+		Def: jobDef,
+	}
+
+	return job
+}
+
+func GenTask() TaskDefinition {
+	return TaskDefinition{
+		Command: Command{
+			Argv: []string{"cmd1", "cmd2"},
```
dbentley: FWIW, I think these would be more like "cmd", "arg1".

sched/generators.go
```diff
@@ -0,0 +1,51 @@
...
+		},
+	}
+}
+
+// Randomly generates an Id that is valid for
+// use as a sagaId or taskId
+func GenJobId() gopter.Gen {
+	return func(genParams *gopter.GenParameters) *gopter.GenResult {
```
dbentley: It's super-annoying that gopter.gen package wants you to work with monads in a language that's so hostile (syntactically, and type-system-wise) to monads... : /

sched/scheduler/jobRunner.go
```diff
@@ -82,6 +60,32 @@ func runJob(job sched.Job, saga *s.Saga, nodes []cm.Node) {
 	return
 }
 
+// Logic to execute a single task in a Job.  Ensures Messages are logged to SagaLog
+// Logs StartTask, executes Tasks, Logs EndTask
+func runTask(saga *s.Saga, node cm.Node, taskId string, task sched.TaskDefinition) {
+	// Put StartTask Message on SagaLog
+	stErr := saga.StartTask(taskId, nil)
```
dbentley: why have different err's instead of just reusing err?

sched/scheduler/jobRunner.go
```diff
@@ -82,6 +60,32 @@ func runJob(job sched.Job, saga *s.Saga, nodes []cm.Node) {
...
+	}
+
+	// TODO: After a number of attempts we should stop
+	// Trying to run a task, could be a poison pill
+	// Implement deadletter queue
+	taskExecuted := false
+	for !taskExecuted {
+		execErr := node.SendMessage(task)
+		if execErr == nil {
+			taskExecuted = true
```
dbentley: if you just break here you can skip having taskExecuted variable. If you want to keep it, I'd make it a 3-clause for loop (with an empty increment, for now)

sched/scheduler/scheduler_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+			cluster, clusterState := ci.DynamicLocalNodeClusterFactory(int(numNodes))
+			sagaCoord := saga.MakeInMemorySagaCoordinator()
+			scheduler := NewScheduler(cluster, clusterState, sagaCoord)
+
+			job := sched.GenJob(jobId, int(numTasks))
+			err := scheduler.ScheduleJob(job)
+
+			if err != nil {
+				fmt.Println("UnExpected Error Scheduling Job")
```
dbentley: s/UnE/Une/

dbentley:
LGTM

sched/generators.go
```diff
@@ -0,0 +1,51 @@
...
+		const chars = "abcdefghijklmnopqrstuvwxyz0123456789"
+		length := int(genParams.NextUint64()%20) + 1
+		result := make([]byte, length)
+		for i := 0; i < length; i++ {
+			result[i] = chars[genParams.Rng.Intn(len(chars))]
```
jasonbs10: What's the different between nextUint64 and rng.Intn?

sched/generators.go
```diff
@@ -0,0 +1,51 @@
...
+
+	job := Job{
+		Id:  id,
+		Def: jobDef,
+	}
+
+	return job
+}
+
+func GenTask() TaskDefinition {
```
jasonbs10: Is this too deep to generate for property tests?

jasonbs10:
lgtm

sched/generators.go
CaitieM20: cool fixed!  We can make these more realistic or random if that'd be useful for worker testing in the future too!

sched/generators.go
CaitieM20: yea its a Quick Check port which was originally written in Haskell and Erlang so its sort of from that world.  I agree it gets a little ugly sometimes

sched/scheduler/jobRunner.go
CaitieM20: why have the empty increment?  the golang Tour of Go shows this pattern when you want a while loop? https://tour.golang.org/flowcontrol/3

sched/scheduler/scheduler_test.go
CaitieM20: fixed!

sched/generators.go
CaitieM20: I don't think anything, will use rng.Intn much cleaner!

sched/generators.go
CaitieM20: I think as the values applied to tasks gets used more, i.e. worker code gets implemented, then we can definitely make this more random as a generator.  Or we start doing datalocality based scheduling.

CaitieM20: I definitely see this code evolving over time and generating real workloads

PR #55, opened 2016-07-14
Send work from work queue to scheduler

CaitieM20:
- Adding WorkGenerator which reads work of the WorkItem Channel returned by creating a queue
- Updated Scheduler Binary & Scheduler Demo Binary to use this instead of adhoc.  The demo currently runs forever, you must Ctrl + C to quit it.  When we have status it'll end once all work has completed.
- There appears to be a bug in SimpleQueue where Dequeue doesn't unblock the send channel?  I filed a ticket to track.  For now I unblocked myself by adding an initial capacity to SimpleQueue.  If capacity is greater than the number of tasks added everything works fine.  
- Added an interface for Scheduler to decouple the code from WorkGenerator to Scheduler.  WorkGenerator just needs the ScheduleJob method.


binaries/scheduler/main.go
```diff
@@ -17,8 +19,19 @@ func main() {
...
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	// Go Routine which takes data from work queue and schedules it
+	go func() {
+		defer wg.Done()
+		sched.GenerateWork(scheduler, workCh)
+	}()
+
+	wg.Wait()
```
dbentley: This seems like it'll have the same result as running this on the same goroutine. Am I missing something?

binaries/schedulerDemo/main.go
```diff
@@ -25,126 +26,58 @@ func main() {
...
+	go func() {
+		defer wg.Done()
+		scheduler.GenerateWork(localSched, workCh)
+	}()
 
-	fmt.Println("Jobs Completed:", completedSagas)
+	wg.Wait()
 }
 
-// <<<<<<< fd5b67e8f41c7804994527e660a4d6de69d84e09:binaries/schedulerDemo/main.go
```
dbentley: Sorry! My bad. : /

sched/scheduler/workGenerator_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"github.com/golang/mock/gomock"
+	"github.com/scootdev/scoot/saga"
+	"github.com/scootdev/scoot/sched"
+	"github.com/scootdev/scoot/sched/queue"
+	"testing"
+)
+
+func TestGenerateWork_ReadsAllItemsFromChannel(t *testing.T) {
```
dbentley: I think this file would be a lot simpler if you just use a fake queue (the in-memory, simple one) instead of mocking.

sched/scheduler/scheduler.go
```diff
@@ -8,6 +10,10 @@ import (
 	"sync"
 )
 
+type Scheduler interface {
```
dbentley: What if Scheduler just took a chan WorkItem and pulled from it? That might be a clearer interface and also be easier to implement...

sched/scheduler/workGenerator_test.go
jasonbs10: I'm on the fence about this. Dan, would you mind converting one of these to a fake so we can see side-by-side why/if Caitie should eschew the type of testing she's more comfortable/efficient with?

sched/scheduler/scheduler.go
jasonbs10: Either way we want to separate out the management of dead letter queues. So either Scheduler calls that piece or that piece calls Scheduler. It's the latter right now and that seems fine, although I'd probably want an interface/struct instead of a standalone function.

binaries/scheduler/main.go
CaitieM20: In theory yes, But I was trying to mimic real world use of different clients adding work, and the scheduler pulling data off

binaries/schedulerDemo/main.go
CaitieM20: no worries easy fix!

sched/scheduler/workGenerator_test.go
CaitieM20: I'm confused why as well, this function doesn't even require a full queue, just a channel that has work on it.  And its an interface.  I just want to ensure Dequeue is called successfully which the Mock easily tests.  Dequeue doesn't currently work with what's checked into master as well.  

Same with the Scheduler Mock, super easy to test error cases by just telling the mock to return one.

sched/scheduler/scheduler.go
CaitieM20: Yea I viewed this as Jason described.  The scheduler should schedule work, where that work come froms shouldn't matter.  Also The deadletter will have to get handled somewhere.  The WorkGenerator can evolve eventually to do this.

dbentley:
LGTM

PR #57, opened 2016-07-14
sched/queue/memory: make Dequeue do the right thing

dbentley:


sched/queue/memory/simple.go
```diff
@@ -3,40 +3,116 @@ package memory
...
 }
 
-func (i simpleWorkItem) Dequeue() {
+type simpleWorkItem struct {
+	job       sched.Job
+	dequeueCh chan struct{}
+}
+
+func (i *simpleWorkItem) Job() sched.Job {
+	return sched.Job(i.job)
```
jasonbs10: Why cast/construct job, it's already the right type, and you're not doing a deep copy.

jasonbs10: What is the purpose of this cast/copy?

sched/queue/memory/simple.go
```diff
@@ -3,40 +3,116 @@ package memory
...
+	}
+	select {
+	case outCh <- item:
+		s.dequeueCh = item.dequeueCh
+	case req, ok := <-s.inCh:
+		if !ok {
+			s.inCh = nil
+			return
+		}
+		switch req := req.(type) {
```
jasonbs10: If we're going to have multiple types, I'd add a comment about it to justify the switch here and the typeless channel.

sched/queue/memory/simple.go
```diff
@@ -3,40 +3,116 @@ package memory
...
+	inCh      chan queueReq
+	outCh     chan queue.WorkItem
+	dequeueCh chan struct{}
+	items     []*simpleWorkItem
+	nextID    int
+	capacity  int
 }
 
-type simpleWorkItem sched.Job
+func (s *simpleQueueState) loop() {
```
jasonbs10: Do we gain something by not incorporating iter() directly?

sched/queue/memory/simple.go
```diff
@@ -3,40 +3,116 @@ package memory
...
+	job       sched.Job
+	dequeueCh chan struct{}
+}
+
+func (i *simpleWorkItem) Job() sched.Job {
+	return sched.Job(i.job)
+}
+
+func (i *simpleWorkItem) Dequeue() {
+	i.dequeueCh <- struct{}{}
```
jasonbs10: Is this type a placeholder for something? Or is this considered void (I've been using bool in its place).

jasonbs10:
lgtm once the build succeeds.

sched/queue/memory/simple.go
dbentley: Fixed, thanks. (It was a bad merge thing)

sched/queue/memory/simple.go
dbentley: Removed. Thanks.

sched/queue/memory/simple.go
dbentley: Saving a level of indentation, which used to be an issue, but isn't. Fixed.

sched/queue/memory/simple.go
dbentley: It's void. I prefer it to bool because with bool you have to think "wait, does the value matter?"

dbentley: WDYT?

jasonbs10: That's fine. You only have to think about the void once, you have to think about the bool every time you encounter one.

PR #58, opened 2016-07-15
Rewrote latched goroutine test, the previous one was gnarley.

jasonbs10:


common/stats/stats.go
```diff
@@ -149,46 +149,91 @@ func NewCustomStatsReceiver(makeRegistry func() StatsRegistry, latched time.Dura
...
 
 // Called as a goroutine by stats constructor. Loops until ctx is canceled, periodically capturing stats.
-// Note: params are explicit here to make testing easier.
-func latch(stat *defaultStatsReceiver, captured StatsRegistry, latchCh chan chan StatsRegistry,
-	ticker StatsTicker, firstSnapshotAt time.Time, ctx context.Context, testRespCh chan bool,
+func latch(stat *defaultStatsReceiver, captured StatsRegistry, latchCh chan chan CapturedRegistry,
+	ticker StatsTicker, firstSnapshotAt time.Time, ctx context.Context,
 ) {
+	captureTime := Time.Now()
 Loop:
```
dbentley: Do you need this label?

common/stats/stats.go
```diff
@@ -149,46 +149,91 @@ func NewCustomStatsReceiver(makeRegistry func() StatsRegistry, latched time.Dura
...
+		switch m := i.(type) {
+		case metrics.Counter:
+			m.Clear()
+		case metrics.Histogram:
+			m.Clear()
+		}
+	})
+}
+
+type CapturedRegistry struct {
```
dbentley: I'd comment this: why is including time useful?

common/stats/stats_test.go
```diff
@@ -109,52 +107,50 @@ func TestNonLatching(t *testing.T) {
...
+// This function may seem hacky, but it's the simplest way to do ordered testing when we're
+// select'ing on two+ channels: one channel to trigger updates, another accepting
+// requests and immediately responding with a cached update.
+//
+// The Memory Model is light on details regarding select'ing from multiple channels,
+// but testing shows that the following sequence (or something similar) will happen
+// for unbuffered channels:
+// - first send is received/popped but doesn't enter the case statement. The channel is ready again.
+// - second send is is received and blocks further sends as the first send is still pending.
+// - third send blocks until the receive case statement for the first send is fully executed.
```
dbentley: I'd think the second would block until the receive case is fully executed. Are you using a buffered channel?

I'm coming around to we should be using unbuffered channels. A buffered channel patches over a hole until you have more than the buffer can hold, and then, whoops, everything is awful. I prefer to not paper over.

dbentley:
LGTM

Feeling good about this?

common/stats/stats_test.go
jasonbs10: Hah! I a not so subtle hint. I don't remember changing the channel capacity to two. That said, I'm not convinced that in the case where we have a goroutine select'ing on multiple unbuffered channels, that a send on on those channel will be received and executed in the order sent. So I'd do two sends, not three. WDYT?

jasonbs10:
Aside from being unsure about the exact semantics of channel send/receive handoffs, yeah, this seems more reasonable.

Regarding buffered vs unbuffered, buffered is likely a performance boost when the sender/consumer are variably latent, but unbuffered is less forgiving and provides a stronger guarantee that if it works once, it works. I'm happy to do unbuffered until when find it's bogging things down.

common/stats/stats_test.go
dbentley: If a send to an unbuffered channel succeeds, then the receive on the other side succeeded, which means it has entered the case and will not try another receive until the next iteration through the loop... So once should be enough? Or am I missing something?

jasonbs10: Yes, that would be the expected outcome. The relationship between goroutines and channels is more nuanced though and I can't find firm proof that the handoff is synchronous. Is it self evident that it does the right thing? I can always go to the source to mollify myself (I probably should have done that already.)





On Fri, Jul 15, 2016 at 10:18 AM -0700, "Dan Bentley" <notifications@github.com> wrote:












In common/stats/stats_test.go:
>  	rendered = string(stat.Render(true))
>  	if rendered == "{}" {
>  		t.Fatal("Expected non-empty latch with time=0: ", rendered)
>  	}
> -	if stat.Counter().Count() != 0 {
> -		t.Fatal("Expected counter to be cleared after non-latch")
> +}
> +
> +// Three sends are required to ensure the first is handled, not just goroutine-queued...
> +// This function may seem hacky, but it's the simplest way to do ordered testing when we're
> +// select'ing on two+ channels: one channel to trigger updates, another accepting
> +// requests and immediately responding with a cached update.
> +//
> +// The Memory Model is light on details regarding select'ing from multiple channels,
> +// but testing shows that the following sequence (or something similar) will happen
> +// for unbuffered channels:
> +// - first send is received/popped but doesn't enter the case statement. The channel is ready again.
> +// - second send is is received and blocks further sends as the first send is still pending.
> +// - third send blocks until the receive case statement for the first send is fully executed.



If a send to an unbuffered channel succeeds, then the receive on the other side succeeded, which means it has entered the case and will not try another receive until the next iteration through the loop... So once should be enough? Or am I missing something?

PR #60, opened 2016-07-15
Separate Cluster Membership from Communication

dbentley:
Large refactor.

Move cluster membership into cloud/cluster

Create one interface for a Cluster that supports both Static and Dynamic clusters.

Create an implementation that offers both interfaces when hooked up to a channel of NodeUpdates.

Change cluster_test and poolDistributorTest to a more fluent style.

(Note: sched/scheduler and schedulerDemo are currently not building because I haven't yet made the new class that will let you run something, which is the functionality deleted from cluster.node)


cloud/cluster/memory/cluster.go
```diff
@@ -0,0 +1,108 @@
...
+					updates[i].Node = c.wrapper.Wrap(updates[i].Node)
+				}
+			}
+			if !ok {
+				c.inCh = nil
+				continue
+			}
+			c.members.Apply(updates)
+			for _, sub := range c.subs {
+				sub <- updates
```
jasonbs10: I feel like we should create a goroutine for each of these. It might be ok to block if there's only one subscriber allowed, but with multiple it seems iffy.

cloud/cluster/memory/cluster.go
```diff
@@ -0,0 +1,108 @@
...
+	case chan []cluster.Node:
+		// Members()
+		req <- c.members.Current()
+	case chan cluster.Subscription:
+		// Subscribe()
+		ch := make(chan []cluster.NodeUpdate)
+		s := makeSubscription(c.members.Current(), c.reqCh, ch)
+		c.subs = append(c.subs, ch)
+		req <- s
+	case chan []cluster.NodeUpdate:
```
jasonbs10: Shouldn't there be a simpleCluster member function for this case as well? I'd prefer that we pass makeSubscription a cancelFn() rather than closeCh.

cloud/cluster/memory/node.go
```diff
@@ -0,0 +1,48 @@
...
+	return string(n.id)
+}
+
+func NewIdNode(id string) cluster.Node {
+	return &idNode{cluster.NodeId(id)}
+}
+
+var _ cluster.Node = (*idNode)(nil)
+
+func NewIdNodes(num int) []cluster.Node {
```
jasonbs10: We can't call NewIdNodes() a second time to add to our pool. This relates to the question of when to label something as 'fake', and what we consider to be a valid impl.

cloud/cluster/memory/state.go
```diff
@@ -0,0 +1,31 @@
+package memory
+
+import (
+	"github.com/scootdev/scoot/cloud/cluster"
+)
+
+// State holds the state of a Cluster (which is just its members)
+type State struct {
```
jasonbs10: I'm not sure why this deserves its own file. I'd prefer to see it folded into cluster.go

cloud/cluster/memory/subscription.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func makeSubscription(initial []cluster.Node, closeCh chan interface{}, inCh chan []cluster.NodeUpdate) cluster.Subscription {
+	s := &subscriber{
+		inCh:    inCh,
+		outCh:   make(chan []cluster.NodeUpdate),
+		closeCh: closeCh,
+		queue:   nil,
+	}
+	go s.loop()
```
jasonbs10: I don't love how we automatically start goroutines when making a type, and that outside observers have no way of telling if it's still running or not (not including whoever closes inCh).

sched/distributor/poolDistributor.go
```diff
@@ -1,180 +1,96 @@
...
-				dist.mutex.Lock()
-				dist.toRemove[node.Id()] = true
-				dist.mutex.Unlock()
+		case reserveCh <- outgoing:
+			d.free = d.free[1:]
+		case node, ok := <-d.Release:
+			if !ok {
+				return
+			}
+			if _, ok := d.nodes[node.Id()]; ok {
```
jasonbs10: Might as well do the !ok case, possibly panic depending on if we think this should happen or not. Actually, have we thought about how to make our code extra strict in dev/staging, if not all the time?

sched/distributor/poolDistributor_test.go
```diff
@@ -1,230 +1,127 @@
...
-	if node1.Id() != node2.Id() {
-		t.Error("Expected nodes to be the same Id since there is only one in the cluster")
-	}
-
-	select {
-	case <-dist.freeCh:
-		t.Error("Expected cluster to be full schedlued")
-	default:
-	}
+	h := makeHelper(t, "node1")
```
jasonbs10: While this certainly makes the tests much clearer, I think the original mock tests could make use of a helper as well to greatly improve readability. That said, I don't think it would ever be quite as clean as this (which might be ok as long as it's close.)

jasonbs10:
What this refactor has highlighted for me is that our project could use some unified coding guidelines. People have their own valid ways of writing code, but should nonetheless try to adapt to the codebase at hand whenever feasible. It's kind of a free free-for-all right now and that's not ideal. Dan, how onerous/disagreeable were the code standards at Google?

cloud/cluster/memory/cluster.go
dbentley: We do: that's what subscriber is in subscription.go.

cloud/cluster/memory/cluster.go
dbentley: Done.

cloud/cluster/memory/node.go
dbentley: Can you tell me more about that?

Calling NewIdNodes doesn't add it to our pol the first time. If we want to add, we can call NewIdNodes, and then wrap them in NodeUpdates and send them in to the channel.

I agree about "fake" versus "valid", which is why I haven't called any of these fake. They're in-memory implementations.

cloud/cluster/memory/state.go
dbentley: Done.

cloud/cluster/memory/cluster.go
jasonbs10: No, I saw that, which is why this is only "iffy" and clearly not wrong. Like, is there a case where subscriber is shutting down and this send blocks unnecessarily? Maybe not? Having a goroutine on the sender side means it's impossible.

cloud/cluster/memory/subscription.go
dbentley: Why's that?

I think it's intentional. That's why there's no way to kill a goroutine in Go (because killing a thread in Java was never correct). It's like memory, yeah, with garbage collection, there's no way to know it's reclaimed, but that's intentional.

The cost of a goroutine is much lower than the cost of a thread in another language. So I think we should think of it like orphaning memory: bad, worth doing right, important to fix if it starts killing tasks, but not something to be anxious about.

sched/scheduler/jobRunner.go
```diff
@@ -22,23 +22,22 @@ func runJob(job sched.Job, saga *s.Saga, nodes []cm.Node) {
...
-				node := dist.ReserveNode()
-
 				wg.Add(1)
-				go func(node cm.Node, taskId string, task sched.TaskDefinition) {
-					defer dist.ReleaseNode(node)
-					defer wg.Done()
-					runTask(saga, node, taskId, task)
-				}(node, id, task)
+				node := <-workers.Reserve
+				worker := (node.(worker.WorkerNode)).Worker()
```
jasonbs10: Did we make distributor too generic? Is it anything other than WorkerNode / can we do away with this cast somehow?

sched/distributor/poolDistributor.go
dbentley: This should happen. It happens if there's an update to remove a node that's currently reserved and then the node gets released. In that case, we shouldn't add it to the free list. We should process the release and the continue. So the !ok case is handled correctly.

I agree with making code extra strict and failing quickly and loudly.

jasonbs10:
lgtm

cloud/cluster/memory/node.go
jasonbs10: I think a valid impl would allow you to to create a set of nodes that start at id>0. Otherwise, this seems hardcoded and therefore 'fake' (and that's my understanding of fake, you've hardcoded behavior that would otherwise be more dynamic).

sched/distributor/poolDistributor_test.go
dbentley: Yes, mock-based tests can be changed to this style.

I think that mock tests are somewhat harder to read, but that's not the reason they're bad.

They're bad because they're brittle. Take our current structure: Pool is a Library that's used by Scheduler. If we refactor Pool, of course we have to change Pool's tests. And we have to change Scheduler. But if Scheduler's tests use mock Pools instead of fake Pools, then we also have to refactor Scheduler's tests, and these changes are often laborious.

Mock-based tests discourage refactors. Ooof. That's a big one.

Mock-based tests are the right test when you want to test exactly how Scheduler calls Pool, which often happens when you need to meet certain performance goals (e.g., make sure that we're making batch requests). But in most cases, you want to test that Scheduler gets its job done. Mock-based tests also test how Scheduler does its job. That's more work to test the wrong thing.

cloud/cluster/memory/subscription.go
jasonbs10: This is an interface/doc issue primarily. If the expectation is that all users will read the impl then, yeah it's easy to understand, but it's unclear how this works just by looking at the functionality as seen in something like godoc.

sched/scheduler/jobRunner.go
dbentley: Yeah, what we really want is a Pool<Foo>, but go doesn't have generics.

We could make it just deal with WorkerNode, which moves this cast from jobRunner.go to poolDistributor.go (specifically, line 86, when we get the Node in from the NodeUpdate)

cloud/cluster/memory/cluster.go
dbentley: What would that goroutine look like? I tried to make it as simple as possible. Could it be any simpler than subscription.go:loop()? I'd like to understand what you're describing, cause maybe I'm missing how to do this more easily?

cloud/cluster/memory/node.go
dbentley: Ohhhhhh. Yup, you're right.

You can create that set of nodes by using the function above, you just don't get the helper function that creates the whole slice.

cloud/cluster/memory/subscription.go
dbentley: No, you shouldn't have to read the impl!

But as a user, why would you care that this creates a goroutine? You wouldn't care that it allocates memory.

cloud/cluster/memory/cluster.go
jasonbs10: This would be in addition to the subscription.go:loop(). Let me put it this way, it's unlikely that what I'm suggesting is actually necessary, it falls under the category of 'it signals intent' in the case where there are multiple listeners and we're not guaranteed that those listeners are implemented a specific way (we could wrap each sub channel in a type...)

cloud/cluster/memory/subscription.go
jasonbs10: If we pretend this isn't golang for a second, would you raise your eyebrows
if a java constructor created a thread? Maybe it's something I have to get
used to in this language, but I think it's clearly different than how we
expect other languages to be written.

cloud/cluster/memory/cluster.go
dbentley: Yup, I saw your lgtm so I know this isn't blocking.

I'd still like to get on the same page. What would that goroutine look like?

cloud/cluster/memory/subscription.go
dbentley: Yeah, in java, it would be a little weird to not mention.

I think it is different in go, and so I'm pushing myself to try to adapt to this style: goroutines are so cheap that we don't worry about them.

cloud/cluster/memory/cluster.go
jasonbs10: I may have misread how subs is constructed, let's forget about this for the moment.

dbentley:
I agree we could get on more of a same page about how to write code.

I think Google was a lot like Twitter: it's a lot up to teams to decide what they want their code to look like. So, let's do that more.

cloud/cluster/cluster.go
```diff
@@ -0,0 +1,65 @@
+package cluster
```
jasonbs10: What besides 'cluster/' are we putting in the 'cloud/' dir?

CaitieM20: Yea i'm a bit confused why we are refactoring this out of scheduler?  What other code is going to use it?

cloud/cluster/cluster.go
```diff
@@ -0,0 +1,65 @@
...
+	UpdateType NodeUpdateType
+	Id         NodeId
+	Node       Node // Only set for adds
+}
+
+func (u *NodeUpdate) String() string {
+	return fmt.Sprintf("%v %v %v", u.UpdateType, u.Id, u.Node)
+}
+
+type Cluster interface {
```
CaitieM20: Before the refactor the Cluster interface had comments for documentation please don't just delete them. 

The new cluster interface needs comments/docs to reflect context

cloud/cluster/memory/cluster.go
```diff
@@ -0,0 +1,138 @@
+package memory
```
CaitieM20: so most Golang projects I've seen don't do nested interfaces this way.  
Its a bit weird since it forces us to alias packages when we import them.  Do we want to continue following this pattern?

sched/distributor/poolDistributor_test.go
CaitieM20: I disagree with this assessment of Mock based tests.  Since in Golang you can only Mock Interfaces, a mock is just as good as a fake, without having to support as much code.  If you refactor or change the interface of course you have to change the test.  With mocks then you just need to generate the new mock class instead of updating the fake.  

The case you cited can't happen in Golang because PoolDistributor isn't an interface.

I also find handling error cases without mocks much harder, and [testing error handling code has been shown to prevent numerous catastrophic failures in distributed systems](https://www.usenix.org/system/files/conference/osdi14/osdi14-paper-yuan.pdf).  

Mocks are pretty standard way of testing in Industry and at Twitter. I really don't understand the desire to remove all of them from the codebase.  I think we should have a meeting to discuss this because it keeps coming up in code review.

sched/scheduler/jobRunner_test.go
```diff
@@ -40,8 +36,7 @@ func Test_RunTask_PanicWhenWritingStartTaskReturnsFatalError(t *testing.T) {
 	mockCtrl := gomock.NewController(t)
 	defer mockCtrl.Finish()
 
-	// ensures node does not execute task when a fatal error occurs logging starttask
-	nodeMock := cm.NewMockNode(mockCtrl)
+	worker := worker.NewNoopWorker()
```
CaitieM20: I dislike removing the mock here.  Which is a critical part of the test.  If we   No way to ensure the task did not get sent to the worker here.  

If the code gets changed and we introduce a bug that sends to the worker or changes the order of operations our unit tests will no longer catch this. 

This is one of the downsides to using mocks instead of fakes

sched/scheduler/scheduler_test.go
```diff
@@ -6,31 +6,31 @@ import (
...
 
 func Test_ScheduleJob_WritingStartSagaFails(t *testing.T) {
 	mockCtrl := gomock.NewController(t)
 	defer mockCtrl.Finish()
 
-	//nodeMock used to ensure no tasks are executed if StartSaga Fails
-	nodeMock := cm.NewMockNode(mockCtrl)
-	nodeMock.EXPECT().Id().Return("node1")
-	cluster, clusterState := cm.DynamicClusterFactory([]cm.Node{nodeMock})
+	nodes := memory.WrapNodes(memory.NewIdNodes(1), worker.NoopWorkerWrapper)
```
CaitieM20: I would prefer to still use mocks here.  

Once we remove the assumption that everything is running locally, and nodes can fail then we have to write the error handling logic.  Testing this with fakes is difficult.  Testing this with mocks is easy.

sched/worker/noop.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"github.com/scootdev/scoot/cloud/cluster"
+	"github.com/scootdev/scoot/sched"
+)
+
+type noopWorkerController struct{}
+
+func (c *noopWorkerController) RunAndWait(task sched.TaskDefinition) error {
+	return nil
```
CaitieM20: LocalNode used to have a random sleep in here, to make the demo more realistic.  Can you please add this back

CaitieM20:
You didn't update binaries/schedulerDemo/main.go  This refactor will break the demo

cloud/cluster/cluster.go
dbentley: We'll want service discovery and cluster membership in every binary in Cloud Scoot.

I noticed this when I implemented service discovery within Twitter and it felt weird to have it reference details of the scheduler when it was implementing a common interface. So I think the interface should be binary-agnostic.

Does that make sense?

cloud/cluster/cluster.go
dbentley: Sure, I'm happy to add them if we think this is a good direction to go. Can we get a quick agreement on that?

CaitieM20: Yea this looks good!

cloud/cluster/memory/cluster.go
dbentley: I agree that having a whole bunch of packages named "memory" is weird.

I think it's important to separate the heavy-duty implementation into a separate package (especially given that most of our production implementations will be closed-source). I find it's good practice to have the simple, fake implementation live in a separate package (to make sure that implementations don't require unexported features of the interface).

sched/scheduler/jobRunner_test.go
dbentley: How about if the test fails if the worker gets something?

CaitieM20:
So i'm ok with this change being committed to unblock some things if we add the NodeMock back.  That one in particular is being used to explicitly test that a method was or wasn't called as commented.  And moving forward that code will need to handle errors returned by the Node Interface.  This needs to easily be tested.  

Since the Cluster code never returns an error I'm ok with a fake for now.

cloud/cluster/cluster.go
CaitieM20: Sure sounds good

sched/scheduler/jobRunner.go
```diff
@@ -62,7 +61,7 @@ func runJob(job sched.Job, saga *s.Saga, nodes []cm.Node) {
 
 // Logic to execute a single task in a Job.  Ensures Messages are logged to SagaLog
 // Logs StartTask, executes Tasks, Logs EndTask
-func runTask(saga *s.Saga, node cm.Node, taskId string, task sched.TaskDefinition) {
+func runTask(saga *s.Saga, worker worker.WorkerController, taskId string, task sched.TaskDefinition) {
```
CaitieM20: So I dislike this coupling.  Having the Interface which was just send Message allows us to implement any Node.  Not just Workers

cloud/cluster/cluster.go
```diff
@@ -0,0 +1,65 @@
...
+type Subscription struct {
+	InitialMembers []Node
+	Updates        chan []NodeUpdate
+	Closer         io.Closer
+}
+
+// NodeWrapper lets users wrap a Node.
+// E.g., if service discovery hands you an address, you can wrap it with a Node that also contains
+// the Thrift client to talk to that address
+type NodeWrapper interface {
```
CaitieM20: I don't totally understand what the NodeWrapper buys us over the Interface?  Why not just have factory methods that return an implementation of this interface

The idea behind the interface is that it allows any implementation to be called by the Scheduler.  The current code now couples the scheduler to NodeWorker etc...

The interface also defined types of errors that the node could return.  The wrapper defines no common errors?  The scheduler will eventually need some insight on the types of possible errors that could be returned when implementing dead letter queues etc...

sched/scheduler/jobRunner_test.go
CaitieM20: With the way it was written a mock will cause a failure?  How will the fake do this?

dbentley: Hmm, thinking about this more: why? Why is it important that the task did not get sent to the worker?

I agree that testing is super important, but also that testing the wrong things leads us down a path that makes it harder to do all the testing we could.

sched/worker/noop.go
dbentley: Sure, how's this: now it's in a separate class, so we can use it, or not.

(And I added it to the worker demo. I had to up it from micro second to millisecond to make it noticeable)

sched/scheduler/scheduler_test.go
dbentley: I agree that testing errors is important. That's why I want to use fakes: I think testing it with fakes is easier than with mocks.

Obviously, we disagree.

We clearly both think that testing is important. Would you be willing to trust me that I'm not trying to get rid of testing, but make it better and give me the chance to show you that as our codebase grows?

sched/scheduler/jobRunner.go
dbentley: SendMessage was very specific, because it took a sched.TaskDefinition.

That meant that cluster.Node conflated:
*) discovery
*) communication
*) type

So to implement the Node, you had to implement Worker, so my goal was to decouple that and have Node represent any Node, and then Worker adds to Node by letting you send the relevant Worker messages.

dbentley:
OK, updated with:
a) do the factory explicitly in jobRunner.go
b) move the fake workers into their own package

sched/worker/fake/fake.go
```diff
@@ -13,35 +14,22 @@ func (c *noopWorkerController) RunAndWait(task sched.TaskDefinition) error {
...
-	return &workerNode{
-		worker: &noopWorkerController{},
-		node:   node,
-	}
-}
-
-var NoopWorkerWrapper = &noopWorkerWrapper{}
-
-type waitingWorkerWrapper struct {
+func NoopWorkerFactory(node cluster.Node) worker.WorkerController {
```
jasonbs10: I know we refer to these as factories in conversation, but we also had an earlier discussion where we replaced '*Factory' with 'New*' or 'Make*'. Probably don't want to have all three.

jasonbs10:
lgtm, this definitely reads better without the wrapper.

sched/worker/fake/fake.go
dbentley: OK, I updated the function names, but I think we still need the type name to be factory (in worker.go). WDYT? Does that sound right?

dbentley:
Oh, I also updated worker/fake.go to add a panic worker, so that jobRunner_test.go uses a fake and ensures that tasks aren't executed.

sched/worker/fake/fake.go
jasonbs10: Yeah, having Factory there looks fine.

cloud/cluster/memory/cluster.go
```diff
@@ -0,0 +1,125 @@
...
+	case chan []cluster.Node:
+		// Members()
+		req <- c.members.Current()
+	case chan cluster.Subscription:
+		// Subscribe()
+		ch := make(chan []cluster.NodeUpdate)
+		s := makeSubscription(c.members.Current(), c, ch)
+		c.subs = append(c.subs, ch)
+		req <- s
+	case chan []cluster.NodeUpdate:
```
CaitieM20: The old dynamic cluster code used to check for duplicates and not re-add them.  I think we should add this back here.  Making Add idempotent will avoid a lot of bugs, and easier to reason about.

cloud/cluster/memory/cluster.go
```diff
@@ -0,0 +1,125 @@
...
+		// subscription.Close()
```
CaitieM20: nit: should this comment be deleted?

cloud/cluster/memory/cluster.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+}
+
+func (c *simpleCluster) closeSubscription(s *subscriber) {
+	c.reqCh <- s.inCh
+}
+
+// State holds the state of a Cluster (which is just its members)
+type State struct {
+	members []cluster.Node
```
CaitieM20: why not just use a map like dynamic cluster did.  Will make lookup and removal faster

cloud/cluster/memory/cluster.go
CaitieM20: Is this really an in memory fake like inMemoryQueue or inMemorySagaLog?

shouldn't it just be part of cluster like dynamicCluster was, we don't plan on replacing this with a durable version.  The idea was that this class is reused and however you do service discovery updates just sends updates to this class/object

cloud/cluster/cluster.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"fmt"
+	"io"
+)
+
+type NodeId string
+
+type Node interface {
+	Id() NodeId
```
CaitieM20: Do we want to expand this to something more definitive than an Id?  Like host, port, etc... so the functions that take in a Node and generate a worker don't have to parse some special format?

At the very least we need a comment specifying how to get the connection info from the Id()

cloud/cluster/cluster.go
```diff
@@ -0,0 +1,64 @@
...
+}
```
CaitieM20: nit: can Node Interface and related structs/types be in its own file.  I keep forgetting its in cluster and spend a bunch of time looking for it.

cloud/cluster/cluster.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (u *NodeUpdate) String() string {
+	return fmt.Sprintf("%v %v %v", u.UpdateType, u.Id, u.Node)
+}
+
+// Cluster represents a cluster of Nodes.
+type Cluster interface {
+	// Members returns the current members, or an error if they can't be determined.
+	Members() ([]Node, error)
```
CaitieM20: Does it need to return an error? Either we know of some nodes or it doesnt and returns none. 

I agree that the code reading from service discovery needs to handle errors, but I don't think that should leak all the way to the consumers and force them to handle those errors.

cloud/cluster/cluster.go
```diff
@@ -0,0 +1,64 @@
...
+	// Subscribe subscribes to changes to the cluster.
+	Subscribe() (Subscription, error)
```
CaitieM20: Why would an error be returned here?  Do we need to have the consumers of this class write additional error handling logic?

sched/worker/worker.go
```diff
@@ -0,0 +1,47 @@
+package worker
+
+import (
+	"github.com/scootdev/scoot/cloud/cluster"
+	"github.com/scootdev/scoot/sched"
+)
+
+type WorkerFactory func(node cluster.Node) WorkerController
```
CaitieM20: Comment here will be great for readability, and to describe what the intended behavior of functions that implement this type do.

sched/scheduler/jobRunner_test.go
```diff
@@ -40,8 +36,7 @@ func Test_RunTask_PanicWhenWritingStartTaskReturnsFatalError(t *testing.T) {
 	mockCtrl := gomock.NewController(t)
 	defer mockCtrl.Finish()
 
-	// ensures node does not execute task when a fatal error occurs logging starttask
-	nodeMock := cm.NewMockNode(mockCtrl)
+	worker := fake.NewPanicWorker()
```
CaitieM20: So this works but will get very cumbersome to create a new fake every time you want a slightly different behavior, resulting in  more code to maintain / update.  I prefer mocks, since you get all the behavior at once.  We can discuss on Thursday, this is fine for now

cloud/cluster/memory/cluster.go
dbentley: Done.

(I think I had been scared of the sort, but I think this is fine)

cloud/cluster/cluster.go
dbentley: Yeah, I buy that. Done.

cloud/cluster/cluster.go
dbentley: Done.

sched/worker/worker.go
dbentley: OK, I put something in; WDYT? What would be clearer?

cloud/cluster/cluster.go
dbentley: Done (but kept NodeUpdate with cluster.go; wdyt?)

cloud/cluster/cluster.go
dbentley: I think that's probably right, but we should wait until we're using it for real and can design it as we use it.

We might also have this be one of those cases where you expand on the interface, like, you might have some cluster implementations end up having two ports to talk to different services (e.g., http vs. thrift), so I don't think we can proscribe it too much just yet...

cloud/cluster/memory/cluster.go
dbentley: Done.

cloud/cluster/memory/cluster.go
dbentley: No, I was trying to say "this line is the result of someone else calling subscription.Close()", but was unclear, so I tried to clarify. How's that?

cloud/cluster/memory/cluster.go
dbentley: The polling cluster ends up being almost the inverse of this, where it polls and then creates updates, but because it's polling, it has its own state of the current members, as well. So I think it may actually make sense in production to skip with this code altogether, to not have a second pool of current nodes. So, yes, I think this implementation is not necessary in production, and thus belongs in a separate package.

CaitieM20: I agree you could do it this way, but then everyone who implements Cluster logic has to reimplement all this core subscription logic which seems wasteful.  If we truly want this to be consumed as an OSS project, we should provide as much core logic as possible and then just have people plug in their infrastructure specific things like service discovery.  That's why it was originally designed that way where DynamicCluster contained core logic on how to do this and the you could just notify it when nodes were added or removed.

cloud/cluster/cluster.go
CaitieM20: Looks good thanks!

cloud/cluster/cluster.go
CaitieM20: Sounds good to me

cloud/cluster/memory/cluster.go
CaitieM20: Great thanks!

CaitieM20:
LGTM we can keep iterating on this moving forward

PR #61, opened 2016-07-22
Adding Job Status

CaitieM20:
Adding the Ability to Query Job Status.  Right now this is just code in the scheduler.  Will build the server API on top once this all looks good.

* Added Code to convert SagaState to JobStatus
* Added Non Recovery Status Query To SagaCoordinator.  Removed Rollback RecoveryType from recoverState method not being used.  
* Updated Demo to Check that all Scheduled Jobs have completed before exiting

### DIFF
--- a/binaries/schedulerDemo/main.go
+++ b/binaries/schedulerDemo/main.go
@@ -44,7 +44,7 @@ func main() {
 	go func() {
 		defer wg.Done()
 		ids := generateTasks(workQueue, 1000)
-		checkState(ids, sagaCoord)
+		waitUntilJobsCompleted(ids, sagaCoord)
 		fmt.Println("all tasks completed")
 	}()
 
@@ -59,7 +59,7 @@ func main() {
 
 // Checks the State of all Passed in jobs.  When they are all
 // completed returns
-func checkState(ids []string, sc s.SagaCoordinator) {
+func waitUntilJobsCompleted(ids []string, sc s.SagaCoordinator) {
 
 	completedJobs := 0
 
--- a/binaries/schedulerDemo/main.go
+++ b/binaries/schedulerDemo/main.go
@@ -36,13 +36,16 @@ func main() {
 	//}()
 
 	var wg sync.WaitGroup
-	wg.Add(2)
+	wg.Add(1)
 
 	workQueue := memory.NewSimpleQueue(1000)
 
+	// generates tasks, then checks if they are all done before returning
 	go func() {
 		defer wg.Done()
-		generateTasks(workQueue, 1000)
+		ids := generateTasks(workQueue, 1000)
+		checkState(ids, sagaCoord)
+		fmt.Println("all tasks completed")
 	}()
 
 	//This go routine will never exit will run forever
@@ -52,6 +55,25 @@ func main() {
 	}()
 
 	wg.Wait()
+}
+
+// Checks the State of all Passed in jobs.  When they are all
+// completed returns
+func checkState(ids []string, sc s.SagaCoordinator) {
+
+	completedJobs := 0
+
+	for completedJobs < len(ids) {
+		time.Sleep(100 * time.Millisecond)
+		completedJobs = 0
+		for _, id := range ids {
+
+			status, _ := scheduler.GetJobStatus(id, sc)
+			if status.Status == sched.Completed {
+				completedJobs++
+			}
+		}
+	}
 }
 
 /*
--- a/saga/sagaRecovery_test.go
+++ b/saga/sagaRecovery_test.go
@@ -18,7 +18,7 @@ func TestRecoverState_GetMessagesReturnsError(t *testing.T) {
 	sagaLogMock.EXPECT().GetMessages(sagaId).Return(nil, errors.New("test error"))
 	sc := MakeSagaCoordinator(sagaLogMock)
 
-	state, err := recoverState(sagaId, sc, RollbackRecovery)
+	state, err := recoverState(sagaId, sc)
 
 	if err == nil {
 		t.Error("Expected GetMessages return error to cause recoverState to return an error")
@@ -39,7 +39,7 @@ func TestRecoverState_GetMessagesReturnsEmptyList(t *testing.T) {
 	sagaLogMock.EXPECT().GetMessages(sagaId).Return(nil, nil)
 	sc := MakeSagaCoordinator(sagaLogMock)
 
-	state, err := recoverState(sagaId, sc, RollbackRecovery)
+	state, err := recoverState(sagaId, sc)
 
 	if state != nil {
 		t.Error("Expected returned Saga to be nil when no messages in SagaLog")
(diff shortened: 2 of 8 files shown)


saga/sagaRecovery_test.go
```diff
@@ -92,7 +92,7 @@ func TestRecoverState_UpdateSagaStateFails(t *testing.T) {
 	sagaLogMock.EXPECT().GetMessages(sagaId).Return(msgs, nil)
 	sc := MakeSagaCoordinator(sagaLogMock)
 
-	state, err := recoverState(sagaId, sc, RollbackRecovery)
```
dbentley: Were these tests trying to test rollback recovery? Are they still relevant?

sched/scheduler/scheduler_test.go
```diff
@@ -1,6 +1,6 @@
 package scheduler
 
-import (
+/*import (
```
dbentley: I'm guessing you commented this out so you could run other tests (which is a silliness of go's tooling that you had to) and that you're in the process or reenabling it.

sched/definitions.go
```diff
@@ -24,3 +24,26 @@ type Command struct {
...
+	// Currently Scheduled and In Progress Job/Task
+	InProgress
+
+	// Successfully Completed Job/Task
+	Completed
+
+	// Job was Aborted, Compensating Tasks are being Applied.
+	// A RollingBack task has not finished its compensating
+	// tasks yet.
+	RollingBack
```
dbentley: If I have Job J with Task T.

If I abort J, what is the status of T?
If I then startCompensatingTask of T, what is the status of T?

It seems weird that they'd be the same at those two different points of time.

binaries/schedulerDemo/main.go
```diff
@@ -54,6 +57,25 @@ func main() {
 	wg.Wait()
 }
 
+// Checks the State of all Passed in jobs.  When they are all
+// completed returns
+func checkState(ids []string, sc s.SagaCoordinator) {
```
dbentley: Maybe call it some variation on wait?

sched/scheduler/jobStatus.go
```diff
@@ -0,0 +1,83 @@
+package scheduler
+
+import (
+	s "github.com/scootdev/scoot/saga"
+	"github.com/scootdev/scoot/sched"
+)
+
+type JobStatus struct {
+	Id         string
+	Status     sched.Status
```
dbentley: It always seems weird to me to have a struct with Status in the name with an enum that's named Status. Does this bug you? How can we

dbentley:
LGTM

sched/definitions.go
jasonbs10: As soon as we abort we either know we're 'Completed' (although I don't see the requisite errCode or anything of that nature in JobStatus) or 'RollingBack'. Is that not sufficient, keeping in mind that we already have code that marks something 'Completed' in both the failure/success cases?

sched/scheduler/jobStatus.go
jasonbs10: I think we already do this elsewhere. I can't think of anything better unless you'd prefer JobStatus and JobStatusMsg.

sched/definitions.go
dbentley: It's asymmetric wrt the forward-case: there we have ready, running, done. But here we seem to have only one state for ready/running.

Not a huge deal, especially since rollback isn't our roadmap (but then I wonder why we want to handle it now, instead of just panicking because ahhh we don't put sagas into this state at all).

sched/scheduler/jobStatus.go
dbentley: Yeah, it's bugged me before, so I thought I'd mention it so we could figure out if there's a way to reduce confusion in both this specific and the general case, cause, darnit...

sched/scheduler/jobStatus_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"fmt"
+	"github.com/golang/mock/gomock"
+	"github.com/leanovate/gopter"
+	"github.com/leanovate/gopter/prop"
+	s "github.com/scootdev/scoot/saga"
+	"github.com/scootdev/scoot/sched"
+	"testing"
+)
+
+func Test_GetJobStatus_Error(t *testing.T) {
```
jasonbs10: We should really have that meeting about testing methodology. I don't care if it's fake or mock but I would want to refactor these to look more like Dan's latest PR (it's what I'm more comfortable with given shared ownership of [test] code).

jasonbs10:
lgtm

sched/scheduler/scheduler_test.go
CaitieM20: yes whoops uncommented locally and didn't publish

saga/sagaRecovery_test.go
CaitieM20: No they were not testing RollbackRecovery, the logic for recovering state and applying the safe state logic used to be in one function, when it got refactored I forgot to remove the RecoveryState

Its now split sagaRecovery recoverState is all the logic to take SagaLog messages and reapply them to create a *SagaState.  SagaCoordinator.RecoverSagaState now handles applying the Recovery Method.

sched/definitions.go
CaitieM20: So first off a user cannot Abort a Saga.  Just like a user cannot Abort a Transaction.  User cancelled Jobs is a feature we can add down the line and would require additional status codes to be added.

To answer your question if a Saga is Aborted, then the Job is RollingBack and all the Tasks will be in status RollingBack, Even if their compensating action hasn't started.  Once the compensating task completes then the task is RolledBack state.  When all compensating tasks complete the Job is in the RolledBack state.  I felt this was the clearest way to convey that we are no longer making forward progress on the Job.   Open to suggestions if that's confusing.  

RollingBack indicates that the Saga was aborted for some reason and we are in the process of cleaning up the job.  Since none of these actions are Atomic you do have intermediary states where some tasks are RolledBack and some are still RollingBack.

The property based tests are a great indication of the whole state transition from Saga to Job/Task Status.  

@dbentley I started with just the ForwardRecovery case but then decided it was easy to handle all states.  Also helped me think through it.  Jobs with Rollback Recovery have Ready(Not Started) Running (In Progress), Aborted (RollingBack), Done(Completed/RolledBack) so the user can differentiate between the successful and unsuccessful branch

binaries/schedulerDemo/main.go
CaitieM20: done

sched/scheduler/jobStatus.go
CaitieM20: If you have a suggestion that differentiates this happy to update, but best I could come up with other than JobStatusObject

sched/scheduler/jobStatus_test.go
CaitieM20: Agree added a meeting to the calendar on Friday for us to discuss this and come with a path forward.

PR #63, opened 2016-07-25
Exposing JobStatus via apiserver

CaitieM20:
Exposing JobStatus via apiserver

* moved JobStatus struct & enum into thrift file defintions & added a GetStatus API
* moved jobStatus conversion code to scootapi/server.  This is just translation code.  Also it decouples this code from scheduler should we want to make the API Server its own service one day as previously discussed
* Added GetStatus API to scootapi binary, so it can be queried from command line tool
* ran go generate ./... which updated the MockQueue code


Makefile
```diff
@@ -24,6 +24,9 @@ dependencies:
 	# We use 'go get' here because 'go install' will not build out of our vendored mock repo.
 	go get github.com/golang/mock/mockgen
 
+	# Necessary to generate thrift files, in addition to manually installing thrift via homebrew
+	go get github.com/apache/thrift/lib/go/thrift
```
dbentley: Hrm, will this cause a problem because it's different than the vendored version?

jasonbs10: Shouldn't. We can have copies in both GOPATH and vendor. GOPATH is used with the thrift compiler, and vendor is used for the thrift lib in the code, I think it's fine.

sched/scheduler/jobStatus.go
```diff
@@ -1,83 +0,0 @@
-package scheduler
```
dbentley: I'd suggest keeping this file. That sounds crazy and like duplication, so let me explain:

We'll want to test the status of jobs in scheduler code. It'll be easier to do that if it's a POGO struct instead of being a Thrift object with all of its weirdness.

Then scheduler code does saga -> jobStatus, and server does jobStatus -> thrift.

This keeps thrift at the edges of the system.

(Though it's not important for this review, it's a shame to delete this nice code you've already written!)

Makefile
dbentley: But if they change thrift overnight, it means it'll use one version to generate the files and then another to compile against, which could lead to an error?

scootapi/server/serve_test.go
```diff
@@ -106,9 +122,10 @@ func TestQueueError(t *testing.T) {
 
 func TestQueueFillsAndEmpties(t *testing.T) {
 	q := memory.NewSimpleQueue(1)
-
 	defer q.Close()
-	handler := server.NewHandler(q)
+	sc, _ := CreateSagaCoordMock(t)
```
dbentley: I bet you can just pass nil?

dbentley:
LGTM

Makefile
jasonbs10: I'd hope they don't make that type of change lightly, but you're right, it's an uncontrolled variable. Actually, I'm not even sure why we'd need the thrift libraries to generate code.

CaitieM20: I'll remove it for now,  We may not need it.  I thought I did at one point trying to get it to work, but I think it may have just been my thrift version needed to be 0.9.3

sched/scheduler/jobStatus.go
CaitieM20: So I deleted it for a few reasons:
* Status is just an API all the Status is basically an interpretation of the SagaLog.  All the app logic is Saga Based this is just wire struct info.  The code would have just had an additional layer of indirection which translates it from this object to a thrift object, which I thought was kind of wasteful and didn't gain us anything
* Having it in scheduler could potentially couple status to scheduler and we may want to separate them one day for load reasons.
* The scheduler tests actually just check SagaStatus so I think we don't need this?  The thrift stays just in the server.  If it starts creeping into scheduler we can add it back.  The beauty of git having the entire history.

scootapi/server/serve_test.go
CaitieM20: doesn't work because SagaCoordinator is a struct not a pointer

jasonbs10:
lgtm

PR #64, opened 2016-07-26
daemon: improve server start-up to reclaim old socket

dbentley:
I accidentally submitted my change to the daemon without the comment about making the directory, so I decided to fix the code to do the todo. And while I was there, it was easy to make it handle this case, as well.

### DIFF
--- a/daemon/client/cli/cli.go
+++ b/daemon/client/cli/cli.go
@@ -14,7 +14,7 @@ func (c *CliClient) Exec() error {
 	return c.rootCmd.Execute()
 }
 
-func (c *CliClient) Close(*cobra.Command, []string) error {
+func (c *CliClient) Close() error {
 	return c.comms.Close()
 }
 
@@ -25,7 +25,7 @@ func NewCliClient(dialer conn.Dialer) (*CliClient, error) {
 		Use:                "scootcl",
 		Short:              "Scootcl is a command-line client to Scoot Daemon",
 		Run:                func(*cobra.Command, []string) {},
-		PersistentPostRunE: r.Close,
+		PersistentPostRunE: func(*cobra.Command, []string) error { return r.Close() },
 	}
 
 	r.rootCmd = rootCmd
--- a/daemon/integration/integration_test.go
+++ b/daemon/integration/integration_test.go
@@ -1,0 +1,67 @@
+package integration_test
+
+import (
+	"fmt"
+	"github.com/scootdev/scoot/daemon/client/cli"
+	"github.com/scootdev/scoot/daemon/client/conn"
+	"github.com/scootdev/scoot/daemon/integration"
+	"github.com/scootdev/scoot/daemon/server"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
+	"testing"
+)
+
+func TestIntegration(t *testing.T) {
+	tempDir, err := ioutil.TempDir("", "scoot-listen-")
+	if err != nil {
+		t.Fatal("could not make temp directory", err)
+	}
+	// defer os.RemoveAll(tempDir)
+
+	scootDir := path.Join(tempDir, "scoot")
+	err = os.Setenv("SCOOTDIR", scootDir)
+
+	s, err := server.NewServer(nil)
+	if err != nil {
+		t.Fatal("could not make server", err)
+	}
+
+	l, err := server.Listen()
+
+	go func() {
+		err := s.Serve(l)
+	}()
+
+	if err != nil {
+		t.Fatal("could not set env", err)
+	}
+
+	dialer, err := conn.UnixDialer()
+	if err != nil {
+		t.Fatal("could find Scoot Daemon", err)
+	}
+
+	err = testEcho(dialer)
+}
+
+func testEcho(dialer conn.Dialer) error {
+	stdout, _, err := run(dialer, "echo", "foo")
+	if err != nil {
+		return fmt.Errorf("error echo'ing: %v", err)
+	}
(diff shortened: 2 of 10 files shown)


daemon/server/listen.go
```diff
@@ -0,0 +1,87 @@
...
+	return nil, err
+}
+
+// replaceDeadServer handles the common case where a daemon has died but the socket file still exists.
+// If the address is already in use, we try connecting to it.
+// If we get connection refused, we infer the server is dead and
+// remove the socket file, and then try serving.
+// Returns a valid listener or nil.
+func replaceDeadServer(socketPath string, err error) net.Listener {
+	if !isAddrAlreadyInUse(err) {
```
jasonbs10: I'd clarify why InUse is the only valid case for proceeding with replacement (because we consider every other error to be unrecoverable and to require manual intervention).

daemon/server/listen.go
```diff
@@ -0,0 +1,87 @@
...
+		return nil
+	}
+
+	conn, connErr := net.Dial("unix", socketPath)
+	if connErr == nil {
+		// There is an active server, so bow out gracefully
+		conn.Close()
+		return nil
+	}
+	if !isConnectionRefused(connErr) {
```
jasonbs10: Again, clarify why ConnRefused is special (because the server only accepts one connection and that's already in use?).

daemon/server/server.go
```diff
@@ -10,32 +10,41 @@ import (
 )
 
 // Create a protocol.ScootDaemonServer
-func NewServer(runner runner.Runner) (protocol.ScootDaemonServer, error) {
-	return &Server{runner}, nil
+func NewServer(runner runner.Runner) (*Server, error) {
+	s := &Server{
+		runner: runner,
+	}
+	s.server = grpc.NewServer()
```
jasonbs10: nit: move this into the instantiation.

daemon/client/cli/cli.go
```diff
@@ -14,7 +14,7 @@ func (c *CliClient) Exec() error {
 	return c.rootCmd.Execute()
 }
 
-func (c *CliClient) Close(*cobra.Command, []string) error {
+func (c *CliClient) Close() error {
```
jasonbs10: Oh, what is the significance of removing these args?

daemon/integration/utils.go
```diff
@@ -0,0 +1,76 @@
...
+
+func captureOutput() *output {
+	oldStdout, oldStderr := os.Stdout, os.Stderr
+	stdout, stdoutCh := makeDiversion()
+	stderr, stderrCh := makeDiversion()
+	os.Stdout, os.Stderr = stdout, stderr
+	return &output{stdout, stdoutCh, stderr, stderrCh, oldStdout, oldStderr}
+}
+
+func makeDiversion() (*os.File, chan string) {
```
jasonbs10: Comment here would be nice since the meaning of diversion isn't immediately clear.

daemon/client/cli/cli.go
dbentley: You need to accept those args to match the type for PersistentPostRunE, but then it means you can't just Close it. Now that I needed to Close it somewhere else (integration_test.go), I changed this to the bare form, and then pass an anonymous func to PersistentPostRunE below, that just discards these args.

daemon/server/server.go
dbentley: Done.

daemon/server/listen.go
dbentley: Done (in the comment above replaceDeadServer)

daemon/server/listen.go
dbentley: Done.

daemon/integration/integration_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"path"
+	"testing"
+)
+
+func TestIntegration(t *testing.T) {
+	tempDir, err := ioutil.TempDir("", "scoot-listen-")
+	if err != nil {
+		t.Fatal("could not make temp directory", err)
+	}
+	// defer os.RemoveAll(tempDir)
```
CaitieM20: nit: delete commented out code?  or is this necessary

daemon/integration/utils.go
dbentley: Added more comments to this file; WDYT?

(Note: this is not new code; this is moved over from cli_test.go to be shared by both client and integration tests)

daemon/integration/integration_test.go
dbentley: It is not necessary (the tmpreaper will eventually get it), but good to do. I accidentally pushed code as I was testing it; whoops!

Fixed.

daemon/integration/integration_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	dialer, err := conn.UnixDialer()
+	if err != nil {
+		t.Fatal("could find Scoot Daemon", err)
+	}
+
+	err = testEcho(dialer)
+}
+
+func testEcho(dialer conn.Dialer) error {
```
CaitieM20: nit: maybe a different name then testEcho. suggestion: validateEcho  Kind of confusing to have a method named test that isn't actually a test case in a test file.

CaitieM20:
LGTM

jasonbs10:
lgtm

daemon/integration/integration_test.go
dbentley: OK, refactored to use m.TestMain instead for setup, and now TestEcho is a real test.

PR #65, opened 2016-07-26
Combinte Scheduler and ApiServer into one binary

CaitieM20:
Combined Scheduler & API Server into one binary.  
Tested by running scootapi command line tool against combined server locally. 
We still need to rename this ... open to suggestions


binaries/scheduler/main.go
```diff
@@ -24,6 +30,14 @@ func main() {
...
+	handler := server.NewHandler(workQueue, sagaCoordinator)
+
+	// Start API Server
+	err := server.Serve(handler, addr, transportFactory, protocolFactory)
+	if err != nil {
+		log.Fatal("Error serving Scoot API: ", err)
+	}
+
 	var wg sync.WaitGroup
 	wg.Add(1)
```
dbentley: Reading this, it seems like it won't actually get to the work generator, because the Serve will wait. Am I missing something?

CaitieM20: You are right.  I didn't realize that it blocked forever.  Fixed!

dbentley:
LGTM

CaitieM20:
LGTM checker is being slow.  Merging anyway to unblock work

PR #66, opened 2016-07-27
adding test suite

CaitieM20:
Adding a Smoke Test / Alpha Test Suite to the Command Line tool.  This is an end to end test from the users perspective.  It generates random jobs with a random number of tasks.  These are sent to the scheduler via the RunJob API.  The scheduler will run the job by sending it to the specified worker (currently just a local node).  The client checks that status of each job/task to ensure its done before moving on.  

To Run:
Start Scoot API & Scheduler Service
go run ./binaries/scheduler/main.go

Start The Smoke Test

Also removing SchedulerDemo.  This code is now redundant since we have a full end to end path working.


scootapi/testHelpers/generators.go
```diff
@@ -0,0 +1,51 @@
...
+
+// Generates a scoot.JobDefinition
+func GenJobDefinition(rng *rand.Rand) *scoot.JobDefinition {
+
+	def := scoot.NewJobDefinition()
+	def.Tasks = make(map[string]*scoot.TaskDefinition)
+
+	numTasks := rng.Intn(100) + 1
+	for i := 0; i < numTasks; i++ {
+		taskId := GenTaskId(rng)
```
dbentley: So, it'll be an error if these end up colliding. So maybe just have 

taskId := fmt.Sprintf("task%d", i) ?

dbentley:
LGTM

scootapi/testHelpers/generators.go
```diff
@@ -0,0 +1,51 @@
+package testHelpers
```
dbentley: can you rename the package to be all lowercase?

https://blog.golang.org/package-names
https://golang.org/doc/effective_go.html#names

CaitieM20: yup done

scootapi/testHelpers/generators.go
CaitieM20: It won't throw an error, since its inserting into a map i'll just overwrite the old one.  Will update to ensure no collisions though

dbentley: Ugh, you're right, of course.

PR #67, opened 2016-07-27
Integrated Runner into WorkerServer.

jasonbs10:


jasonbs10:
TODO: additional testing + unit tests

workerapi/client/commands.go
```diff
@@ -113,3 +113,5 @@ func (c *Client) queryworker(cmd *cobra.Command, args []string) error {
 	log.Println(render.Render(status))
 	return err
 }
+
+//TODO: implement clear()
```
dbentley: What is clear?

workerapi/server/server.go
```diff
@@ -3,16 +3,20 @@ package server
...
 	"log"
-	"math/rand"
 	"strconv"
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/luci/go-render/render"
 	"github.com/scootdev/scoot/common/stats"
+	"github.com/scootdev/scoot/runner"
+	execer "github.com/scootdev/scoot/runner/execer/os"
```
dbentley: osexecer?

jasonbs10:
I think this is an improvement (although it touched more files than I wanted). Dan, is this ok for you?

daemon/client/conn/conn.go
```diff
@@ -101,25 +102,37 @@ func (c *conn) Echo(arg string) (string, error) {
...
 	r, err := c.client.Status(context.Background(), &protocol.StatusQuery{RunId: string(run)})
 	if err != nil {
-		return runner.ProcessStatus{}, err
+		return runner.ProcessStatus{State: runner.FAILED, Error: err.Error()}
 	}
-	return protocol.ToRunnerStatus(r), nil
+	return protocol.ToRunnerStatus(r)
+}
+
+func (c *conn) StatusAll() []runner.ProcessStatus {
```
dbentley: BTW, let me just highlight this as a place where I have no idea what the right thing to do is.

Conn wants to offer "hey, you can run this". Options for how to do that:
a) just offer grpc interface
b) create a new domain model
c) be a runner.Runner

All of these seem bad, but for different reasons. So, yeah, clumsy, and I'm sad. What do you think?

(Not a comment on this review, really, but just a conversation to keep in the back of our minds)

runner/execer/os/os_execer.go
```diff
@@ -14,9 +16,13 @@ func NewExecer() execer.Execer {
 type osExecer struct{}
 
 func (e *osExecer) Exec(command execer.Command) (result execer.Process, err error) {
+	if len(command.Argv) <= 0 {
```
dbentley: I'd just say == 0; if len starts returning negative numbers, everything is awful.

(But that's so far below a nit...)

runner/fake/fake.go
```diff
@@ -11,12 +11,23 @@ func NewRunner() runner.Runner {
 type fakeRunner struct {
```
dbentley: Huh, maybe we don't need fake runner at all now that we have the simple runner and the sim execer?

runner/local/simple.go
```diff
@@ -2,69 +2,136 @@ package local
...
-	return result, nil
+	result.Abort()
+	delete(r.running, run)
+	r.runs[run] = runner.AbortStatus(run)
+	return r.runs[run]
 }
 
-func (r *simpleRunner) markFinished(p execer.Process, runId runner.RunId, status execer.ProcessStatus) {
+func (r *simpleRunner) Erase(run runner.RunId) {
+	// Best effort is fine here.
```
dbentley: need to lock

runner/local/simple.go
```diff
@@ -2,69 +2,136 @@ package local
...
+	}
+
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	r.running = nil
+
+	// Check if we already marked this run as done (i.e. in Abort())
+	// Further, check if we aborted due to a timeout.
+	delete(r.running, runId)
+	if r.runs[runId].State.IsDone() {
```
dbentley: It it's already done, shouldn't we not change its status? Like, if it were aborted, it shouldn't later change to timed out?

runner/local/simple_test.go
```diff
@@ -19,27 +21,24 @@ func TestRun(t *testing.T) {
 	assertWait(t, r, firstId, complete(0), "complete 0")
 }
 
-func TestSimul(t *testing.T) {
+func XTestSimul(t *testing.T) {
```
dbentley: X? So this isn't running?

runner/local/simple.go
```diff
@@ -2,69 +2,136 @@ package local
...
 	"github.com/scootdev/scoot/runner"
 	"github.com/scootdev/scoot/runner/execer"
-	"sync"
 )
 
 func NewSimpleRunner(exec execer.Execer) runner.Runner {
+	return NewMultiRunner(exec, 1)
+}
+
+func NewMultiRunner(exec execer.Execer, maxRunning int) runner.Runner {
```
dbentley: Why the change to support running multiple simultaneously?

This seems like a big change that makes things more complicated (e.g., we can no longer just have one checkout we reuse, we have to allocate). Is there something it enables right now?

I didn't see tests for this more complex functionality, but I think they're important. Did I miss them, or have you not done them yet?

runner/runner.go
```diff
@@ -85,8 +93,17 @@ type Runner interface {
...
 	// wait a very short period of time for cmd to finish
 	// Run may not wait indefinitely for cmd to finish. This is an async API.
-	Run(cmd *Command) (ProcessStatus, error)
+	Run(cmd *Command) ProcessStatus
 
 	// Status checks the status of run.
-	Status(run RunId) (ProcessStatus, error)
+	Status(run RunId) ProcessStatus
+
+	// Current status of all runs, running and finished.
```
dbentley: FWIW, I think it's fine if a Runner does throw out some statuses, so maybe say it's best effort for finished runs? WDYT?

workerapi/api.go
```diff
@@ -1,4 +1,89 @@
...
+	if thrift.Error != nil {
+		domain.Error = *thrift.Error
+	}
+	if thrift.ExitCode != nil {
+		domain.ExitCode = int(*thrift.ExitCode)
+	}
+	return domain
+}
+
+func DomainRunStatusToThrift(domain *runner.ProcessStatus) *worker.RunStatus {
```
dbentley: Tests?

(When I did this, Caitie suggested property-based testing, which is kinda nifty, but maybe overkill.)

I think table-based testing is easy to do with this kind of input-output so it's not super painful...

workerapi/server/server.go
```diff
@@ -21,99 +24,99 @@ func Serve(handler worker.Worker, addr string, transportFactory thrift.TTranspor
...
-	rs := worker.NewRunStatus()
-	rs.Status = status
-	rs.Info = &info
-	rs.ExitCode = &exitCode
-	return rs
+// Implement thrift worker.Worker interface.
+//
+func (h *handler) QueryWorker() (*worker.WorkerStatus, error) {
+	h.stat.Counter("workerQueries").Inc(1)
+	h.timeLastContact = time.Now()
```
dbentley: Why do this, instead of just updating the gauge directly right here?

workerapi/server/server.go
```diff
@@ -21,99 +24,99 @@ func Serve(handler worker.Worker, addr string, transportFactory thrift.TTranspor
...
-	wstatus.Running = []string{}
-	for runId, status := range h.runs {
-		if status.Status == worker.Status_RUNNING {
-			wstatus.Running = append(wstatus.Running, runId)
+// Periodically output stats
+func (h *handler) stats() {
+	ticker := time.NewTicker(time.Millisecond * time.Duration(500))
+	for {
+		select {
+		case <-ticker.C:
```
dbentley: Hrm, this code looks a little meh... It would be nice if the runner could just update the stats as it updates its code.... Probably not worth rejiggering now... But maybe the Runner could also accept a listener that just calls the stats? (so the runner doesn't have to know about stats for, like, a local runner?)

workerapi/worker.thrift
```diff
@@ -1,37 +1,42 @@
...
+  1: required list<RunStatus> runs  # All runs excepting what's been Erase()'d
+  2: optional string versionId      # Version associated with this worker.
+}
+
 struct RunCommand {
-  1: required list<string> argv
-  2: optional string snapshotId
-  3: optional i32 timeoutMs
+  1: required list<string> argv       # Binary followed by any number of arguments.
+  2: optional map<string,string> env  # Binary followed by any number of arguments.
```
dbentley: This doc looks wrong....

binaries/workerserver/main.go
```diff
@@ -26,7 +26,7 @@ func main() {
 	protocolFactory := thrift.NewTBinaryProtocolFactoryDefault()
 	transportFactory := thrift.NewTTransportFactory()
 
-	handler := server.NewHandler(stat.Scope("workerserver"))
+	handler := server.DefaultHandler(stat.Scope("workerserver"))
```
dbentley: Any chance I could convince you to undo DefaultHandler and require injecting the runner dependency? I know it looks a little worse right here, but I think it's really good if you have to pass the Runner, because it means that when you're looking at the workerserver, you know it's running something.

(It also means that we don't need to agonize over what the default should be)

dbentley:
LGTM

Yeah, that looks great.

I think the only really blocking comment is about the multiple runner. (If it's important for now, it should be tested)

runner/execer/execer.go
```diff
@@ -30,8 +30,7 @@ type Execer interface {
 type Process interface {
```
CaitieM20: What is the expected behavior of Wait & Abort.  Comment on what this interface represents and the expected behavior of methods is would be nice.

I assume Wait() is waiting for something to execute.  Should it potentially return an error then? What if Abort is called while we are Waiting?

runner/execer/os/os_execer.go
```diff
@@ -14,9 +16,13 @@ func NewExecer() execer.Execer {
...
+		return nil, errors.New("No command specified.")
```
CaitieM20: Should we return a specific error type so the caller can differentiate between bad request and something the execer couldn't do?

runner/runner.go
```diff
@@ -19,14 +19,18 @@ const (
 	// States below are end states
 	// a Process in an end state will not change its state
 
-	// Ran to completion
+	// Run completed, possibly with nonzero exit code.
 	COMPLETE
-	// Could not run to completion
+	// Could not complete, bad request or golang runtime err.
```
CaitieM20: I think we should differentiate between bad request and failed to run.  Having this distinction will help us abort sending poison pill messages around, or falsely marking nodes as dead

BadRequest means it should not be retried by the caller. 
FailedToRun is a retryable error.

workerapi/api.go
CaitieM20: Yea whatever is easiest here for testing is nice since the statespace is fairly limited a table test can work just as well!

workerapi/server/server.go
CaitieM20: Agree the typical pattern in most Twitter code is to pass the stats object in, and then just update next to the code its effecting.

This is also really nice when you are debugging code and trying to figure out what statistic names correspond to what part of the code.

workerapi/server/server.go
CaitieM20: Agree why not update the stats next to the code they are tracking thats a more common pattern.

workerapi/worker.thrift
```diff
@@ -1,37 +1,42 @@
...
-//TODO: add useful load information when it comes time to have multiple runs.
-struct WorkerStatus {
-  1: required list<string> running
+  TIMEDOUT = 6
 }
 
 struct RunStatus {
   1: required Status status
-  2: optional string runId
+  2: required string runId
```
CaitieM20: Does RunId correspond to task or jobId?  Do these need to be unique?

workerapi/worker.thrift
```diff
@@ -1,37 +1,42 @@
...
   6: optional i32 exitCode
 }
 
+// VersionId will be empty by default. To change this, pass a getVersion() func to server.Handler.
+// - Semantics are up to the implementer, this needn't be a library version, it could be a gitSha, etc.
+//
+// TODO: add useful load information when it comes time to have multiple runs.
+struct WorkerStatus {
+  1: required list<RunStatus> runs  # All runs excepting what's been Erase()'d
+  2: optional string versionId      # Version associated with this worker.
```
CaitieM20: what is versionId tracking.  Version of what?

workerapi/worker.thrift
```diff
@@ -1,37 +1,42 @@
...
 }
 
 //TODO: add a method to kill the worker if we can articulate unrecoverable issues.
 service Worker {
-  WorkerStatus QueryWorker()
-  RunStatus Run(1: RunCommand cmd)
-  RunStatus Query(1: string runId)
-  RunStatus Abort(1: string runId)
+  WorkerStatus QueryWorker()         # Overall worker node status.
+  RunStatus Run(1: RunCommand cmd)   # Run a command and return job Status.(RUNNING|FAILED).
```
CaitieM20: nit: delete Running|Failed since there are more potential statuses that could be returned

daemon/client/conn/conn.go
CaitieM20: I agree I'm not sure the best way to proceed here, but Im a fan of starting with the simplest thing that works, so the current idea of this just being another implementation of Runner works for now.  And then we can iterate if this becomes problematic.  

It also makes it really clear in code that this is the equivalent to the Runner in the CloudWorker.  Which is a nice parallel to draw.

runner/execer/os/os_execer.go
jasonbs10: Done.

runner/fake/fake.go
jasonbs10: Let's do that in a new PR, it affects other tests.

runner/local/simple.go
jasonbs10: Done.

runner/local/simple.go
jasonbs10: Fixed.

runner/local/simple_test.go
jasonbs10: I forget to re-enable it....

runner/runner.go
jasonbs10: Ehhh, -1, I prefer completeness.

workerapi/server/server.go
jasonbs10: I don't do this at the call sites since I'm measuring an incrementing time and if there's no rpc (which is what I'm measuring), the stats become stale.

workerapi/server/server.go
jasonbs10: Added a TODO to move this into an extended runner.

workerapi/worker.thrift
jasonbs10: Fixed.

binaries/workerserver/main.go
jasonbs10: Absolutely, done.

runner/execer/os/os_execer.go
jasonbs10: Sure, I've added BADREQUEST, but without the transient connotation.

runner/runner.go
jasonbs10: Updated this comment.

workerapi/worker.thrift
jasonbs10: Commented. This runId is only unique within each worker

workerapi/worker.thrift
jasonbs10: Oh, there's a comment above noting this could be sha1 etc.

workerapi/worker.thrift
jasonbs10: Done.

runner/execer/execer.go
jasonbs10: Added some very basic comments. To elaborate slightly, every combination of calls returns process.FAILED except a solitary call to Wait() when the process exits cleanly.

runner/local/simple.go
dbentley: Seems like it's still here?

jasonbs10: I usually push after finishing with comments... I still have to address the missing tests. 






On Sat, Jul 30, 2016 at 11:56 AM -0700, "Dan Bentley" <notifications@github.com> wrote:












In runner/local/simple.go:
>  )
>  
>  func NewSimpleRunner(exec execer.Execer) runner.Runner {
> +	return NewMultiRunner(exec, 1)
> +}
> +
> +func NewMultiRunner(exec execer.Execer, maxRunning int) runner.Runner {



Seems like it's still here?

workerapi/api.go
jasonbs10: Done.

jasonbs10:
Ok, I think I addressed most of the comments. Let's do a final lgtm if it looks good.

dbentley:
LGTM

PR #68, opened 2016-08-01
First pass at hooking up scheduler and workerserver.

jasonbs10:
Not quite done but ready for comments.

### DIFF
--- a/binaries/scheduler/main.go
+++ b/binaries/scheduler/main.go
@@ -1,16 +1,21 @@
 package main
 
 import (
+	"log"
+	"sync"
+
 	"github.com/apache/thrift/lib/go/thrift"
+	//clusterdef "github.com/scootdev/scoot/cloud/cluster"
 	clusterimpl "github.com/scootdev/scoot/cloud/cluster/memory"
 	"github.com/scootdev/scoot/saga"
 	"github.com/scootdev/scoot/sched/distributor"
 	queueimpl "github.com/scootdev/scoot/sched/queue/memory"
 	"github.com/scootdev/scoot/sched/scheduler"
+	//"github.com/scootdev/scoot/sched/worker"
+
 	"github.com/scootdev/scoot/sched/worker/fake"
+	// "github.com/scootdev/scoot/sched/worker/rpc"
 	"github.com/scootdev/scoot/scootapi/server"
-	"log"
-	"sync"
 )
 
 func main() {
@@ -32,6 +37,11 @@ func main() {
 	// TODO: Replace with Durable SagaLog, currently In Memory Only
 	sagaCoordinator := saga.MakeInMemorySagaCoordinator()
 	sched := scheduler.NewScheduler(dist, sagaCoordinator, fake.MakeWaitingNoopWorker)
+	//TODO: real workers.
+	// workerFactory := func(node clusterdef.Node) worker.Worker {
+	// 	return rpc.NewThriftWorker(transportFactory, protocolFactory, string(node.Id()))
+	// }
+	// sched := scheduler.NewScheduler(dist, sagaCoordinator, workerFactory)
 	// TODO: Replace with Durable WorkQueue, currently in Memory Only
 	workQueue := queueimpl.NewSimpleQueue(1000)
 
--- a/binaries/scheduler/main.go
+++ b/binaries/scheduler/main.go
@@ -1,33 +1,61 @@
 package main
 
 import (
+	"flag"
+	"fmt"
 	"log"
+	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/apache/thrift/lib/go/thrift"
-	//clusterdef "github.com/scootdev/scoot/cloud/cluster"
+	clusterdef "github.com/scootdev/scoot/cloud/cluster"
 	clusterimpl "github.com/scootdev/scoot/cloud/cluster/memory"
 	"github.com/scootdev/scoot/saga"
 	"github.com/scootdev/scoot/sched/distributor"
 	queueimpl "github.com/scootdev/scoot/sched/queue/memory"
 	"github.com/scootdev/scoot/sched/scheduler"
+	"github.com/scootdev/scoot/sched/worker"
 	//"github.com/scootdev/scoot/sched/worker"
 
 	"github.com/scootdev/scoot/sched/worker/fake"
+	"github.com/scootdev/scoot/sched/worker/rpc"
 	// "github.com/scootdev/scoot/sched/worker/rpc"
 	"github.com/scootdev/scoot/scootapi/server"
 )
 
+var addr = flag.String("addr", "localhost:9090", "Bind address for api server.")
+var workers = flag.String("workers", "", "Comma separated list of workers (host:port,...)|NUM:mem.")
+
 func main() {
 	log.Println("Starting Cloud Scoot API Server & Scheduler")
-	addr := "localhost:9090"
+
+	flag.Parse()
+	workersList := strings.Split(*workers, ",")
+	workerNodes := []clusterdef.Node{}
+	if len(workersList) == 1 && (workersList[0] == "" || strings.Contains(workersList[0], ":mem")) {
+		//Keep the original behavior for now if no workers specified on cmdline.
+		numNodes := 10
+		if workersList[0] != "" {
+			numNodes, _ = strconv.Atoi(strings.Split(workersList[0], ":")[0])
+		}
+		workersList = []string{}
+		for idx := 0; idx < numNodes; idx++ {
+			workersList = append(workersList, fmt.Sprintf("inmemory%d", idx))
+		}
+	}
+	for _, worker := range workersList {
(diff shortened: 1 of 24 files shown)


sched/worker/worker.go
```diff
@@ -5,44 +5,19 @@ import (
...
 
-// WorkerController lets the Scheduler control a worker.
-type WorkerController interface {
+// Worker controller gives the Scheduler a generic way to complete work.
+type Worker interface {
 	// TODO(dbentley): include more info on positive results
 	RunAndWait(task sched.TaskDefinition) error
 }
 
 type workerNode struct {
```
dbentley: Ooof, I thought I got rid of this... Do we still need workerNode?

sched/worker/worker.go
```diff
@@ -5,44 +5,19 @@ import (
 	"github.com/scootdev/scoot/sched"
 )
 
-// Create a WorkerController that talks to node
-type WorkerFactory func(node cluster.Node) WorkerController
+// Create a Worker controller that talks to node
```
dbentley: Drop controller?

workerapi/client/cli.go
```diff
@@ -1,36 +1,30 @@
...
 )
 
-type Client struct {
-	rootCmd          *cobra.Command
-	addr             string // modified by flag parsing
-	transportFactory thrift.TTransportFactory
-	protocolFactory  thrift.TProtocolFactory
-	client           *worker.WorkerClient
+type cliClient struct {
+	client
```
dbentley: And then have this not be incorporating, but just have client be of type *client...

That will take a little more typing in commands.go, but really not much. c.Dial() will return a client.

(Note: this separation is why the daemon/client is split up into conn vs cli... Maybe worth looking at)

workerapi/client/client.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+	"log"
+
+	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/scootdev/scoot/runner"
+	"github.com/scootdev/scoot/workerapi"
+	"github.com/scootdev/scoot/workerapi/gen-go/worker"
+)
+
+type Client interface {
```
dbentley: This, and the fact that you have to panic in one, is code smell.

Fortunately, it's easy to avoid. Just have NewCliClient return a *CliClient.

workerapi/client/commands.go
```diff
@@ -3,46 +3,42 @@ package client
...
 	runcmd.Argv = strings.Split(*runArgv, ",")
 	log.Println("Calling run rpc to cloud worker", args, render.Render(runcmd))
 
-	client, err := c.Dial()
-	if err != nil {
-		return err
-	}
-
-	status, err := client.Run(runcmd)
+	status := c.Run(runcmd)
```
dbentley: Wait, there has to be some kind of error, right? Like, what happens if there was just an error sending across the network, because we're not connected? We shouldn't have that just disappear as a status, right?

sched/worker/worker.go
jasonbs10: I'll remove it for now. In order for work to flow through though, we need a way to convert Node.Id to an address etc. We may have decided that can all go in cluster.Node, I don't remember.

sched/worker/worker.go
jasonbs10: -0, I kept it there intentionally because it helps clarify the purpose vs code living under /workerapi.

workerapi/client/client.go
jasonbs10: Done.

workerapi/client/commands.go
jasonbs10: Errors are incorporated into runner.ProcessStatus as came up in the previous PR. I only print it because I didn't necessarily see the value in returning it to cobra. I'm not 100% comfortable with my decision to remove err from a bunch of function signatures, so I can be convinced to add it back.

workerapi/client/cli.go
jasonbs10: I'm not sure about this, I'll ping you in hipchat to clarify.

workerapi/client/commands.go
dbentley: I think one lesson of the history of RPCs is it's bad to make RPCs look like normal function calls. So, the way I would think of it:

If you get a ProcessStatus, it comes from some runner somewhere.

If you couldn't even talk to the runner (e.g., RPC or network fail), then it's an error.

But also don't worry about blocking this review on this, but please add a TODO if you don't think it's worth doing now so we can keep track.

jasonbs10: Sure, I'll add it back to the thrift receive boundary, Client in this case, and we can maybe revisit the thrift write boundary later if desired.

binaries/scheduler/main.go
```diff
@@ -1,33 +1,61 @@
...
+var addr = flag.String("addr", "localhost:9090", "Bind address for api server.")
+var workers = flag.String("workers", "", "Comma separated list of workers (host:port,...)|NUM:mem.")
+
 func main() {
 	log.Println("Starting Cloud Scoot API Server & Scheduler")
-	addr := "localhost:9090"
+
+	flag.Parse()
+	workersList := strings.Split(*workers, ",")
+	workerNodes := []clusterdef.Node{}
```
dbentley: I like how you let people specify different behaviors in one flag.

This is great for now, so consider what I'm about to say just food for thought for moving forward.

I suspect we'll also want a way to change the cluster implementation (so you can test locally or with static cluster or what not).

I'd make a function like:

func parseCluster(spec string) (cluster.Cluster, workerFactory)

And then this one string will control:
a) static cluster of talking to service discovery
b) what the initial nodes are
c) what worker factory to use

I think that also makes it easy to think about what the language (yes, one string to this is a DSL!) is like.

One thing I've seen work is just using JSON (and golang's easy JSON parsing) to do something like:

{"cluster": {"type": "static"},
 "initial_nodes": ["localhost:2345", "localhost:2346"],
 "workerFactory": "thrift"}

Then there's a parse func that's string -> ClusterSpec, and then another func that's ClusterSpec -> (Cluster, WorkerFactory)

Like I said, none of this is necessary for right now, but worth thinking about how to balance:
a) easy to specify
b) flexible enough to let you specify whatever
c) making the code clear and compartmentalized

sched/scheduler/scheduler_test.go
```diff
@@ -2,6 +2,8 @@ package scheduler
 
 import (
 	"fmt"
+	"testing"
```
dbentley: It seems like we're pushing towards an import style.

Do we have a (living) doc of our coding style? Maybe in our repo?

I'd even be fine having one more level, so it's like:

<stdlib imports>

<3rdparty imports>

<scoot imports>

sched/worker/rpc/thriftWorker.go
```diff
@@ -25,17 +25,24 @@ func NewThriftWorker(
...
+			return err
+		}
 		if workerStatus.Error != nil {
 			return workerStatus.Error
 		}
 		for _, s := range workerStatus.Runs {
 			if s.RunId == status.RunId {
 				status = s
-				continue
+				goto Loop
```
dbentley: Is there a reason to use goto instead of a loop?

workerapi/client/cli.go
```diff
@@ -5,26 +5,30 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type CliClient interface {
+	Cli() error
+}
+
 type cliClient struct {
-	client
+	client  client
```
dbentley: Thanks!

dbentley:
LGTM

Is there anything big you think's left?

binaries/scheduler/main.go
jasonbs10: Added a TODO. This is similar to what we discussed for scootfs, so we'll probably want to stick to one format, and json seems fine.

sched/scheduler/scheduler_test.go
jasonbs10: I don't really look at this beyond the block syntax. I let goimports and gofmt handle this. I'm not sure those tools even let you decide how to arrange it?

sched/worker/rpc/thriftWorker.go
jasonbs10: Hmm, I'm not sure what you mean. I changed it to used a 'var updated bool' since that convention doesn't usually raise any questions.

workerapi/client/cli.go
jasonbs10: This is actually cleaner, I realized, because the caller of NewCliClient doesn't care about the rpc calls so there's no need to expose those.

jasonbs10:
I'm going to submit this as-is. We'll need another PR that moves logic out of the scheduler's main.go and into a lib, plus a script to do an end-to-end run which we can add to 'make test'.

sched/scheduler/scheduler_test.go
dbentley: Ohhhh! goimports!

Yes, gofmt preserves blocks, so if you leave a blank line, it will sort them within that block.

PR #70, opened 2016-08-09
Initial sched config.

jasonbs10:


dbentley:
LGTM

I think this'll evolve, but it'd be good to unblock for now...

sched/config/config.go
```diff
@@ -0,0 +1,449 @@
...
+	"github.com/apache/thrift/lib/go/thrift"
+	clusterdef "github.com/scootdev/scoot/cloud/cluster"
+	clusterimpl "github.com/scootdev/scoot/cloud/cluster/memory"
+	"github.com/scootdev/scoot/saga"
+	"github.com/scootdev/scoot/sched/distributor"
+	"github.com/scootdev/scoot/sched/queue"
+	queueimpl "github.com/scootdev/scoot/sched/queue/memory"
+	"github.com/scootdev/scoot/sched/scheduler"
+	"github.com/scootdev/scoot/sched/worker"
+	//"github.com/scootdev/scoot/sched/worker"
```
CaitieM20: nit: remove commented out imports

CaitieM20:
LGTM.

One future thinking thing that we may want to add to the backlog.  In prod we probably never want to use the default plugins.  Signals a configuration error.  We could add a flag to the aurora config based on what environment its deployed in and disallow the use of plugins in prod.

CaitieM20:
One more note can we add a test case to prove the config files parse correctly.  I've found this super valuable in the past

PR #71, opened 2016-08-10
sched/config refactor

dbentley:
Separate plugins/built-ins

Separate core logic into config.go vs. built-ins into defaults.go

Now there is a Parser that is configurable that will create a Config.

### DIFF
--- a/binaries/scheduler/main.go
+++ b/binaries/scheduler/main.go
@@ -7,35 +7,30 @@ import (
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/scootdev/scoot/sched/config"
 	"github.com/scootdev/scoot/scootapi/server"
-	"golang.org/x/net/context"
 )
 
 var addr = flag.String("addr", "localhost:9090", "Bind address for api server.")
-var cfgFile = flag.String("cfg", "", "Path to sched overlay config file.")
+var cfgText = flag.String("sched_config", "", "Scheduler Configuration.")
 
 func main() {
 	log.Println("Starting Cloud Scoot API Server & Scheduler")
 	flag.Parse()
 
-	plugins := config.Plugins{}
-	ctx := context.Background()
-	transportFactory := thrift.NewTTransportFactory()
-	protocolFactory := thrift.NewTBinaryProtocolFactoryDefault()
+	cfg := config.Default()
 
 	// Construct scootapi server handler based on config.
-	cfg, err := config.DefaultConfig(*cfgFile, plugins)
-	if err != nil {
-		log.Fatal("Error configuring Scoot API: ", err)
-	}
 
-	handler, err := config.ConfigureCloudScoot(cfg, ctx, transportFactory, protocolFactory)
+	handler, err := cfg.Create([]byte(*cfgText))
 	if err != nil {
 		log.Fatal("Error configuring Scoot API: ", err)
 	}
 
 	// Start API Server
 	log.Println("Starting API Server")
-	err = server.Serve(handler, *addr, transportFactory, protocolFactory)
+
+	err = server.Serve(handler, *addr,
+		thrift.NewTTransportFactory(),
+		thrift.NewTBinaryProtocolFactoryDefault())
 	if err != nil {
 		log.Fatal("Error serving Scoot API: ", err)
 	}
--- a/binaries/scheduler/main.go
+++ b/binaries/scheduler/main.go
@@ -16,11 +16,11 @@ func main() {
 	log.Println("Starting Cloud Scoot API Server & Scheduler")
 	flag.Parse()
 
-	cfg := config.Default()
+	parser := config.DefaultParser()
 
 	// Construct scootapi server handler based on config.
 
-	handler, err := cfg.Create([]byte(*cfgText))
+	handler, err := parser.Create([]byte(*cfgText))
 	if err != nil {
 		log.Fatal("Error configuring Scoot API: ", err)
 	}
--- a/sched/config/config.go
+++ b/sched/config/config.go
@@ -3,445 +3,201 @@ package config
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
-	"log"
 
-	"github.com/apache/thrift/lib/go/thrift"
-	clusterdef "github.com/scootdev/scoot/cloud/cluster"
-	clusterimpl "github.com/scootdev/scoot/cloud/cluster/memory"
+	"github.com/scootdev/scoot/cloud/cluster"
 	"github.com/scootdev/scoot/saga"
 	"github.com/scootdev/scoot/sched/distributor"
 	"github.com/scootdev/scoot/sched/queue"
-	queueimpl "github.com/scootdev/scoot/sched/queue/memory"
 	"github.com/scootdev/scoot/sched/scheduler"
 	"github.com/scootdev/scoot/sched/worker"
-
-	"github.com/scootdev/scoot/sched/worker/fake"
-	"github.com/scootdev/scoot/sched/worker/rpc"
 	"github.com/scootdev/scoot/scootapi/gen-go/scoot"
 	"github.com/scootdev/scoot/scootapi/server"
-
-	"golang.org/x/net/context"
 )
 
-//
-//TODO: handle and propagate all config values. handle errors. decide extensibility story.
-//
-
-// Return a Config where each field is initialized only if there is no plugin for it.
-func DefaultConfig(cfgFile string, plugins Plugins) (*Config, error) {
-	rawCfg, _ := defaultConfig(cfgFile)
-	realizedCfg := Config{raw: rawCfg, plugins: plugins}
-
-	//TODO: catch errors unmarshaling.
-	if _, ok := plugins[ClusterTypeId]; !ok {
-		var cfg ClusterConfig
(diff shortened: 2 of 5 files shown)


binaries/scheduler/main.go
```diff
@@ -7,35 +7,30 @@ import (
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/scootdev/scoot/sched/config"
 	"github.com/scootdev/scoot/scootapi/server"
-	"golang.org/x/net/context"
 )
 
 var addr = flag.String("addr", "localhost:9090", "Bind address for api server.")
-var cfgFile = flag.String("cfg", "", "Path to sched overlay config file.")
+var cfgText = flag.String("sched_config", "", "Scheduler Configuration.")
```
jasonbs10: Why change this to text? Seems unwieldy. Maybe have it be textConfig or fileConfig based on contents.

binaries/scheduler/main.go
```diff
@@ -7,35 +7,30 @@ import (
...
 
 func main() {
 	log.Println("Starting Cloud Scoot API Server & Scheduler")
 	flag.Parse()
 
-	plugins := config.Plugins{}
-	ctx := context.Background()
-	transportFactory := thrift.NewTTransportFactory()
-	protocolFactory := thrift.NewTBinaryProtocolFactoryDefault()
+	cfg := config.Default()
```
jasonbs10: It's slightly confusing to have config.Default() return a parser. Rename to DefaultParser()?

sched/config/config.go
```diff
@@ -3,445 +3,199 @@ package config
...
-	if _, ok := plugins[HealthTypeId]; !ok {
-		var cfg HealthConfig
-		realizedCfg.Health = &cfg
-		json.Unmarshal(rawCfg.Health, &realizedCfg.Health)
-	}
-
-	return &realizedCfg, nil
+// Config is the top-level configuration for the Scheduler. It defines how to create
+// each of our (configurable) dependencies and then creates the top-level handler.
+type Config struct {
```
jasonbs10: Any reason for getting rid of the Stats and Health configs?

binaries/scheduler/main.go
dbentley: Yeah, good question.

1) It's not super hard in a shell script to pass an arbitrarily-large flag.
2) It is harder to have to create a file just to hold something
3) I agree. If you pass "file:foo.txt" it should read from foo.txt. I put a TODO about this in config.go (at the top of the function Parse)

binaries/scheduler/main.go
dbentley: Done.

sched/config/config.go
```diff
@@ -3,445 +3,199 @@ package config
...
+	Queue   QueueConfig
+	Cluster ClusterConfig
+	SagaLog SagaLogConfig
```
jasonbs10: I was on the fence between having a SagaLog config vs SagaCoord (and switched going from v1 to v2). Is the only configurable item in SagaCoord only ever going to be SagaLog?

sched/config/config.go
dbentley: It seemed like they weren't being actively used; did I misread the code?

sched/config/config.go
dbentley: If we start making the coordinator configurable, then you'll want to configure it separately from the underlying log. So if we do that, we'll change the top-level from log to coord, and then the coord will have an element that's saga log config.

sched/config/config.go
jasonbs10: I guess we lose the ability to inspect/modify the config, which is fine I guess.

sched/config/config.go
```diff
@@ -3,445 +3,199 @@ package config
...
-
-## Example json. Any subset can be specified to override specific defaults.
-##
+func (p *Parser) Parse(configText []byte) (*Config, error) {
+	// TODO(dbentley): allow this to refer to a file instead of being a giant string
+	if len(configText) == 0 {
+		configText = emptyJson
+	}
+	var cfg topLevelConfig
+	err := json.Unmarshal(configText, &cfg)
```
jasonbs10: So no overriding a single value within any of the concrete config types? We should document somewhere that a complete set of config values is required.

sched/config/config.go
dbentley: I hope not! Keeping that was super-important to me since you explained it. Let me show you two tricks this allows, in the realm of inspecting/modifying the config:
1)
cfg := config.DefaultParser().Parse(nil)
json, err := json.Marshal(cfg)

2)
(cfg.Queue.(QueueMemoryConfig)).Capacity = 3

So you can inspect and modify, you just have to do so with type casting.

But this is good, because it means that you have the same code to deal with built-in configs vs. injected configs.

sched/config/config.go
```diff
@@ -3,445 +3,199 @@ package config
...
-##
-{
- "cluster": {"type": "memory",  "initial": 5, "count": 10} ||
-            {"type": "static",  "initial": 5, "hosts": [":2345", ":2346"]} ||
-            {"type": "dynamic", "initial": 5, "max": 6},
-
- "queue": {"type": "fake", "count": 1000} ||
-          {"type": "memory", "capacity": 1000} ||
-          {"type": "durable"}
+	queueType, queueData := parseType(cfg.Queue)
```
jasonbs10: I like that you've consolidated the the code iterating over config types. Is there any easy way to make this less verbose?

sched/config/defaults.go
```diff
@@ -0,0 +1,103 @@
...
+
+type SagaLogMemoryConfig struct {
+	Type string
+}
+
+func (c *SagaLogMemoryConfig) Create() (saga.SagaLog, error) {
+	return saga.MakeInMemorySagaLog(), nil
+}
+
+type LocalWorkersConfig struct {
```
jasonbs10: Are we planning to add back runner/exec config later?

sched/config/config.go
dbentley: Well, it's up to how you configure the parser. If you set "memory" to have a default...

Basically, this is split so there's a framework (config.go) and then a configuration of it (defaults.go)...

binaries/scheduler/main.go
```diff
@@ -7,35 +7,30 @@ import (
...
+	parser := config.DefaultParser()
 
 	// Construct scootapi server handler based on config.
-	cfg, err := config.DefaultConfig(*cfgFile, plugins)
-	if err != nil {
-		log.Fatal("Error configuring Scoot API: ", err)
-	}
 
-	handler, err := config.ConfigureCloudScoot(cfg, ctx, transportFactory, protocolFactory)
+	handler, err := parser.Create([]byte(*cfgText))
```
jasonbs10: How are we planning to cleanly stop this if we don't pass a context? Are we punting on that?

jasonbs10:
lgtm

sched/config/config.go
jasonbs10: Ah, ok, yeah this is the one case where we would have to cast (I like that json.Unmarshal is the only code that needs to reflect/cast the interfaces).

jasonbs10: Health and Stats weren't hooked up only because we have those endpoints running in source, but for whatever reason weren't in the OSS binary. It's just one more thing we can remove from the main file if implemented.

sched/config/config.go
jasonbs10: I more meant keeping parser Defaults and having a specific nested value that you want to change from cli. There's no way to do that afaik.  Not a big deal, just something to note. Am I still misunderstanding?

sched/config/config.go
jasonbs10: So no way to change something in coord without also having to specify sagaLog? This is probably the only case that would make me prefer the original code, but it's a minor use case.

sched/config/config.go
dbentley: Is it alright if I add them as I'm about to use them in the binary, so we can talk about them as we use them?

sched/config/config.go
dbentley: There are a few ways to handle this, but I think it's not worth trying to design right now...

(E.g., we could have that you specify the log separately, so it can still use the default)

I think it depends what we need to configure in the coordinator.

sched/config/config.go
dbentley: It depends on exactly how json.Unmarshal works, and if you pass a filled-in FooConfig, but, yeah... I think you're generally right.

sched/config/defaults.go
dbentley: I didn't see that before; my bad. Sure, added a comment.

binaries/scheduler/main.go
dbentley: Yup, we're punting until we're ready to use it.

sched/config/config.go
jasonbs10: I'd rather just add those endpoints now but this all can wait until we're ready to add all the metrics to sched.

sched/config/config.go
dbentley: I hoisted out parseDomain, but it ends up being a little messy still because of Go's lack of generics; WDYT?

jasonbs10: This is awkward in go. I'd stick with the original.

PR #72, opened 2016-08-15
Moved wily cluster code to OSS and refactored. Added smoketest script.

jasonbs10:


Makefile
```diff
@@ -52,3 +54,8 @@ clean: clean-mockgen
 	go clean ./...
 
 fullbuild: dependencies generate test
+
+install:
```
dbentley: go install ./binaries/... ?

binaries/scheduler/main.go
```diff
@@ -17,6 +18,11 @@ func main() {
 	flag.Parse()
 
 	parser := config.DefaultParser()
+	parser.Workers[""] = &config.RPCWorkersConfig{}
```
dbentley: If this is going to be the default you should give it a full config. E.g., set Type.

(This means that if you try to marshal it with json, you get a working config)

cloud/cluster/local/config.go
```diff
@@ -0,0 +1,15 @@
+package local
+
+import (
+	"github.com/scootdev/scoot/cloud/cluster"
+	"github.com/scootdev/scoot/cloud/cluster/memory"
+)
+
+type ClusterLocalConfig struct {
+	RegexCapturePort string
```
dbentley: Document this parameter.

(Also, why is it configurable? Do we need configurability right? If not, I'd just hardcode it saying "yes, this is ugly, but it works for now)

cloud/cluster/node.go
```diff
@@ -3,7 +3,28 @@ package cluster
 type NodeId string
 
 type Node interface {
+	// A unique node identifier, like 'host:port'
 	Id() NodeId
+
+	// Comparator for sorting.
+	Less(Node) bool
```
dbentley: Can we just sort on the ID?

cloud/cluster/subscription.go
```diff
@@ -0,0 +1,153 @@
...
+		outCh:    make(chan []NodeUpdate),
+		outgoing: nil,
+	}
+	go c.loop()
+	return c.outCh
+}
+
+// diffSubscription fetches the list of nodes periodically, diffs the list to determine
+// what has changes, and sends updates.
+type diffSubscription struct {
```
dbentley: Should this just implement Cluster?

scripts/smoketest/smoketest.go
```diff
@@ -0,0 +1,123 @@
...
+	signal.Notify(sigchan, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		var s os.Signal
+		s = <-sigchan
+		fmt.Printf("signal %s received, shutting down", s)
+		killcmds()
+		os.Exit(1)
+	}()
+
+	for idx := 0; idx < *numWorkers; idx++ {
```
dbentley: idx -> i?

scripts/smoketest/smoketest.go
```diff
@@ -0,0 +1,123 @@
...
+	"os/exec"
+	"os/signal"
+	"path"
+	"path/filepath"
+	"strconv"
+	"syscall"
+)
+
+//TODO(jschiller)
+var remote = flag.String("remote", "", "Empty string to run locally. Otherwise: 'DC/ROLE/ENV'")
```
dbentley: I'd remove for now; I don't think it even makes sense in OSS Scoot.

Makefile
jasonbs10: Ah, I tried 'go install binaries/...' and that didn't work. Yeah, your way is better (and it doesn't unconditionally build every time).

binaries/scheduler/main.go
jasonbs10: Done.

cloud/cluster/local/config.go
jasonbs10: Done.

cloud/cluster/node.go
jasonbs10: You had a custom sort by addr then port for wilyNode and I didn't want to remove that ability. I assume you had a reason to not sort by id?

cloud/cluster/subscription.go
jasonbs10: I think cluster and subscription are logically different things and it makes some sense to keep them separate. We can discuss code organization offline.

scripts/smoketest/smoketest.go
jasonbs10: Hmm, ok, we can be more consistent about short names, I'll try to get adjust my habits.

scripts/smoketest/smoketest.go
jasonbs10: Oh, this was actually meant for the twitter repo. I'm planning to do a corresponding script there with this flag.

cloud/cluster/node.go
dbentley: I think just because that's how it was stored in that iteration. Is there an example where sorting by addr then port is different than sorting by ID?

cloud/cluster/subscription.go
dbentley: Subscription is something you get from Cluster in the current Cluster interface.

It seems like basically there are two ways to deal with clusters:
a) all the members at once
b) deltas

And Cluster lets you read in either way.

Maybe memoryCluster should just have two input channels:
1) all the Nodes at once
2) updates

dbentley:
LGTM

PR #73, opened 2016-08-15
smoke_test: change generator to produce <=10 sim'able tasks

dbentley:
Change sim to take a sleep task

### DIFF
--- a/scootapi/testhelpers/generators.go
+++ b/scootapi/testhelpers/generators.go
@@ -32,7 +32,7 @@ func GenJobDefinition(rng *rand.Rand) *scoot.JobDefinition {
 func GenTask(rng *rand.Rand) *scoot.TaskDefinition {
 
 	cmd := scoot.NewCommand()
-	cmd.Argv = []string{"sleep", "1"}
+	cmd.Argv = []string{"sleep 500", "complete 0}
 
 	taskDef := scoot.NewTaskDefinition()
 	taskDef.Command = cmd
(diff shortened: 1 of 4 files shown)


scootapi/testhelpers/generators.go
```diff
@@ -32,7 +32,7 @@ func GenJobDefinition(rng *rand.Rand) *scoot.JobDefinition {
 func GenTask(rng *rand.Rand) *scoot.TaskDefinition {
 
 	cmd := scoot.NewCommand()
-	cmd.Argv = []string{"sleep", "1"}
+	cmd.Argv = []string{"sleep 500", "complete 0"}
```
jasonbs10: I think we need to be more opinionated and/or more configurable with regard to testing. Do you want to make the default runner defer to simExec? Or complete the TODO of adding configuration for runner and worker, and specify those in our smoketests?

jasonbs10: Right, but this PR could do more to be end-to-end. Unless I'm misreading, we use os/execer and by redefining tasks here, you'll break anything that currently runs scootapi.smoketest.

dbentley: OK, how's that?

dbentley: (To clarify: I changed workerserver/main.go to now use a SimExecer)

jasonbs10: That should be ok.

jasonbs10:
lgtm

PR #74, opened 2016-08-17
Moved most of the smoketest script into a common lib.

jasonbs10:
Yea or Nay?

### DIFF
--- a/scripts/common/smoketest.go
+++ b/scripts/common/smoketest.go
@@ -1,0 +1,155 @@
+package common
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"net"
+	"os"
+	"os/exec"
+	"os/signal"
+	"path"
+	"path/filepath"
+	"strconv"
+	"syscall"
+)
+
+type SmokeTest []*exec.Cmd
+
+type SmokeTestOpts struct {
+	LogDir     string
+	RepoDir    string
+	NumWorkers int
+	Compile    func() error
+	Run        func() error
+}
+
+func (s *SmokeTest) InitOptions() (*SmokeTestOpts, error) {
+	logDir := flag.String("logdir", "/tmp/scoot-smoketest", "If empty, write to stdout, else to log file")
+	repoDir := flag.String("repo", "$GOPATH/src/github.com/scootdev/scoot", "Path to scoot repo")
+	numWorkers := flag.Int("num_workers", 5, "Number of workerserver instances to spin up.")
+
+	flag.Parse()
+	if *numWorkers < 5 {
+		return nil, errors.New("Need >5 workers (see scheduler.go:getNumNodes)")
+	}
+	compile := func() error { return s.compile(*repoDir) }
+	run := func() error { return s.run(*numWorkers) }
+	opts := &SmokeTestOpts{*logDir, *repoDir, *numWorkers, compile, run}
+	return opts, nil
+}
+
+func (s *SmokeTest) Fatalf(format string, v ...interface{}) {
+	fmt.Printf(format+"\n", v...)
+	s.Kill()
+}
+
+func (s *SmokeTest) Kill() {
+	for _, cmd := range *s {
+		if cmd.ProcessState == nil && cmd.Process != nil {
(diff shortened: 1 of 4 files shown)


scripts/common/smoketest.go
```diff
@@ -0,0 +1,155 @@
...
+	"syscall"
+)
+
+type SmokeTest []*exec.Cmd
+
+type SmokeTestOpts struct {
+	LogDir     string
+	RepoDir    string
+	NumWorkers int
+	Compile    func() error
```
dbentley: Oh cool, I see what you're doing. It's a shame that golang makes it hard to mix functions and data, because an interface just to hold data feels weird, but a struct to hold functions also feels weird...

scripts/common/smoketest.go
```diff
@@ -0,0 +1,155 @@
...
+	"os"
+	"os/exec"
+	"os/signal"
+	"path"
+	"path/filepath"
+	"strconv"
+	"syscall"
+)
+
+type SmokeTest []*exec.Cmd
```
dbentley: Now that I think about it, smoke test is already used in the scootapi binary and this is kinda the same, but kinda different:

You can run a scootapi smoke test against aurora scheduler/workers (or against local scheduler and aurora workers).

WDYT of giving this a different name to emphasize that? E.g., "swarm test", because this creates a swarm.

scripts/common/smoketest.go
```diff
@@ -0,0 +1,155 @@
...
+
+type SmokeTestOpts struct {
+	LogDir     string
+	RepoDir    string
+	NumWorkers int
+	Compile    func() error
+	Run        func() error
```
dbentley: Thought: maybe this binary is the orchestrator (bring up the cluster, and eventually start killing, etc.), but not the runner, so it just starts the cluster...

Anyway, that's a bigger refactor and possibly unnecessary, but a thought.

scripts/common/smoketest.go
```diff
@@ -0,0 +1,155 @@
...
+	*s = append(*s, cmd)
+	if err := cmd.Start(); err != nil {
+		return fmt.Errorf("Error starting %v,%v: %v", bin, args, err)
+	}
+	fmt.Println("Started: ", name, args)
+
+	if !blocking {
+		go func() {
+			if err := cmd.Wait(); err != nil {
+				s.Fatalf("Exiting %v,%v: %v", bin, args, err)
```
dbentley: I'd change it to "exiting because cmd ended early"

scripts/common/smoketest.go
```diff
@@ -0,0 +1,155 @@
...
+	cmd.Dir = os.ExpandEnv(repoDir)
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		fmt.Println(string(out))
+		return err
+	}
+	return nil
+}
+
+func (s *SmokeTest) run(numWorkers int) error {
```
dbentley: startCluster? startSwarm?

scripts/common/smoketest.go
```diff
@@ -0,0 +1,155 @@
...
+	for i := 0; i < numWorkers; i++ {
+		port := strconv.Itoa(s.GetPort())
+		if err := s.RunCmd(false, "workerserver", "-thrift_port", port); err != nil {
+			return err
+		}
+	}
+	//Note: assumes ClusterLocalConfig is the default for scheduler.
```
dbentley: Maybe just set it with --sched_config?

dbentley:
LGTM

scripts/common/smoketest.go
jasonbs10: Seems like we can continue using the name 'Run()' if I rename SmokeTest to SwarmTest?

scripts/common/smoketest.go
jasonbs10: Will do.

scripts/common/smoketest.go
jasonbs10: Yeah, SwarmTest is good. I can't think of anything better right now.

PR #75, opened 2016-08-19
snapshots: Checkouter; the first piece of Filer

dbentley:
Filer offers high-level interaction with Snapshots. Checkout a Snapshot into the
local filesystem, or Ingest the local filesystem into a Snapshot. (For now we
only offer Checkout).

Use this abstraction in Runner and extend the Execer interface to take a Dir.
This means that Runner knows about Scoot (it deals with Snapshots), but Execer
doesn't.

Restructure the run code in Runner. It's now more clear who's responsible for
modifying what. This was necessary because checking out may take some time,
and we want to let the user cancel the job while it's happening. This has the
slight negative that the Worker returning to the client no longer means the
command is running. It would be nice if we could reinstate this. In the common
case, we have to wait for a quick checkout, but we may have to perform a fairly
slow git fetch (or a large checkout). ScootFS will eventually solve this
(because it can fault in the data).

Next steps are:

Make a Checkouter that actually checks out (from git, e.g.)

Use this directory in os execer

Allow sim_execer to examine this data (for testing)

### DIFF
--- a/runner/local/simple.go
+++ b/runner/local/simple.go
@@ -132,7 +132,7 @@ func (r *simpleRunner) run(cmd *runner.Command, runId runner.RunId, cancelCh cha
 	case <-checkoutDone:
 	}
 	if err != nil {
-		r.updateStatus(runner.FailedStatus(runId, fmt.Errorf("could not checkout: %v", err)))
+		r.updateStatus(runner.ErrorStatus(runId, fmt.Errorf("could not checkout: %v", err)))
 		return
 	}
 	defer checkout.Release()
@@ -142,7 +142,7 @@ func (r *simpleRunner) run(cmd *runner.Command, runId runner.RunId, cancelCh cha
 		Dir:  checkout.Path(),
 	})
 	if err != nil {
-		r.updateStatus(runner.FailedStatus(runId, fmt.Errorf("could not exec: %v", err)))
+		r.updateStatus(runner.ErrorStatus(runId, fmt.Errorf("could not exec: %v", err)))
 		return
 	}
 
--- a/runner/local/simple.go
+++ b/runner/local/simple.go
@@ -29,8 +29,8 @@ type simpleRunner struct {
 }
 
 type inflight struct {
-	runId    runner.RunId
-	cancelCh chan struct{}
+	runId  runner.RunId
+	doneCh chan struct{}
 }
 
 func (r *simpleRunner) Run(cmd *runner.Command) runner.ProcessStatus {
@@ -44,11 +44,11 @@ func (r *simpleRunner) Run(cmd *runner.Command) runner.ProcessStatus {
 		return r.runs[runId]
 	}
 
-	r.running = &inflight{runId: runId, cancelCh: make(chan struct{})}
+	r.running = &inflight{runId: runId, doneCh: make(chan struct{})}
 	r.runs[runId] = runner.PreparingStatus(runId)
 
 	// Run in a new goroutine
-	go r.run(cmd, runId, r.running.cancelCh)
+	go r.run(cmd, runId, r.running.doneCh)
 	if cmd.Timeout > 0 { // Timeout if applicable
 		time.AfterFunc(cmd.Timeout, func() { r.updateStatus(runner.TimeoutStatus(runId)) })
 	}
@@ -114,14 +114,14 @@ func (r *simpleRunner) updateStatus(new runner.ProcessStatus) runner.ProcessStat
 		// We are ending the running task.
 		// depend on the invariant that there is at most 1 run with !state.IsDone(),
 		// so if we're changing a Process from not Done to Done it must be running
-		close(r.running.cancelCh)
+		close(r.running.doneCh)
 		r.running = nil
 	}
 	return new
 }
 
-// run cmd in the background, writing results to r as id, unless cancelCh is closed
-func (r *simpleRunner) run(cmd *runner.Command, runId runner.RunId, cancelCh chan struct{}) {
+// run cmd in the background, writing results to r as id, unless doneCh is closed
+func (r *simpleRunner) run(cmd *runner.Command, runId runner.RunId, doneCh chan struct{}) {
 	checkout, err, checkoutDone := (snapshots.Checkout)(nil), (error)(nil), make(chan struct{}, 1)
 	go func() {
 		checkout, err = r.checkouter.Checkout(cmd.SnapshotId)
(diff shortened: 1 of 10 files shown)


runner/local/simple.go
```diff
@@ -32,26 +39,33 @@ func (r *simpleRunner) Run(cmd *runner.Command) runner.ProcessStatus {
...
-	r.running[runId] = p
-
-	go babysit(p, runId, r, cmd.Timeout)
+	r.running = &inflight{runId: runId, cancelCh: make(chan struct{})}
+	r.runs[runId] = runner.PreparingStatus(runId)
 
+	// Run in a new goroutine
+	go r.run(cmd, runId, r.running.cancelCh)
+	if cmd.Timeout > 0 { // Timeout if applicable
+		time.AfterFunc(cmd.Timeout, func() { r.updateStatus(runner.TimeoutStatus(runId)) })
```
jasonbs10: These could pile up after a while if Timeout is big, it seems cleaner to add this to inflight{} and pair it with the cancelCh action.

runner/local/simple.go
```diff
@@ -85,52 +99,64 @@ func (r *simpleRunner) Erase(run runner.RunId) {
...
+	}
+	if old.State.IsDone() {
+		return old
 	}
-	result.Abort()
-	delete(r.running, run)
-	r.runs[run] = runner.AbortStatus(run)
-	return r.runs[run]
+	r.runs[new.RunId] = new
+	if new.State.IsDone() && r.running.runId == new.RunId {
```
jasonbs10: This section is confusing. If we only allow one run at a time, when would the old status indicate not done, the new indicate done, and have the runIds mismatched? And it seems weird to close the cancelCh unless we expect something to be listening, i.e. not after COMPLETE.

runner/runner.go
```diff
@@ -32,7 +34,7 @@ const (
 )
 
 func (p ProcessState) IsDone() bool {
-	return p == UNKNOWN || p == COMPLETE || p == FAILED || p == ABORTED || p == TIMEDOUT
+	return p == COMPLETE || p == FAILED || p == ABORTED || p == TIMEDOUT
```
jasonbs10: I thought UNKNOWN should be here since it seems natural to check for errors after we're done. What case are you handling by removing it?

runner/local/simple.go
```diff
@@ -32,26 +39,33 @@ func (r *simpleRunner) Run(cmd *runner.Command) runner.ProcessStatus {
...
-	r.runs[runId] = runner.RunningStatus(runId)
-	r.running[runId] = p
-
-	go babysit(p, runId, r, cmd.Timeout)
+	r.running = &inflight{runId: runId, cancelCh: make(chan struct{})}
+	r.runs[runId] = runner.PreparingStatus(runId)
 
+	// Run in a new goroutine
+	go r.run(cmd, runId, r.running.cancelCh)
+	if cmd.Timeout > 0 { // Timeout if applicable
```
jasonbs10: User-specific timeout makes less sense if we can have a long init step...

jasonbs10:
lgtm.

runner/local/simple.go
dbentley: What calls it after that time?

I suspect each of these takes... <100 bytes, which means even if we're running 10 commands per second (seems high), after 10 days of running we end up with an extra 1GB of RAM storing these...

Maybe we could cap Timeout at 24H or some such?

runner/local/simple.go
dbentley: I removed the test r.running.runId == new.RunId and added comments describing why we skip it.

I think it's easy to think "cancelCh is closed when the Run is set to its final state" and not try to puzzle out who set it and why.

runner/runner.go
dbentley: From the outside, the status of a given Run ID can change from UNKNOWN -> RUNNING.

Done should mean "this Run will never change state".

runner/local/simple.go
dbentley: Why's that?

Because the long init step can eat into the timeout?

Should we have two timeouts, one for preparation and one for run?

Should there be a default timeout for preparation?

In any case, I think this is good to unblock...

runner/local/simple.go
jasonbs10: I'm fine with a cap on timeout. I only made this comment with respect to stats, like if we want to track how often we hit L109: 'if old.State.IsDone() { return old }'. Not really an important metric, but ideally some low number.

dbentley: Why is it good to be low? Because it means we don't have many things happening concurrently?

runner/local/simple.go
jasonbs10: This is a nit regarding naming then, because I wouldn't have given a second thought to closing a 'doneCh'. It doesn't really make sense to rename it though.

runner/runner.go
jasonbs10: Best to add a comment to that effect because I view UNKNOWN as 'I don't care what subsequent status is'.

runner/local/simple.go
dbentley: Oh, doneCh sounds great; is there a reason not to? It seems like that might be better.

runner/local/simple.go
jasonbs10: We should consider having our own internal configurable timeout for prep, yeah.

runner/local/simple.go
jasonbs10: It hints at how often we abort/timeout a the last moment. Again, not a great example.

runner/local/simple.go
jasonbs10: Maybe it's ok. They both apply - ex: "we're completely done" vs "make it so we're done". I'll leave it to your discretion.

jasonbs10:
This is fine to merge, none of these comments were meant to be blocking.

runner/local/simple.go
```diff
@@ -85,52 +99,67 @@ func (r *simpleRunner) Erase(run runner.RunId) {
 	}
 }
 
-func (r *simpleRunner) abortImpl(run runner.RunId, state runner.ProcessState) runner.ProcessStatus {
+func (r *simpleRunner) updateStatus(new runner.ProcessStatus) runner.ProcessStatus {
```
jasonbs10: Oh, and 'new' is a reserved keyword I thought?

dbentley: https://golang.org/ref/spec#Keywords

PR #76, opened 2016-08-19
Added preliminary stats to sched. Added http_port to swarmtest.

jasonbs10:


sched/config/config.go
```diff
@@ -24,7 +25,7 @@ type Config struct {
 }
 
 // Create creates the thrift handler (or returns an error describing why it couldn't)
-func (c *Config) Create() (scoot.CloudScoot, error) {
+func (c *Config) Create(stat stats.StatsReceiver) (scoot.CloudScoot, error) {
```
dbentley: Maybe add Stats to Config and inject it?

workerapi/server/server.go
```diff
@@ -59,14 +59,15 @@ func (h *handler) stats() {
 					numActive++
 				}
 			}
+			lastContact := int64(time.Now().Sub(h.timeLastRpc) / time.Millisecond)
```
dbentley: lastContact sounds like a Time, not a duration; maybe something like timeSinceLastContact or staleness? Loneliness?

dbentley:
LGTM

Generally looks good once everything passes.

I think the DI story could look a little nicer, but I don't think it's worth blocking on resolving that if you're willing to get it done in the background.

jasonbs10:
How does this look?

sched/config/defaults.go
```diff
@@ -34,6 +37,15 @@ func DefaultParser() *Parser {
 			"rpc":   &RPCWorkersConfig{},
 			"":      &LocalWorkersConfig{Type: "local"},
 		},
+		Report: map[string]ReportConfig{
+			"default": &DefaultReportConfig{},
+			"": &DefaultReportConfig{
+				Type:       "default",
+				StatsPath:  "/admin/metrics.json",
```
dbentley: Do we know of cases to change StatsPath and/or HealthPath? If not, I'd just not make them configurable.

binaries/scheduler/main.go
```diff
@@ -21,9 +23,9 @@ func main() {
 	parser.Workers[""] = &config.RPCWorkersConfig{Type: "rpc"}
 	parser.Cluster[""] = &local.ClusterLocalConfig{Type: "local"}
 	parser.Cluster["local"] = &local.ClusterLocalConfig{}
+	parser.Report[""].(*config.DefaultReportConfig).HttpAddr = fmt.Sprintf("localhost:%d", *httpPort)
```
dbentley: This looks pretty good, but does mean that if you want to set a report config, you can't... oy. Lacking real DI is a shame. Whatever you think is best.

binaries/scheduler/main.go
```diff
@@ -21,9 +23,9 @@ func main() {
...
 
 	// Construct scootapi server handler based on config.
-
 	handler, err := parser.Create([]byte(*cfgText))
```
dbentley: Oh, to be fully parallel, if we want to inject this stuff, mabybe have Create return both handlers:
a) the Scheduler
b) the others

Something like:

scheduler, http, err := parser.Create(*cfgText)
thriftServe(scheduler, *thriftPort)
httpServe(http, *httpPort)

Then it would be more parallel?

sched/config/defaults.go
jasonbs10: It's a balance. It's hardcoded for twitter, but the current target addresses seem stilted for OSS.

binaries/scheduler/main.go
jasonbs10: Let's keep this in mind and address it later when we're adding config to the other binaries.

PR #78, opened 2016-08-22
ice: injection & configuration environment

dbentley:
ICE extracts the logic in sched/config.go and makes it general so it can work for all our binaries.

The central object is an Environment, which binds providers. An Environment can then Eval to create objects. DI, tada!

But we want more: we want configuration. So ice also has Options, which lets a program define Variables with multiple Variants. Then Parse takes JSON and creates a Module that can be Installed to the Environment before Eval.

This gives us:
a) injection
b) configuration
c) extensibility (so it works across all binaries)

This works to create the scheduler, but currently doesn't create the associated health checking, which would take extending Eval to create multiple objects in one go, which I can do easily if we think this is a good approach.


saga/inMemorySagaLog.go
```diff
@@ -29,6 +31,14 @@ func MakeInMemorySagaLog() SagaLog {
 	}
 }
 
+type SagaLogMemoryConfig struct {
```
CaitieM20: why does in memory saga log need a config?

dbentley: It doesn't for now, because right now it's the only option.

It *would* need a config *if*:
a) there are multiple options
b) we want to let people switch them at runtime (based on the command-line).

Right now, neither of these is true, so I can remove it.

binaries/scheduler/main.go
```diff
@@ -17,21 +19,22 @@ func main() {
 	log.Println("Starting Cloud Scoot API Server & Scheduler")
 	flag.Parse()
 
-	parser := config.DefaultParser()
-	parser.Workers[""] = &config.RPCWorkersConfig{Type: "rpc"}
-	parser.Cluster[""] = &local.ClusterLocalConfig{Type: "local"}
-	parser.Cluster["local"] = &local.ClusterLocalConfig{}
+	env, o := config.Default()
```
CaitieM20: why is env being determined by the scheduler config?  Shouldn't it be determined by a flag passed in as part of the aurora config? 

Ideally we want different config for local, devel, staging, and prod

binaries/scheduler/main.go
```diff
@@ -17,21 +19,22 @@ func main() {
...
+	var handler scoot.CloudScoot
+	err := ice.ParseAndEval(env, o, *cfgText, &handler)
 	if err != nil {
 		log.Fatal("Error configuring Scoot API: ", err)
 	}
 
 	// Start API Server
 	log.Println("Starting API Server")
 
+	log.Println("Could this possible work?!", handler)
```
CaitieM20: Is this necessary?  Seems like debug text

dbentley: Yes, it is, because this is currently not ready for submission, as I mentioned in the PR description:
It creates the scheduler, but not the health-checker/stats handler, which should be easy to put in if we like the approach.

binaries/scheduler/main.go
dbentley: Here env is different than {devel, staging, prod}. It's probably a bad name because of that collision.

The env here is then populated based on what we parse from the command-line (that, in Aurora, is passed from the config) in the line ParseAndEval.

The point of config.Default() is: multiple binaries can share much of the same environment. This is useful so that closed-source Scoot binaries can use much of the same from open source, but we haven't proven it out.

It gets to how much we value implicit (fewer lines of code; yay!) vs. explicit (easier to understand; yay!). I'm happy with other choices.

CaitieM20:
I don't understand combining Dependency Injection & Config.  They seem totally unrelated. 

I'm also not a huge fan of the config objects (which are part of operations) bleeding into the application logic.  I'd prefer that the necessary values are passed into new/make functions (these may be parsed from config or created via a test case) this automatically supports dependency injection without the need for config files in each object.   

We should be specifying environment as a go flag that is either passed in on the command line or as part of the aurora file.  The environments we want to support are local, devel, staging, and prod.  We should also supply datacenter as a config flag since there will be different config values depending on where the code is deployed.  Ideally then the config code can be modified to parse the correct config file based on the env and datacenter.  On a previous team we named files <env>.<datacenter>.config and could easily construct the config file name based on the two supplied flags.   

Another very valuable aspect of config definition & parsing is that it fails to parse at server startup time if required values are not supplied, or values are not in the correct range.  Basically it fails fast.  I think this code does that, i.e. it parses the code before starting the service but is only as restrictive as the type system is expressive.  A lot of time its worth just handwriting the extra validation code to ensure config parses correctly.  

In previous projects I've had config defined only at the server/binary level.  This makes operations / maintenance easier since all config for the service is in one location and is easy to find instead of hunting and pecking throughout the entire code base.  From an operation perspective its a bit weird to have it in the application code.  

Finally if we want a framework instead of hand writing the parsers & validating why not use an already existing one like [Viper](https://github.com/spf13/viper)

binaries/scheduler/main.go
CaitieM20: When it comes to config I very much value Explicit easy to understand definitions, even if it requires more lines of code.  You want the operational values of a system to be easy to understand.  

Every time I've seen a clever config system it generally becomes fairly unmaintainable.  I did a fairly large rewrite of the Cuckoo one on Observability to make it easy to test and understand.  It meant writing more lines of config and code, however this code is easy to write and test.  Its super simple.  So low cost to create and maintain

dbentley:
I'm happy to change how we do Configuration! In fact, this change tries to change it as little as possible from what's currently in the Repo. I haven't thought that much about configuration, so I'm happy to change it.

I care a lot about getting our DI story right. I think this change is an improvement in how we do DI (at the cost of using reflection, which, ugh). (And, to reiterate, not a change in how we do configuration).

I think that this change separates Integration and Configuration from where we are today, which I like. The Configuration just changes what Modules are installed. So, yeah, we can totally change them.

I think there are lots of different desires for Configuration. We'd like to balance: flexibility, strictness (and failing fast), documentability, etc.

I'd be happy to use Viper for Configuration, or anything else that's easier.

Concretely, my big fear was that sched/config was doing both DI and Configuration and that was dangerous, and we're about to replicate it into worker/config, which would start nearly the same, but then start diverging. Eeek!

My hope is that we can use fewer, more common (and also better!) libraries.

It sounds like we should get together to get on the same page.

binaries/scheduler/main.go
dbentley: Cool, that sounds great to me. I know Jason had some concerns about how to make it easy to iterate on development, so it sounds like you two have the most thoughts about Config.

My focus is that Open Source Scoot is easy to use as a library and compose into a binary with different implementations of interfaces (and without having to repeat a lot of boilerplate)

binaries/scheduler/main.go
```diff
@@ -6,38 +6,97 @@ import (
...
+		scheduler.NewScheduler,
+		server.StartScheduler,
+		server.MakeServer,
+		thrift.NewTTransportFactory,
+		func() thrift.TProtocolFactory {
+			return thrift.NewTBinaryProtocolFactoryDefault()
+		},
+	)
+
+	o := json.Options(map[string]json.Interface{
```
jasonbs10: We decided to make our deps/config explicit in main.go I guess. I still slightly prefer pulling this stuff out so anyone wanting to create their own main.go doesn't have to copy the whole thing and then keep it synced.

nit: something other than the name 'json' would be nice. Seeing json.Interface is weird, as is having config.go be 'package json' in combination with 'import encoding/json'.

binaries/scheduler/main.go
```diff
@@ -6,38 +6,97 @@ import (
...
-	parser.Cluster["local"] = &local.ClusterLocalConfig{}
-	parser.Report[""].(*config.DefaultReportConfig).HttpAddr = fmt.Sprintf("localhost:%d", *httpPort)
+	bag := ice.NewMagicBag()
+	bag.PutMany(
+		func() (thrift.TServerTransport, error) { return thrift.NewTServerSocket(*addr) },
+		endpoints.MakeStatsReceiver,
+		func(s stats.StatsReceiver) *endpoints.ObservabilityServer {
+			return endpoints.NewObservabilityServer(fmt.Sprintf(":%d", *httpPort), s)
+		},
+		makeServers,
```
jasonbs10: We should have a way of trimming unused dependencies from our main.go. It's relatively simple to print those in ice (separately or as part of a graph printout). That said, that list is missing information if a function is named anything other than NewX (like StartScheduler which returns CloudScoot). We should either make it super clear what conventions are expected or pass that information to bag (unless I'm missing some capability of reflect).

nit: why PutMany and Install()? The significance of Put vs Install is unclear.

jasonbs10:
This is looking better, particularly after that meeting. I don't really have more than a couple interface-level suggestions, so let me know when you want the nits.

binaries/scheduler/main.go
dbentley: I like the idea of listing unused deps, though it's hard to figure out what the range of configurations is...

As for methods to install:
1) Put puts a function
2) PutMany puts many functions
3) Install installs a Module, which looks like overhead in the simple cases, but is a nice interface to let clients write modules that make sense. It's just for Inversion of Control that makes it look like normal code.

binaries/scheduler/main.go
dbentley: I hear what you're saying, but then making the defaults is weird. Can we try this explicit, and *then* clean it up to the shared?

I agree config/json is weird. Can you think of what would be better?

I'm mainly trying to kick the Config can down the road so that you and Caitie can figure out what the right long-term path is.

jasonbs10: I'm ok with deferring the config decision but I'm inclined to leave it as-is if adding required json fields requires only a few additional loc. 

config/jsonIce seems better.

binaries/scheduler/main.go
jasonbs10: My preference is to be a little more verbose. I'd have said InstallModule() to make it explicit and to differentiate the inversion fn names.

binaries/scheduler/main.go
dbentley: config/jsonconfig ?

jsonIce seems to link it to ICE, and Caitie convinced me we should view these as separate.

jasonbs10: It is linked to ice, it imports it...

binaries/scheduler/main.go
dbentley: Changed: InstallModule

binaries/scheduler/main.go
dbentley: You're right; this uses ice, but ice does not require it. (and it would be easy to change it work with another system)

I prefer jsonconfig and renamed it to that. Is that acceptable for now? Or can we think of something better?

jasonbs10: I'm fine with it although I still don't follow why ice is bad naming here. Maybe I meant icedJson since the call site expects to install it into ice. Anyway, jsonconfig is acceptable. 






On Thu, Aug 25, 2016 at 9:22 AM -0700, "Dan Bentley" <notifications@github.com> wrote:












In binaries/scheduler/main.go:
> +			return endpoints.NewObservabilityServer(fmt.Sprintf(":%d", *httpPort), s)
> +		},
> +		makeServers,
> +		saga.MakeSagaCoordinator,
> +		saga.MakeInMemorySagaLog,
> +		distributor.NewPoolDistributorFromCluster,
> +		scheduler.NewScheduler,
> +		server.StartScheduler,
> +		server.MakeServer,
> +		thrift.NewTTransportFactory,
> +		func() thrift.TProtocolFactory {
> +			return thrift.NewTBinaryProtocolFactoryDefault()
> +		},
> +	)
> +
> +	o := json.Options(map[string]json.Interface{



You're right; this uses ice, but ice does not require it. (and it would be easy to change it work with another system)



I prefer jsonconfig and renamed it to that. Is that acceptable for now? Or can we think of something better?

binaries/scheduler/main.go
```diff
@@ -6,38 +6,97 @@ import (
...
-		thrift.NewTTransportFactory(),
-		thrift.NewTBinaryProtocolFactoryDefault())
+	var servers servers
+	err = bag.Extract(&servers)
 	if err != nil {
-		log.Fatal("Error serving Scoot API: ", err)
+		log.Fatal("Error injecting servers", err)
 	}
+
+	// TODO(dbentley): if one fails and the other doesn't, we should do
```
CaitieM20: how does one go routine fail? If one panics the whole thing crashes right?

I think I'd prefer for safety to fail fast and hard.  So just crash the whole thing.  Aurora will restart it

common/endpoints/endpoints.go
```diff
@@ -6,38 +6,56 @@ import (
...
 	"log"
 	"net/http"
+	"time"
 
 	"github.com/scootdev/scoot/common/stats"
 )
 
-func RegisterHealthCheck(path string) {
-	http.HandleFunc(path, healthHandler)
+func NewObservabilityServer(addr string, stats stats.StatsReceiver) *ObservabilityServer {
```
CaitieM20: I would just call this ScootServer to avoid name confusion.  This functionality more closely mimics TwitterServer functionality in the JVM. 

Whereas Observability implies something different at Twitter.  Overloaded terms FTW

config/jsonconfig/config.go
```diff
@@ -0,0 +1,109 @@
+/*
```
CaitieM20: So we've been mostly doing comments using // on every line can we keep it consistent.  This also seems to be consistent with godocs best practices

CaitieM20: Also put the comments for a function above the function.  This is also pretty standard will help with Go Docs

config/jsonconfig/config.go
```diff
@@ -0,0 +1,109 @@
...
+	"fmt"
+
+	"github.com/scootdev/scoot/ice"
+)
+
+type Options map[string]Interface
+
+type Interface map[string]Implementation
+
+type Implementation interface {
```
CaitieM20: The name Implementation is very overloaded as well could we change this to something more specific like ConfigObject. 

From what I can gather this more directly corresponds to a section of config thats defining and object to instantiate right?

config/jsonconfig/config.go
```diff
@@ -0,0 +1,109 @@
...
+	// The Implementation needs to do 3 things:
+	// 1) parse the JSON config
+	// 2) add the relevant providers to the ice MagicBag
+	// 3) print its configuration
+	// 1 & 3 are handled implicitly (ugh) by json.(Un)marshal
+	// 2 is handled by being an ice Module
+	ice.Module
+}
+
+/*
```
CaitieM20: make comments consistent please!

config/jsonconfig/config.go
CaitieM20: Oh wait I'm super confused why do we have this interface and Configuration object?

config/jsonconfig/config.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/scootdev/scoot/ice"
+)
+
+type Options map[string]Interface
+
+type Interface map[string]Implementation
```
CaitieM20: please don't use reserved words as object types this is super confusing.  I don't have any idea what this does

config/jsonconfig/config.go
```diff
@@ -0,0 +1,109 @@
...
+	// 3) print its configuration
+	// 1 & 3 are handled implicitly (ugh) by json.(Un)marshal
+	// 2 is handled by being an ice Module
+	ice.Module
+}
+
+/*
+Example JSON of a Configuration:
+{
+ "Storage": {
```
CaitieM20: So my one concern with this config format is it couples the implementation type to the config for that type.
Basically assumes that all the implementations have the same config value.  We can address this later though

binaries/scheduler/main.go
```diff
@@ -6,38 +6,97 @@ import (
...
+		scheduler.NewScheduler,
+		server.StartScheduler,
+		server.MakeServer,
+		thrift.NewTTransportFactory,
+		func() thrift.TProtocolFactory {
+			return thrift.NewTBinaryProtocolFactoryDefault()
+		},
+	)
+
+	o := jsonconfig.Options(map[string]jsonconfig.Interface{
```
CaitieM20: Should this code be in Config? 

We will have multiple main.go's for schedulers etc... so having the code to define what an acceptable scheduler config should be somewhere shared perhaps?

binaries/scheduler/main.go
```diff
@@ -6,38 +6,97 @@ import (
...
+			"memory": &scootconfig.ClusterMemoryConfig{},
+			"local":  &scootconfig.ClusterLocalConfig{},
+			"": &scootconfig.ClusterMemoryConfig{
+				Type:  "memory",
+				Count: 10,
+			},
+		},
+		"Queue": {
+			"memory": &scootconfig.QueueMemoryConfig{},
+			"": &scootconfig.QueueMemoryConfig{
```
CaitieM20: is this supposed to be the default value?  Could we name it default so its more clear?

ice/magic_bag.go
```diff
@@ -0,0 +1,111 @@
+/*
```
CaitieM20: comments should be //

ice/magic_bag.go
```diff
@@ -0,0 +1,111 @@
...
+ice is a lightweight Dependency Injection Framework
+
+ice's cenral metaphor is a "Magic Bag".
+
+It's a Bag because you put things in and then take things out.
+
+It's Magic because what you take out is different (and more useful)
```
CaitieM20: This line is unclear, suggest removing

ice/types_test.go
```diff
@@ -0,0 +1,82 @@
+package ice_test
```
CaitieM20: Why is this called _test.go it doesn't implement any tests?

scootapi/server/serve.go
```diff
@@ -9,20 +9,26 @@ import (
...
 )
 
-func Serve(handler scoot.CloudScoot, addr string, transportFactory thrift.TTransportFactory, protocolFactory thrift.TProtocolFactory) error {
-	transport, err := thrift.NewTServerSocket(addr)
-	if err != nil {
-		return err
-	}
-	processor := scoot.NewCloudScootProcessor(handler)
-	server := thrift.NewTSimpleServer4(processor, transport, transportFactory, protocolFactory)
+func StartScheduler(q queue.Queue, s scheduler.Scheduler, sc saga.SagaCoordinator, stat stats.StatsReceiver) scoot.CloudScoot {
```
CaitieM20: If we want to keep ScootApi and Scheduler separate doesn't this couple them more tightly?

binaries/scheduler/main.go
dbentley: Done.

common/endpoints/endpoints.go
dbentley: Renamed to TwitterServer

config/jsonconfig/config.go
dbentley: I was trying to follow the guidelines for godocs:
https://blog.golang.org/godoc-documenting-go-code

But, rereading it, I realize I missed the convention to put the comment into doc.go; so I've moved it there.

I'm following specifically the example they give, which points me to https://golang.org/src/encoding/gob/doc.go , which uses // for copyright lines but then /* */ for the long comment in doc.go, even though the rest of the comments in the package use //

Does that make sense?

config/jsonconfig/config.go
dbentley: I'm happy to change it to whatever's best.

Options lets you describe multiple Interfaces, each of which has multiple Implementations. E.g., Cluster is an Interface with multiple Implementations.

So you set up the Options, parse it, and get a Configuration (which is a Module you can install)

config/jsonconfig/config.go
dbentley: Happy to; how can I make it more consistent in this case?

config/jsonconfig/config.go
dbentley: Like I said, I'm happy to do whatever's better.

config/jsonconfig/config.go
dbentley: It lets different Implementations for the same Interface have different config values. I test that in config_test.go, where you can see:

	defaultConfig = `{
 "Bar": {
  "Type": "default",
  "Arg3": "3",
  "Arg4": {
   "a": "b"
  }
 },
 "Foo": {
  "Type": "default"
 }
}`
	config1 = `{
 "Bar": {
  "Type": "twoarg",
  "Arg1": 1,
  "Arg2": [
   1,
   2,
   3
  ]
 },
 "Foo": {
  "Type": "noarg"
 }
}`

They take different arguments; yay.

binaries/scheduler/main.go
dbentley: Yeah, it seems we've gone back-and-forth on this, because when we had the defaults it was weird how they were grouped, so for now I'd like to see what it looks like once I pull it into main...

binaries/scheduler/main.go
dbentley: It is the default; I added a comment to config.go describing this behavior.

I think it makes sense to have it be a more-obviously special value ("") instead of "default" happens to mean default.

ice/magic_bag.go
dbentley: Done. (as moved to doc.go)

ice/types_test.go
dbentley: Fixed.

scootapi/server/serve.go
dbentley: Done.

common/endpoints/endpoints.go
```diff
@@ -6,38 +6,55 @@ import (
...
 	"log"
 	"net/http"
+	"time"
 
 	"github.com/scootdev/scoot/common/stats"
 )
 
-func RegisterHealthCheck(path string) {
-	http.HandleFunc(path, healthHandler)
+func NewTwitterServer(addr string, stats stats.StatsReceiver) *TwitterServer {
```
jasonbs10: nit: I'd prefer to not tie this to Twitter and would instead use ReportsServer or similar. Not a big deal though.

jasonbs10:
lgtm

CaitieM20:
lgtm

PR #79, opened 2016-08-22
saga,scootapi/server: remove logging, reduce size

dbentley:
This makes our tests faster and quieter (so they're easier to run and debug)


jasonbs10:
lgtm, although we'll want to replace logs with metrics where it makes sense.

PR #81, opened 2016-08-22
rcouto/scoot 119

ryancouto:
Refactor cloud/cluster memory and packages, removing subscription and diffsubscription. Fetch cron speaks to cluster, the state then updates, then updates are given to subscribers by the cluster.


cloud/cluster/fetch_cron.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+type FetchCron struct {
+	Ticker  *time.Ticker
+	f 	    Fetcher
+	Cl 		*Cluster
```
dbentley: Alignment? Do you have gofmt working in your editor?

cloud/cluster/fetch_cron.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// Returns a full list of visible nodes.
+type Fetcher interface {
+	Fetch() ([]Node, error)
+}
+
+func NewFetchCron(f Fetcher, t time.Duration, cl *Cluster) *FetchCron {
```
dbentley: Maybe just pass the chan instead of the whole cluster.

(FWIW, this is an application of the https://en.wikipedia.org/wiki/Law_of_Demeter )

cloud/cluster/fetch_cron_test.go
```diff
@@ -0,0 +1,81 @@
...
+	h.assertFetch(t, h.setupTest("host1:1234"))
+	h.assertFetch(t, h.setupTest("host1:1234", "host2:8888"))
+	h.assertFetch(t, h.setupTest("host1:1234", "host1:1234"))
+	h.assertFetch(t, h.setupTest(""))
+}
+
+func (h *cronHelper) setupTest(nodeNames ...string) []cluster.Node {
+	nodes := nodes(nodeNames)
+	h.f.setResult(nodes)
+	time.Sleep(h.time * 2)
```
dbentley: Why a sleep?

cloud/cluster/fetch_cron_test.go
```diff
@@ -0,0 +1,81 @@
+package cluster_test
+
+import (
+	"testing"
+	"github.com/scootdev/scoot/cloud/cluster"
+	"time"
+	"reflect"
+	"sync"
+	"sort"
+	// "fmt"
```
dbentley: Remove

cloud/cluster/fetch_cron_test.go
```diff
@@ -0,0 +1,81 @@
...
+	c     *cluster.FetchCron
+	cl 	  *cluster.Cluster
+	f  	  *fakeFetcher
+	time  time.Duration
+}
+
+func makeCronHelper(t *testing.T) *cronHelper {
+	h := &cronHelper{t: t}
+	h.cl = cluster.NewCluster([]cluster.Node{}, make(chan []cluster.NodeUpdate), make(chan []cluster.Node))
+	h.time = time.Second
```
dbentley: Eeek! Way too long! We'd like tests to finish in less than a second.

cloud/cluster/fetch_cron_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"time"
+	"reflect"
+	"sync"
+	"sort"
+	// "fmt"
+)
+
+func TestFetchCron(t *testing.T) {
+	h := makeCronHelper(t)
+	h.assertFetch(t, h.setupTest("host1:1234"))
```
dbentley: I think this test might feel easier if you just wait for the expected states on the channel, not try to wait for it to wind its way through cluster, e.g.

cloud/cluster/node.go
```diff
@@ -1,32 +1,53 @@
 package cluster
 
+import (
+	"fmt"
+)
+
 type NodeId string
 
 type Node interface {
```
dbentley: Y'know, I'm not sure yet, but I'm starting to wonder what other implementations of Node there will be.... (Just an idle thought)

cloud/cluster/state.go
```diff
@@ -0,0 +1,68 @@
...
+	s.Nodes = make(map[NodeId]Node)
+	for _, n := range newState {
+		s.Nodes[n.Id()] = n
+	}
+	// present updates in predictable order
+	sort.Sort(NodeUpdateSorter(outgoing))
+	return outgoing
+}
+
+func (s *State) Update(newUpdates []NodeUpdate) {
```
dbentley: Should this handle the case that a node was added that was already present?

cloud/cluster/subscriber.go
```diff
@@ -0,0 +1,46 @@
+package cluster
+
+type Subscriber struct {
```
dbentley: Here's a great place to comment to describe what Subscriber does

cloud/cluster/fetch_cron.go
ryancouto: didn't know about gofmt, that's very cool! seems to be working now.

cloud/cluster/fetch_cron_test.go
ryancouto: without a sleep the fetched state didn't have enough time to propagate to the state and back to the subscribers

cloud/cluster/fetch_cron.go
dbentley: We've configured pants to error on code that isn't gofmt'ed. (See also go vet, which complained about your first push)

cloud/cluster/subscriber.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (s *Subscriber) Close() error {
+	s.cl.closeSubscription(s)
+	return nil
+}
+
+func (s *Subscriber) loop() {
+	for s.inCh != nil || len(s.queue) > 0 {
+		var outCh chan []NodeUpdate
+		var outgoing []NodeUpdate
```
jasonbs10: Do you still need 'outgoing' if you send s.queue all at once?

cloud/cluster/state_test.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+func TestState(t *testing.T) {
+	s := cluster.MakeState([]cluster.Node{})
+	// no nodes removed or added
+	assertUpdates(t, s, []string{}, []cluster.NodeUpdate{})
+	// 1 node added
+	assertUpdates(t, s, []string{"host1:1234"}, []cluster.NodeUpdate{cluster.NewAdd(cluster.NewIdNode("host1:1234"))})
+	// 1 node removed
+	assertUpdates(t, s, []string{}, []cluster.NodeUpdate{cluster.NewRemove(cluster.NewIdNode("host1:1234").Id())})
```
jasonbs10: Instead of making a new node, just do NodeId("host1:1234")?

cloud/cluster/node_update.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func NewRemove(id NodeId) NodeUpdate {
+	return NodeUpdate{
+		UpdateType: NodeRemoved,
+		Id:         id,
+	}
+}
+
+type NodeUpdateSorter []NodeUpdate
```
jasonbs10: Do we need to sort NodeUpdates?

cloud/cluster/local/config.go
```diff
@@ -10,6 +10,7 @@ type ClusterLocalConfig struct {
 }
 
 func (c *ClusterLocalConfig) Create() (cluster.Cluster, error) {
-	sub := Subscribe()
-	return memory.NewCluster(sub.InitialMembers, sub.Updates), nil
+	sub := cluster.Subscribe(MakeFetcher())
+	stateCh := make(chan []cluster.Node)
```
jasonbs10: Does stateCh get used?

cloud/cluster/state.go
```diff
@@ -0,0 +1,85 @@
...
+			Id:         n.Id(),
+		})
+	}
+	// reset nodes map, assign to new state
+	s.Nodes = make(map[NodeId]Node)
+	for _, n := range newState {
+		s.Nodes[n.Id()] = n
+	}
+	// present updates in predictable order
+	sort.Sort(NodeUpdateSorter(outgoing))
```
jasonbs10: I'm not sure we want to sort node updates...

cloud/cluster/state.go
```diff
@@ -0,0 +1,85 @@
...
+	return outgoing
+}
+
+func (s *State) FilterAndUpdate(newUpdates []NodeUpdate) []NodeUpdate {
```
jasonbs10: Can you add a summary comment for these update functions?

cloud/cluster/fetch_cron_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (h *cronHelper) assertFetch(t *testing.T, expected []cluster.Node) {
+	actual := <-h.c.Ch
+	if !reflect.DeepEqual(expected, actual) {
+		t.Fatalf("Unequal, expected %v, received %v", expected, actual)
+	}
+}
+
+func nodes(ids []string) []cluster.Node {
+	nodeNames := make(map[string]bool)
```
jasonbs10: nodeNames seems unnecessary here, it seems like you can just range over ids.

cloud/cluster/fetch_cron.go
```diff
@@ -0,0 +1,38 @@
+package cluster
+
+import (
+	"fmt"
+	"time"
+)
+
+type FetchCron struct {
```
jasonbs10: This file seems kind of empty. Do we really need FetchCron for this small amount of work (also I only see it used in its test file)?

cloud/cluster/cluster.go
```diff
@@ -1,51 +1,121 @@
...
+	State    *State
+	reqCh    chan interface{}
+	updateCh chan []NodeUpdate
+	StateCh  chan []Node
+	subs     []chan []NodeUpdate
 }
 
-func (u *NodeUpdate) String() string {
-	return fmt.Sprintf("%v %v %v", u.UpdateType, u.Id, u.Node)
+func NewCluster(state []Node, updateCh chan []NodeUpdate, stateCh chan []Node) *Cluster {
```
jasonbs10: I'm assuming we only need one of updateCh or stateCh, if so there should be a comment about that.

cloud/cluster/cluster.go
```diff
@@ -1,51 +1,121 @@
...
+
+func (c *Cluster) closeSubscription(s *Subscriber) {
+	c.reqCh <- s.inCh
 }
 
-func NewRemove(id NodeId) NodeUpdate {
-	return NodeUpdate{
-		UpdateType: NodeRemoved,
-		Id:         id,
+func (c *Cluster) Current() []Node {
```
jasonbs10: Should this be private?

cloud/cluster/cluster.go
```diff
@@ -1,51 +1,121 @@
...
-	Updates        chan []NodeUpdate // Updates as they happen
-	Closer         io.Closer         // How to stop subscribing
+func (c *Cluster) Subscribe() Subscriber {
+	ch := make(chan Subscriber)
+	c.reqCh <- ch
+	return <-ch
 }
 
-// Helper functions to create NodeUpdates
+func (c *Cluster) Close() error {
```
jasonbs10: We might want to close all all the subs and clear c.subs, otherwise we're left with subs that can unsubscribe. Just a thought, since we don't have a clear shutdown story afaik.

cloud/cluster/subscriber.go
ryancouto: nope - good catch

cloud/cluster/state_test.go
ryancouto: will do

cloud/cluster/node_update.go
ryancouto: The cluster test and state test both use node update sort to compare expected vs actual. How do you feel about sorting updates only in the test?

cloud/cluster/fetch_cron_test.go
ryancouto: I put this in to deal with the possibility of duplicate nodes being passed - but it isn't really testing the functionality of the fetch cron at all. I'll take this out.

cloud/cluster/cluster.go
ryancouto: not sure what the intended functionality of this one is... if we don't want individual subs to be able to unsubscribe i'll clear subs here.

jasonbs10: Typo: 'left with subs that *can't* unsubscribe'. If they can't request unsubscribe over the reqCh that seems bad.

ryancouto: gotcha - yeah i agree

cloud/cluster/cluster.go
ryancouto: i was thinking that it would be nice to split up the responsibilities of the channels to be focused on either nodes or node updates, but it does simplify it a bit to only have one channel for both.

jasonbs10: I mean, you can keep both, but I shouldn't have to read the impl to know that I can use nil for one of the channels. You could also make NewCluster private and replace it with one constructor for each update type (that's not necessarily better though).

cloud/cluster/cluster.go
```diff
@@ -1,51 +1,115 @@
...
+type Cluster struct {
+	State *State
+	reqCh chan interface{}
+	ch    chan interface{}
+	subs  []chan []NodeUpdate
 }
 
-func (u *NodeUpdate) String() string {
-	return fmt.Sprintf("%v %v %v", u.UpdateType, u.Id, u.Node)
+func NewCluster(state []Node, ch chan interface{}) *Cluster {
```
jasonbs10: Comment indicating the types we accept on the channel.

cloud/cluster/fetch_cron.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (c *FetchCron) loop() {
+	defer close(c.closer)
+	for {
+		select {
+		case <-c.ticker.C:
+			nodes, err := c.f.Fetch()
+			if err != nil {
+				// Log? Return?
```
jasonbs10: Add a "TODO(rcouto)". One of our goals is to handle as many of our errors as possible. There are a few approaches we could take...

cloud/cluster/local/config.go
```diff
@@ -2,14 +2,12 @@ package local
...
 
 type ClusterLocalConfig struct {
 	Type string
 }
 
-func (c *ClusterLocalConfig) Create() (cluster.Cluster, error) {
-	sub := Subscribe()
-	return memory.NewCluster(sub.InitialMembers, sub.Updates), nil
+func (c *ClusterLocalConfig) Create() (*cluster.Cluster, error) {
+	return cluster.NewCluster([]cluster.Node{}, make(chan interface{})), nil
```
jasonbs10: I'm still not sure about spots like this where you pass an arg (chan) that doesn't seem to get used.

cloud/cluster/node_update.go
jasonbs10: I believe []NodeUpdate should accurately represent a stream of updates. Can you rework the tests to handle default ordering? It's deterministic so it should be doable. Is there something else going on here that you're trying to address?

cloud/cluster/state.go
```diff
@@ -0,0 +1,102 @@
...
+			UpdateType: NodeRemoved,
+			Id:         n.Id(),
+		})
+	}
+	// reset nodes map, assign to new state
+	s.Nodes = make(map[NodeId]Node)
+	for _, n := range newState {
+		s.Nodes[n.Id()] = n
+	}
+	// present updates in predictable order
```
jasonbs10: Delete?

jasonbs10:
Aside from my minor nits, you should merge master into your branch and make sure that 'make test-integration' runs and passes (this will happen automatically on travis if you push the branch after merging.)

lgtm otherwise.

cloud/cluster/node_update.go
ryancouto: If a state receives a new state []Node with multiple changes, i.e. one node is missing and there's a new node included, we can create a diff against the old state, represented by a two element []NodeUpdate, but there's no guarantee it's correctly ordered. If we can guarantee we're only ever adding or removing one node at a time this isn't a problem but I don't know if we can make that guarantee.

jasonbs10: Doesn't ordering come for free since []Node is already ordered?

ryancouto: good point - i'll make that change

jasonbs10:
Actually, our integration test is flaky on travis so that's been removed for now - you'll have to run it locally.

cloud/cluster/cluster_test.go
```diff
@@ -53,20 +53,35 @@ func TestSubscribe(t *testing.T) {
...
 // Below here are helpers that make it easy to write more fluent tests.
 
 type helper struct {
-	t  *testing.T
-	c  cluster.Cluster
-	ch chan []cluster.NodeUpdate
+	t        *testing.T
+	c        *cluster.Cluster
+	updateCh chan []cluster.NodeUpdate
+	f  		 cluster.Fetcher
```
dbentley: This looks weird; does all this code gofmt cleanly?

cloud/cluster/fetch_cron.go
```diff
@@ -0,0 +1,87 @@
...
+		closer:   make(chan struct{}),
+		cl:       cl,
+	}
+	go c.loop()
+	return c
+}
+
+func (c *FetchCron) loop() {
+	c.inCh = make(chan nodesAndError)
+	go func() {
```
dbentley: This seems pretty complex.... If Cluster's ignoring us, is it really important to fetch?

Why was the simple thing insufficient?

cloud/cluster/fetch_cron.go
```diff
@@ -0,0 +1,87 @@
...
+	"time"
+)
+
+type FetchCron struct {
+	tickCh   <-chan time.Time
+	f        Fetcher
+	inCh     chan nodesAndError
+	outCh    chan []NodeUpdate
+	outgoing []NodeUpdate
+	closer   chan struct{}
```
dbentley: where is this used?

cloud/cluster/fetch_cron.go
```diff
@@ -0,0 +1,87 @@
...
+
+func NewFetchCron(f Fetcher, t time.Duration, cl *Cluster) *FetchCron {
+	c := &FetchCron{
+		tickCh:   time.NewTicker(t).C,
+		f:        f,
+		inCh:     nil,
+		outCh:    make(chan []NodeUpdate),
+		outgoing: nil,
+		closer:   make(chan struct{}),
+		cl:       cl,
```
dbentley: I thought we were getting rid of this?

cloud/cluster/subscription.go
```diff
@@ -1,153 +1,79 @@
...
 import (
-	"sort"
+	"io"
 	"time"
 )
 
-// Returns a full list of visible nodes.
-type Fetcher interface {
-	Fetch() ([]Node, error)
+// Subscriber receives node updates from its cluster and maintains a queue
```
dbentley: I think this comment is now misplaced.

cloud/cluster/fetch_cron.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (c *fetchCron) handleFetch(nodes []Node, err error) {
+	if err != nil {
+		// TODO(rcouto): Correctly handle as many errors as possible
+		return
+	}
+	c.outCh <- nodes
+}
+
+func (c *fetchCron) close() {
```
dbentley: not exported so people can't actually use it

PR #82, opened 2016-08-24
Swarmtest refactor

CaitieM20:
* refactoring swarmtest so it can run as an integration test.  It currently is very flaky as a unit test as part of travisCI filed a bug to investigate.  
* added a mutex to swarmtest to avoid race conditions, also had to remove cmd.Wait()  due to complicated race conditions.  Not sure how to have a cmd.Wait() in one go routine followed by a cmd.Kill() in another one and not trigger the race detector without doing a big refactor...
* added swarmtest binary to binaries folders so it can be ran locally as well
* moved job & task generators to tests folder
* added numJobs and timeout to run_smoketest_cmd


tests/swarmtest/swarmtest.go
```diff
@@ -20,20 +22,27 @@ type SwarmTest struct {
 	NumWorkers int
 	Compile    func() error
 	Run        func() error
+	NumJobs    int
+	Timeout    time.Duration
+	mutex      *sync.Mutex
```
jasonbs10: What happens without the mutex? I never saw anything bad come of it since we always exit right after killing.

scootapi/client/smoke_test_cmd.go
```diff
@@ -40,10 +63,16 @@ func (c *Client) runSmokeTest(cmd *cobra.Command, args []string) error {
 
 	wg.Wait()
 
-	return nil
+	// if any errors were looged return an error
```
jasonbs10: typo.

jasonbs10:
lgtm

scootapi/client/smoke_test_cmd.go
CaitieM20: fixed

tests/swarmtest/swarmtest.go
CaitieM20: Running go test -race ./... returns an error because there is access from multiple go routines to the cmd objects.  

I assume this is safe because its processes, but to make our test suite happy I added the mutex

PR #84, opened 2016-08-24
Add SwarmTest Instructions to Readme

CaitieM20:


Readme.md
```diff
@@ -1,3 +1,9 @@
 Scoot placeholder.
```
jasonbs10: Let's get rid of this finally? The top paragraph on our wiki should be a fine substitute:
"Scoot is infrastructure to make developer tools smaller, simpler, and more distributed. It's the common bedrock for Developer Tools so each tool doesn't have to become a silo, with its own buggy, ad hoc distributed infrastructure."

jasonbs10:
lgtm

PR #86, opened 2016-08-26
Add Async Library

CaitieM20:
Add an Async Library to Scoot which will facilitate receiving return values (errors only for now) from go functions.  For an example of how this should be used see the godoc comments in mailbox.go.  AsyncError is similar to a Promise that contains an error value.  

This helps us run an event loop in the scheduler with go routines and then executing callbacks when the go routine completes (successfully or unsuccessfully).  

The goal is to use this in the scheduler refactor, which will have an event loop.  Every loop it will do the following as prototyped in [dbentley/actors](https://github.com/scootdev/scoot/pull/80).  This is similair to the RPC abstract Dan provides in actions.  But I prefer this because
* the functions we are invoking are not actually remote procedure calls, they are just go routines which may perform an off box call or not
* mimics promises & callbacks from languages like node.js & scala.  
* separate lib from scheduler its reusable
* keep go routine and callback closer together, for easier readability.  The RPC Solution requires the definition of message types and processing of these messages in a big switch statement.  So the RPC and the Handling of the RPC are in different places in the code.   This will make updates more brittle and less modular.
* actions are planed and messages sent at the same time.  This is different than the other pull request where the loop is (receive messages & update state, plan actions, execute rpc).  The last two are combined in this model.  However the at the end of the loop you can easily check what messages have been added so no information is lost to verify that the correct actions were taken.  
* less error prone, the user only needs to remember to create a mailbox & call the processupdates() method once per loop.  Then any created AsyncError is automatically tracked and the callback invoked, as opposed to the RPC which required the user to specify separate message types and then handle th

### DIFF
--- a/async/Mailbox.go
+++ b/async/Mailbox.go
@@ -61,8 +61,8 @@ package async
 // are always executed within the same context and only one at at ime.
 // A Mailbox for keeping track of in progress AsyncMessages.
 // This structure is not thread-safe.
-type AsyncMailbox struct {
-	msgs []asyncMessage
+type Mailbox struct {
+	msgs []message
 }
 
 // The function type of the callback invoked when an AsyncError is Completed
@@ -70,38 +70,38 @@ type AsyncErrorResponseHandler func(error)
 
 // async message is a struct composed of an AsyncError
 // and its associated callback
-type asyncMessage struct {
+type message struct {
 	Err      *AsyncError
 	callback AsyncErrorResponseHandler
 }
 
-func newAsyncMessage(cb AsyncErrorResponseHandler) asyncMessage {
-	return asyncMessage{
+func newMessage(cb AsyncErrorResponseHandler) message {
+	return message{
 		Err:      newAsyncError(),
 		callback: cb,
 	}
 }
 
-func NewAsyncMailbox() *AsyncMailbox {
-	return &AsyncMailbox{
-		msgs: make([]asyncMessage, 0),
+func NewMailbox() *Mailbox {
+	return &Mailbox{
+		msgs: make([]message, 0),
 	}
 }
 
 // Creates a NewAsyncError and assocaites the supplied callback with it.
 // Once the AsyncError has been completed, SetValue called, the callback
 // will be invoked on the next execution of ProcessMessages
-func (bx *AsyncMailbox) NewAsyncError(cb AsyncErrorResponseHandler) *AsyncError {
-	msg := newAsyncMessage(cb)
+func (bx *Mailbox) NewAsyncError(cb AsyncErrorResponseHandler) *AsyncError {
+	msg := newMessage(cb)
 	bx.msgs = append(bx.msgs, msg)
 	return msg.Err
 }
 
 // Processes the mailbox.  For all messages with completed AsyncErrors
 // the callback function and removes the message from the mailbox
-func (bx *AsyncMailbox) ProcessMessages() {
+func (bx *Mailbox) ProcessMessages() {
 
-	var unCompletedMsgs []asyncMessage
+	var unCompletedMsgs []message
 	for _, msg := range bx.msgs {
 
 		ok, err := msg.Err.TryGetValue()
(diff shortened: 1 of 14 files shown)


async/asyncError.go
```diff
@@ -0,0 +1,41 @@
...
+//
+// The returned bool is true if Completed, false if Pending
+// If Completed the returned error is the Value of this AsyncError
+// If AsyncError is not completed the returned error is nil.
+//
+// Calling this method after true has been returned will panic
+func (p *AsyncError) TryGetValue() (bool, error) {
+	select {
+	case err := <-p.errCh:
+		close(p.errCh)
```
dbentley: I'd move the close to SetValue so that it's the writer closing instead of the reader.

async/asyncError.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// Sets the value for the AsyncError.  Marks AsyncError as Completed or Fulfilled.
+// This method should only ever be called once per AsyncError instance.
+// Calling it more than once it will block indefinitely
+func (p *AsyncError) SetValue(err error) {
+	p.errCh <- err
+}
+
+// Returns the Status of this AsyncError, Completed or Pending
```
dbentley: Maybe make an enum?

async/mailbox.go
```diff
@@ -0,0 +1,119 @@
...
+//    mailbox := NewAsyncMailbox()
+//
+//    writeCallback := func (err error) {
+//      if err != nil {
+//        successfulWrites++
+//      }
+//      returnedWrites++
+//    }
+//
+//    // Send to Replica One
```
dbentley: This means the async goroutine looks less like a function (instead of returning an error, it has to SetValue).

WWYT of mailbox.RunAsync(func() error) that does the wrapping? So it takes a function that looks normal, and then adds the async composition?

CaitieM20: Added AsyncRunner class which implements what we talked about this morning

async/asyncError.go
CaitieM20: eh I think it adds more complexity than its worth since there really are only two states.  Also TryGet() is a common api in C# on concurrent collections where a boolean specifies if it exists.  Also this is inline with golang where maps return bool, value

async/asyncError.go
CaitieM20: oh cool didn't realize you could drain a channel once it had been closed! Makes way more sense!

dbentley:
LGTM

async/Mailbox.go
```diff
@@ -0,0 +1,119 @@
...
+// will be invoked on the next execution of ProcessMessages
+func (bx *Mailbox) NewAsyncError(cb AsyncErrorResponseHandler) *AsyncError {
+	msg := newMessage(cb)
+	bx.msgs = append(bx.msgs, msg)
+	return msg.Err
+}
+
+// Processes the mailbox.  For all messages with completed AsyncErrors
+// the callback function and removes the message from the mailbox
+func (bx *Mailbox) ProcessMessages() {
```
jasonbs10: It's still unclear to me if scoot will need multiple mailboxes or if not, how sibling components can interact with a single mailbox. It seems like we'd need locking for ProcessMessages() and NewAsyncError() so we can pass mailbox around to different funcs and goroutines.

I'm still not a big fan of AsyncError even though I completely see why it makes sense now. Maybe ErrorPromise since that emphasizes that we need to fulfill that promise and await on it, whereas Async relates more to the concept of concurrency and goroutines.

jasonbs10:
lgtm to merge and start writing code against this so we can see how well the api fits our needs.

async/Mailbox.go
CaitieM20: So the Scheduler will only need one mailbox, I have an initial changes made, that I used to refine this abstraction, that I will clean up and send out as the refactor today.

If AsyncError gets super confusing let me know we can change it, given my background its seems super natural.  It was used to represent promises in Orleans initially, and the async/await terminology from C#.  To me AsyncError means an Error that's given a value asynchronously.

PR #87, opened 2016-08-29
Rcouto/smoketest ui

ryancouto:
Print grouped job statuses intermittently while smoke test is running.


scootapi/client/smoke_test_cmd.go
```diff
@@ -61,6 +61,29 @@ func (c *Client) runSmokeTest(cmd *cobra.Command, args []string) error {
 		}()
 	}
 
+	// current status of every job
+	jobStatusMap := make(map[string]scoot.Status)
+	// jobs grouped together by status
+	statusJobMap := make(map[scoot.Status][]string)
+	ch <- jobStatusMap
+	// print current status of all jobs
+	go func() {
```
dbentley: It would be nice to end this for real; is C closed anywhere?

scootapi/client/smoke_test_cmd.go
```diff
@@ -113,8 +136,11 @@ func (c *Client) generateAndRunJob(timeout time.Duration) error {
 				return fmt.Errorf("Error getting status: %v", err.Error())
 			}
 		}
-
-		fmt.Println("Status Update Job", jobId.ID, status.Status)
+		// get current version of jobStatusMap
+		jobStatusMap := <-ch
```
dbentley: So you're handing it back and forth a whole bunch. It looks like this works; it might work even better to just send the name and status back to one goroutine that both owns the maps and prints the status.

ryancouto:
View of job statuses in terminal

scootapi/client/smoke_test_cmd.go
```diff
@@ -61,6 +61,41 @@ func (c *Client) runSmokeTest(cmd *cobra.Command, args []string) error {
 		}()
 	}
 
+	// current status of every job
+	jobStatusMap := make(map[string]scoot.Status)
```
jasonbs10: I don't think jobStatusMap is threadsafe, it's being accessed from two different goroutines.  And it seems like we don't need to track anything in this scope, we can have: a goroutine which keeps this map, receives updates from generateAndRunJob() and  ranges over ticker to print updates. And I'd create a type JobAndStatus to send over the ch rather than a map with a single entry.

scootapi/client/smoke_test_cmd.go
```diff
@@ -61,6 +61,41 @@ func (c *Client) runSmokeTest(cmd *cobra.Command, args []string) error {
...
+	statusJobMap := make(map[scoot.Status][]string)
+Loop:
+	for range ticker.C {
+		for job, status := range jobStatusMap {
+			// populate statusJobMap
+			statusJobMap[status] = append(statusJobMap[status], job)
+		}
+		for status, jobs := range statusJobMap {
+			sort.Sort(sort.StringSlice(jobs))
+			fmt.Println(status, ":", jobs)
```
jasonbs10: Can there be a category for timed out, and can you prefix the list of jobs with the length of that list?

ryancouto: Should a timed out job have a status of Status_ROLLED_BACK? Or should it have an explicit timed out status?

ryancouto: Nvm - going to store timeouts in a slice of errors and just print its length.

ryancouto:
with # of timeouts

CaitieM20:
LGTM!  Much nicer output :)

PR #88, opened 2016-08-29
snapshots/fake: initingCheckouter waits for Init

dbentley:


snapshots/fake/fake_filer_test.go
```diff
@@ -0,0 +1,41 @@
...
+	}()
+
+	select {
+	case <-resultCh:
+		t.Fatalf("should't be able to checkout until init is done")
+	default:
+	}
+
+	i.wg.Done()
+	checkout := <-resultCh
```
jasonbs10: Might as well do a select{} here too so the test can't block.

jasonbs10:
lgtm

PR #89, opened 2016-08-30
cloud/cluster: simplify construction

dbentley:
I found that I couldn't easily change the duration on the fetch cron because it was pulled in.

I made a few changes to make things more injected instead of constructed at the same time.

Got rid of all functions Subcribe(), which don't make sense outside of the context of a Cluster.

### DIFF
--- a/cloud/cluster/cluster.go
+++ b/cloud/cluster/cluster.go
@@ -2,30 +2,31 @@ package cluster
 
 import (
 	"sort"
-	"time"
 )
 
 // Cluster represents a cluster of Nodes.
 type Cluster struct {
 	state    *state
 	reqCh    chan interface{}
-	ch       chan interface{}
-	updateCh chan []NodeUpdate
+	updateCh chan ClusterUpdate
 	subs     []chan []NodeUpdate
 }
 
+// Clusters can be updated in two ways:
+// *) a new state of the Cluster, which is a []Node
+// *) updates to specific Nodes, which is a []NodeUpdate
+type ClusterUpdate interface{}
+
 // Cluster's ch channel accepts []Node and []NodeUpdate types, which then
 // get passed to its state to either SetAndDiff or UpdateAndFilter
-func NewCluster(state []Node, updateCh chan []NodeUpdate, ch chan interface{}, fetcher Fetcher) *Cluster {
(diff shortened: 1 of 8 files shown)


cloud/cluster/cluster.go
```diff
@@ -2,30 +2,31 @@ package cluster
...
-	ch       chan interface{}
-	updateCh chan []NodeUpdate
+	updateCh chan ClusterUpdate
 	subs     []chan []NodeUpdate
 }
 
+// Clusters can be updated in two ways:
+// *) a new state of the Cluster, which is a []Node
+// *) updates to specific Nodes, which is a []NodeUpdate
+type ClusterUpdate interface{}
```
jasonbs10: Maybe make clusterUpdate private and add a couple constructor methods? That or enforce types by adding a noop func NodeOrUpdate() to this interface for []Node and []NodeUpdate to implement, but I think we've removed similar code in the past.

jasonbs10:
lgtm (tmp, testing if undo lgtm works)

cloud/cluster/cluster.go
dbentley: I think ClusterUpdate has to be private so that others can create it.

I'm happy to add func IsClusterUpdate(), but the downside is then you can't pass []Node, but Nodes([]Node{foo, bar})

If we think that's too ugly, I'm also alright with having one channel of []NodeUpdate and one of []Node, but having one of []Node and one of interface{} was weird.

jasonbs10: You mean has to be public? I'm not sure why that's the case if we have FromNodes() and FromNodeUpdates(). I'm ok with IsClusterUpdate but I'd rename it to ImplementClusterUpdate.

dbentley: Err, yes, public.

so that we can define chan cluster.ClusterUpdate in other packages...

jasonbs10: Hmm, I'd say other packages can use a one of Nodes or NodeUpdates and don't need to know about ClusterUpdate.

dbentley: chan Nodes is not the same type as chan cluster.ClusterUpdate

So just making the Cron wouldn't be possible because it returns a chan cluster.ClusterUpdate

ryancouto:
lgtm

PR #90, opened 2016-08-30
runner: Saver lets clients Save output 

dbentley:
Runner uses Saver to save stdout and stderr.

localSaver saves to a local file (in tempdir)

Future impls of Saver could upload to hdfs, s3, etc.

Saver is temporary until we have the ability to save all output (stdout/stderr but also created files) into a new Snapshot.

### DIFF
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -1,8 +1,8 @@
 package os
 
 import (
-	"bytes"
 	"errors"
+	"io"
 	"os/exec"
 	"syscall"
 
@@ -15,13 +15,23 @@ func NewExecer() execer.Execer {
 
 type osExecer struct{}
 
+type WriterDelegater interface {
+	WriterDelegate() io.Writer
+}
+
 func (e *osExecer) Exec(command execer.Command) (result execer.Process, err error) {
 	if len(command.Argv) == 0 {
 		return nil, errors.New("No command specified.")
 	}
 	cmd := exec.Command(command.Argv[0], command.Argv[1:]...)
-	var stdoutBuf, stderrBuf bytes.Buffer
-	cmd.Stdout, cmd.Stderr = &stdoutBuf, &stderrBuf
+	cmd.Stdout, cmd.Stderr = command.Stdout, command.Stderr
+	// Make sure to get the best possible Writer, so that
+	if stdoutW, ok := cmd.Stdout.(WriterDelegater); ok {
+		cmd.Stdout = stdoutW.WriterDelegate()
+	}
+	if stderrW, ok := cmd.Stderr.(WriterDelegater); ok {
+		cmd.Stderr = stderrW.WriterDelegate()
+	}
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 	err = cmd.Start()
 	if err != nil {
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -16,6 +16,11 @@ func NewExecer() execer.Execer {
 type osExecer struct{}
 
 type WriterDelegater interface {
+	// Return an underlying Writer. Why? Because some methods type assert to
+	// a more specific type and are more clever (e.g., if it's an *os.File, hook it up
+	// directly to a new process's stdout/stderr.)
+	// We care about this cleverness, so Output both is-a and has-a Writer
+	// Cf. runner/local/output.go
 	WriterDelegate() io.Writer
 }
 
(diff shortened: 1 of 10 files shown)


runner/execer/os/os_execer.go
```diff
@@ -15,13 +15,23 @@ func NewExecer() execer.Execer {
 
 type osExecer struct{}
 
+type WriterDelegater interface {
```
CaitieM20: Comments on what this interface is expected to do please!

runner/execer/os/os_execer.go
```diff
@@ -15,13 +15,23 @@ func NewExecer() execer.Execer {
...
+
 func (e *osExecer) Exec(command execer.Command) (result execer.Process, err error) {
 	if len(command.Argv) == 0 {
 		return nil, errors.New("No command specified.")
 	}
 	cmd := exec.Command(command.Argv[0], command.Argv[1:]...)
-	var stdoutBuf, stderrBuf bytes.Buffer
-	cmd.Stdout, cmd.Stderr = &stdoutBuf, &stderrBuf
+	cmd.Stdout, cmd.Stderr = command.Stdout, command.Stderr
+	// Make sure to get the best possible Writer, so that
```
CaitieM20: is this comment necessary?  If so can you complete the sentence

runner/local/output.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/scootdev/scoot/runner"
+	osexecer "github.com/scootdev/scoot/runner/execer/os"
+)
+
+type localSaver struct {
+	hostname string
+}
+
+func NewSaver() (runner.Saver, error) {
```
CaitieM20: Comments on Public APIs are much appreciated

runner/output.go
```diff
@@ -0,0 +1,24 @@
+package runner
+
+import (
+	"io"
+)
+
+// Saver lets clients create new Outputs so they can save data.
+// This is how Runner can save stdout and stderr.
+// (NB: Saver is temporary until we save output into a new Snapshot)
+type Saver interface {
```
CaitieM20: Why not have these two interfaces be combined.  This seems like you just want a Output to have New or Create method associated with it.  

Or is this separate because of how you envision the snapshot migration?

CaitieM20:
LGTM, just some minor nits around comments, and a potential naming suggestion

runner/local/simple.go
```diff
@@ -59,7 +61,7 @@ func (r *simpleRunner) Run(cmd *runner.Command) runner.ProcessStatus {
 
 func makeRunnerStatus(st execer.ProcessStatus, runId runner.RunId) runner.ProcessStatus {
```
jasonbs10: Is this still used? I don't see calls in this file.

runner/local/simple.go
```diff
@@ -140,9 +143,24 @@ func (r *simpleRunner) run(cmd *runner.Command, runId runner.RunId, doneCh chan
 	}
 	defer checkout.Release()
 
+	stdout, err := r.saver.Create(fmt.Sprintf("%s-stdout", runId))
```
jasonbs10: nit: adding runner's pid seems like a good idea since runId resets on restart.

runner/local/simple.go
```diff
@@ -10,10 +10,11 @@ import (
 	"github.com/scootdev/scoot/snapshots"
 )
 
-func NewSimpleRunner(exec execer.Execer, checkouter snapshots.Checkouter) runner.Runner {
+func NewSimpleRunner(exec execer.Execer, checkouter snapshots.Checkouter, saver runner.Saver) runner.Runner {
```
jasonbs10: Not quite so simple anymore... we can revisit naming later though, when we determine if we need different runner impls.

runner/execer/os/os_execer.go
jasonbs10: Maybe copy the comment from output.go.

How will this be used as we complete milestones? Atm we want WorkerServer to save to the cloud and we'll probably want to control batching lines, tee'ing, etc, and we wouldn't implement WriterDelegate for that I think. Then for scootfs, we'll either be writing to a file and using WriterDelegate, or writing directly to manhattan/whatever without WriterDelegate. Are we sure this will be an ongoing useful construct?

jasonbs10:
lgtm for the most part.

runner/execer/os/os_execer.go
dbentley: 0) copied the comment
1) no, I'm not sure it'll be useful then, but I suspect it will be, because...
2) we may still want to write to a file and then have out code in Go tail from the file, which would prevent our Go program from blocking the underlying process
4) it not necessarily being useful is why I didn't add it to Output itself, but just to one implementation

runner/execer/os/os_execer.go
dbentley: Done; thx; whoops.

runner/local/output.go
dbentley: Done.

runner/output.go
dbentley: As discussed in-person:

Saver:FileSystem::Output:File

Renamed Saver -> OutputCreator

PR #91, opened 2016-09-01
binaries/scheduler: start the scheduler explicitly in main

dbentley:
This fixes the scheduler (whoops).

This happens because our Thrift Serving doesn't actually depend on the Scheduler, which is weird. (Instead, it depends on the Saga Coordinator and the Queue, but not the Scheduler).


CaitieM20:
LGTM

PR #92, opened 2016-09-01
scootapi/client: update smoketest on one goroutine

dbentley:
scootapi's smoke_test would sometimes deadlock. We don't actually need to update it concurrently, so just do it simply on one goroutine

### DIFF
--- a/scootapi/client/smoke_test_cmd.go
+++ b/scootapi/client/smoke_test_cmd.go
@@ -45,121 +45,46 @@ func (c *Client) runSmokeTest(cmd *cobra.Command, args []string) error {
 			return err
 		}
 	}
+	jobs := make(map[string]scoot.Status)
 	// run a bunch of concurrent jobs and track their status
-	ch := make(chan jobAndStatus)
-	timedOutJobs := make(chan error)
-	var wg sync.WaitGroup
-	errCh := make(chan error, numTasks)
+	jobIdsCh := make(chan string)
 	for i := 0; i < numTasks; i++ {
-		wg.Add(1)
-		go func() {
-			err := c.generateAndRunJob(timeout, ch)
-			if err != nil {
-				timedOutJobs <- err
-				errCh <- err
-				fmt.Println(err)
-			}
-			wg.Done()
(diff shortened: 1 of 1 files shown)


scootapi/client/smoke_test_cmd.go
```diff
@@ -45,121 +45,94 @@ func (c *Client) runSmokeTest(cmd *cobra.Command, args []string) error {
...
-	if jobInProgress {
-		return fmt.Errorf("Could Not Complete Jobs in Alloted Time %v", timeout)
-	} else {
-		return nil
+	log.Println()
+	log.Println("Job Status")
+
+	log.Println("Waiting", byStatus[scoot.Status_NOT_STARTED])
+	log.Println("Running", byStatus[scoot.Status_IN_PROGRESS])
+	log.Println("Done", byStatus[scoot.Status_COMPLETED])
```
ryancouto: Should there be a category for timed out jobs?

scootapi/client/smoke_test_cmd.go
```diff
@@ -45,121 +45,94 @@ func (c *Client) runSmokeTest(cmd *cobra.Command, args []string) error {
...
 
-	// Error Enqueuing Job
-	if err != nil {
-		switch err := err.(type) {
-		case *scoot.InvalidRequest:
-			return fmt.Errorf("Invalid Request: %v", err.GetMessage())
-		default:
-			return fmt.Errorf("Error running job: %v %T", err, err)
+func (c *Client) waitForJobs(jobs map[string]*scoot.JobStatus, timeout time.Duration) error {
+	end := time.Now().Add(timeout)
```
ryancouto: Do all jobs timeout at the same time?

CaitieM20: yes the original idea behind the timeout is the smoketest as a whole should run for "X seconds" before timing out.  In-case there is a bug and it doesn't just wait forever

CaitieM20:
LGTM

ryancouto:
lgtm too

PR #93, opened 2016-09-02
Caitie/stateful sched

CaitieM20:
Adding a Stateful Scheduler.  For now I've left the old stateless scheduler in as well to keep the review small, once this goes in I'll update config and other binaries in an additional pass and remove old files.

Stateful Scheduler keeps cluster state and job state so it can do a better job of scheduling jobs.  Currently it just schedules them on anyone that is free, but now we have a structure that can evolve over time.  

StatefulScheduler's main work is done in an event loop.  Every step it processes any messages received using Async library, and then decides which tasks to schedule and kicks off async tasks to do so.  

To test I added a new binary (temporary) smartsched to start the smartSched + apiServer
```
go run ./binaries/smartSched/main.go
```

Then run the smoke test normally

This is inspired by @dbentley hack week project.


dbentley:
LGTM

Great!

I have some small nits about how to separate the planning even more, but it doesn't need to block, and some style, but I'm getting tired tonight so let's touch base tomorrow.

CaitieM20:
Sounds good @dbentley lets do some pair programming tomorrow!

PR #94, opened 2016-09-02
Replace Scheduler to use StatefulScheduler

CaitieM20:
* updating binaries/scheduler/main.go to use StatefulScheduler
* removing old code


dbentley:
LGTM

PR #97, opened 2016-09-06
Jeanette/scoot25 job serialization

JeanetteBruno:


sched/jobDefSerializer.go
```diff
@@ -0,0 +1,112 @@
...
+
+const (
+	JsonSerialize string = "json"
+	BinarySerialize string = "binary"
+)
+
+func Deserialize(serializedVal []byte, serializeType string) (*JobDefinition, error) {
+
+	// create the thrift deserializer
+	//deserializer := thrift.NewTDeserializer();
```
CaitieM20: nit: comment deserializer := ... no longer necessary?

sched/jobDefSerializer_test.go
```diff
@@ -0,0 +1,249 @@
+package sched
+
+import (
+	"testing"
+	//"github.com/leanovate/gopter"
```
CaitieM20: nit: delete not needed?

sched/sagaLog.thrift
```diff
@@ -0,0 +1,16 @@
+
```
CaitieM20: Suggestion: Rename JobDef.thrift or Definitions.thrift.  This isn't necessarily specific to the SagaLog.  

The SagaLog just stores opaque bytes doesn't care about the format.  Also this serializer/deserializer will be used to place jobs on the Work Queue too.

sched/jobDefSerializer.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	// create the thrift jobDef struct for the saga log
+	sagaLogJobDef := sagalog.NewJobDefinition()
+	// parse the byte string into the thrift jobDef struct
+	if err := deserializer.Read(sagaLogJobDef, serializedVal); err != nil {
+		return nil, err
+	}
+
+	// make a sched JobDefintion from the thrift JobDefinition for the saga log
```
dbentley: This seems like a great thing to hoist into a function...

In fact, it might be in scootapi/server/serve.go (in thriftJobToScoot)?

sched/jobDefSerializer.go
```diff
@@ -0,0 +1,112 @@
...
+	case BinarySerialize:
+		serializer := thrift.NewTSerializer()
+		return serializer, nil
+	default:
+		return nil, fmt.Errorf("Invalid serializer type: %s\n", serializeType)
+	}
+
+}
+
+func getDeserializer(serializeType string) (*thrift.TDeserializer, error) {
```
dbentley: You could just have vars for jsonDeserializer and jsonSerializer?

Go makes it super easy to make an interface and then just add a function to an otherwise-empty struct?

sched/jobDefSerializer.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/scootdev/scoot/sched/gen-go/sagalog"
+	"github.com/scootdev/scoot/runner"
+	"time"
+	"fmt"
+)
+
+const (
+	JsonSerialize string = "json"
+	BinarySerialize string = "binary"
```
ryancouto: nit: running "go fmt" should create whitespace here to line up the two const declarations

sched/sagaLog.thrift
```diff
@@ -0,0 +1,16 @@
...
+struct Command {
```
dbentley: I'd comment why this is different from scoot.thrift (which we discussed in the chat room, so I think we're all on the same page)

sched/jobDefSerializer_test.go
```diff
@@ -0,0 +1,249 @@
...
+			// compare the orig and generated task definitions
+		} else if ok, msg := origJobDef.Equal(*newJobDef); !ok {
+			panic(fmt.Sprintf("fail: task definitions are not equal: %s\n", msg))
+		} else {
+			fmt.Printf(fmt.Sprintf("ok=%t\n", ok))
+		}
+	}
+}
+
+func makeSampleJobDefintion() JobDefinition {
```
CaitieM20: We already have some helper Job Generators in scoot/sched/generators.go.  
Suggest replacing method with GenJob().JobDefinition.  

Or refactor GenJob() function down into two functions GenJob() and GenJobDef()

sched/jobDefSerializer_test.go
```diff
@@ -0,0 +1,249 @@
...
+		taskDefMap[taskName] = taskDef
+
+	}
+	jobDef.Tasks = taskDefMap
+
+	return jobDef
+}
+
+// generate a random string for the type
+func genString(genParams *gopter.GenParameters) string {
```
CaitieM20: We have this method in a common directory scoot/tests/testhelpers/generators.go  GenRandomAlphaNumericString() if you want to reuse that!

sched/jobDefSerializer_test.go
```diff
@@ -0,0 +1,249 @@
...
+func genJobDef() gopter.Gen {
+	return func(genParams *gopter.GenParameters) *gopter.GenResult {
+		jobDef := genJobDefFromParams(genParams)
+		//Print(jobDef)
+		genResult := gopter.NewGenResult(jobDef, gopter.NoShrinker)
+		return genResult
+	}
+}
+
+func genJobDefFromParams(genParams *gopter.GenParameters) JobDefinition {
```
CaitieM20: Can you move this much better random JobDef function into scoot/sched/generators.go 

That would help make that code better overall!

sched/jobDefSerializer_test.go
```diff
@@ -0,0 +1,249 @@
...
+		},
+
+		genJobDef(),
+	))
+
+	properties.TestingRun(t)
+
+}
+
+func genJobDef() gopter.Gen {
```
CaitieM20: suggest moving this into scoot/sched/generators.go

sched/sagaLog.thrift
```diff
@@ -0,0 +1,16 @@
...
+  1: list<string> argv,
+  2: map<string, string> envVars,
+  3: i64 timeout,
+  4: string snapshotId,
+}
+
+struct TaskDefinition {
+  1: Command command,
+}
+
```
CaitieM20: Can you add struct Job which mimics Job struct in definitions.go?  We'll need the Id to be stored in the WorkQueue & the SagaLog when we store this information.

sched/jobDefSerializer_test.go
```diff
@@ -0,0 +1,249 @@
...
+	args := []string{"arg1", "arg2"}
+	taskDefinition.Argv = args
+	jobDef.Tasks["task0"] = taskDefinition
+	//Print(jobDef)  -I enable this for debugging
+
+	return jobDef
+}
+
+
+func (this *JobDefinition) Equal(actualJobDef JobDefinition) (bool, string) {
```
CaitieM20: Would recommend adding this function to definitions.go for readability.  We follow the convention that all functions on a struct are defined in the same file as the struct definition

sched/jobDefSerializer_test.go
```diff
@@ -0,0 +1,249 @@
...
+
+	expectedMap := (*this).Tasks
+	actualMap := actualJobDef.Tasks
+
+
+	if len(expectedMap) != len(actualMap) {
+		return false, fmt.Sprintf("task Definitions maps two different lengths: expected %d, got %d", len(expectedMap), len(actualMap))
+	}
+
+	for taskName, _ := range expectedMap {
```
CaitieM20: nit: maybe also create a Equal method on TaskDefinition and add it to definitions.go.  This will be useful in other test cases as well!

sched/jobDefSerializer_test.go
dbentley: Oooh, it might be nice, actually, to keep this in _test.go so that it's only included in tests and not in binaries......

sched/jobDefSerializer_test.go
```diff
@@ -0,0 +1,249 @@
...
+	for taskName, entry1 := range expectedMap {
+		if entry2, ok := actualMap[taskName]; !ok || entry1 != entry2 {
+			return false, fmt.Sprintf("The map entries (expected %s, got %s) for %s have different values.", expectedMap[taskName], actualMap[taskName], taskName)
+		}
+	}
+
+	return true, ""
+}
+
+func stringSliceCompare(expectedSlice, actualSlice []string) (bool, string) {
```
CaitieM20: golang reflect.DeepEqual() might do what you want here without having to write a method to do it.  

http://stackoverflow.com/questions/15311969/checking-the-equality-of-two-slices

sched/jobDefSerializer_test.go
```diff
@@ -0,0 +1,249 @@
...
+
+const serializeType = BinarySerialize
+
+
+func Test_JobDefCompare(t *testing.T) {
+	origJobDef := makeSampleJobDefintion()
+
+	// validate comparison logic
+	if ok, msg := origJobDef.Equal(origJobDef); !ok {
+		panic(fmt.Sprintf("error: comparison logic doesn't work comparing an object against itself %s\n", msg))
```
CaitieM20: recommend failing test cases with t.Error("error message") or t.Errorf("formated error message")
https://golang.org/pkg/testing/#B.Errorf

This will mark the test case as failed, and allow other test cases to run.  I believe panic will stop all the tests from executing if it fails.

sched/jobDefSerializer_test.go
```diff
@@ -0,0 +1,249 @@
...
+	if ok, msg := origJobDef.Equal(actualJobDef); ok {
+		panic(fmt.Sprintf("error: comparison logic doesn't work comparing an object against nil %s\n", msg))
+	}
+
+}
+
+func Test_NilJobDefinition(t *testing.T) {
+	if ok, err := Serialize(nil, serializeType); err == nil {
+		panic("error: should have gotten an error from serializing a nil object\n")
+		fmt.Printf(string(ok))
```
CaitieM20: nit: would remove the print statement, typically if tests pass there isn't output

sched/jobDefSerializer_test.go
CaitieM20: Oh I see here you want a fixed jobDef not a random one cool!

sched/jobDefSerializer_test.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+		if def1Task.SnapshotId != def2Task.SnapshotId {
+			return false, fmt.Sprintf("Snapshot ids differ, expected %s, got %s", def1Task.SnapshotId, def2Task.SnapshotId)
+		}
+	}
+
+	return true, ""
+}
+
+func stringMapEqual(expectedMap, actualMap map[string]string) (bool, string) {
```
CaitieM20: golang reflect.DeepEqual() might do what you want here without having to write a method to do it.  

http://stackoverflow.com/questions/15311969/checking-the-equality-of-two-slices

sched/jobDefSerializer_test.go
```diff
@@ -0,0 +1,249 @@
...
+			fmt.Printf(fmt.Sprintf("\tenvVar:%s = %s\n", envVarName, envVarVal))
+		}
+		for i, arg := range taskDef.Argv {
+			fmt.Printf(fmt.Sprintf("\targ[%d]:%s\n", i, arg))
+		}
+	}
+	fmt.Printf("\n")
+}
+
+func Test_RandomSerializerDeserializer(t *testing.T) {
```
CaitieM20: This is great!

CaitieM20:
Looks good some style comments.  Also you can click the Details link to see why the TravisCI build fails.  

Looks like go vet is failing.  You can run this locally by running make vet from source (excludes vendor/dependency directories)

sched/jobDefSerializer.go
CaitieM20: The one in serve.go is different its our API format to Scheduler object.  (They are super similar) but we don't want to couple our API object to our internal implementation object.  I would recommend leaving as is

sched/jobDefSerializer_test.go
CaitieM20: Its in the same package though, so right now unless we do the test case package aliasing its available to the whole package ... 

I think Equal is a generically useful function.  No reason to isolate to test only.

sched/jobDefSerializer_test.go
JeanetteBruno: good thought - unfortunately DeepEqual() returns false negative when slices and maps are empty

sched/jobDefSerializer_test.go
dbentley: Files named _test.go are included only in the compile for tests. It's like a build directive. So putting it in _test.go (even if in the same package, which, yes, my bad, I've been doing it the wrong way) will include it for tests but not in the binary.

It's fine if we want it in general; I'm -0 on it just because it's creeping featurism and it's not super clear why you'd need it, and definitions of equality often end up confusing later down the road.

JeanetteBruno:
I did not move the gopter job generation to generators.go because this generator is only for testing.  Note: should GenJobId() gopter.Gen be removed from generators.go? (Intellij is telling me that is isn't used anywhere and the gopter reference makes me think it is only for testing)

sched/jobDefSerializer.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"fmt"
+	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/scootdev/scoot/runner"
+	"github.com/scootdev/scoot/sched/gen-go/schedthrift"
+	"time"
+)
+
+type JobSerializer struct {
+	writer thrift.TSerializer
```
dbentley: I'd leave these as pointers

sched/jobDefSerializer.go
```diff
@@ -0,0 +1,113 @@
...
+
+	var schedJob = Job{}
+	schedJob.Id = internalJobDef.ID
+	sagaLogJobDef := internalJobDef.JobDefinition
+
+	var schedJobDef = JobDefinition{}
+	schedJobDef.JobType = sagaLogJobDef.JobType
+	schedJobDef.Tasks = make(map[string]TaskDefinition)
+	for taskName, sagaLogTaskDefTask := range sagaLogJobDef.Tasks {
+		argvs := (*(*sagaLogTaskDefTask).Command).Argv
```
dbentley: Do you need all these *'s? Go lets you dereference pointers automatically with .

(That is, C's . and -> both collapse to . in golang)

sched/jobDefSerializer.go
```diff
@@ -0,0 +1,113 @@
...
+		snapshotId := (*(*sagaLogTaskDefTask).Command).SnapshotId
+		command := runner.NewCommand(argvs, envVars, timeout, snapshotId)
+		schedJobDef.Tasks[taskName] = TaskDefinition{*command}
+	}
+	schedJob.Def = schedJobDef
+	return schedJob
+}
+
+// NOTE: leaving this method in this file so it lives close to the Deserializer
+func (job *Job) SerializeJobDef(serializer JobSerializer) ([]byte, error) {
```
dbentley: Why put this on *Job, as opposed to just having it be a function (not bound to a type) that takes a job and a serializer?

sched/jobDefSerializer.go
```diff
@@ -0,0 +1,113 @@
+package sched
```
dbentley: Sorry for missing this initially, I think we're trying to prefer foo_bar.go to fooBar.go

sched/jobDefSerializer.go
```diff
@@ -0,0 +1,113 @@
...
+	jobSerializer := JobSerializer{writer, reader}
+
+	return jobSerializer
+}
+
+func getBinarySerializer() JobSerializer {
+
+	writer := thrift.NewTSerializer()
+	reader := thrift.NewTDeserializer()
+	jobSerializer := JobSerializer{*writer, *reader}
```
dbentley: Dereferencing pointers in Go is a little odd; generally if something gives you a pointer you end up passing it down as a pointer, so maybe JobSerializer could store pointers instead of structs.

sched/jobDefSerializer.go
```diff
@@ -0,0 +1,113 @@
...
+	schedJob.Id = internalJobDef.ID
+	sagaLogJobDef := internalJobDef.JobDefinition
+
+	var schedJobDef = JobDefinition{}
+	schedJobDef.JobType = sagaLogJobDef.JobType
+	schedJobDef.Tasks = make(map[string]TaskDefinition)
+	for taskName, sagaLogTaskDefTask := range sagaLogJobDef.Tasks {
+		argvs := (*(*sagaLogTaskDefTask).Command).Argv
+		envVars := (*(*sagaLogTaskDefTask).Command).EnvVars
+		timeout, _ := time.ParseDuration(fmt.Sprintf("%dns", (*(*sagaLogTaskDefTask).Command).Timeout))
```
dbentley: Why sprintf just to parse? This should be just a multiplication (from ms to ns or vice versa?)

CaitieM20:
LGTM!

sched/jobDefSerializer_test.go
JeanetteBruno: done

sched/jobDefSerializer.go
JeanetteBruno: done

sched/jobDefSerializer.go
JeanetteBruno: this was the only way I could get a Duration object - needed to use time.ParseDuration()

sched/jobDefSerializer.go
JeanetteBruno: I removed them

sched/jobDefSerializer.go
JeanetteBruno: I pulled it into a private method (for readability, not reusability)

sched/jobDefSerializer.go
JeanetteBruno: removed

sched/jobDefSerializer.go
JeanetteBruno: done - thanks!

sched/jobDefSerializer.go
JeanetteBruno: ok - making the change

sched/jobDefSerializer.go
JeanetteBruno: fixed

sched/jobDefSerializer.go
JeanetteBruno: done

dbentley:
LGTM

sched/job_def_serializer.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+type Serializer interface {
+ 	Read(msg thrift.TStruct, b []byte) (err error)
+	Write(obj thrift.TStruct) (b []byte, err error)
+	DeserializerClose()
+}
+
+type JobSerializer struct {
+	*thrift.TSerializer
```
dbentley: Give these field names to make it more explicit.

JeanetteBruno:
fixed tests to be able to use DeepEqual() and removed Equal() methods in definitions.go

Updated the serialization/deserialization to create new serializers with each request - thrift serialization appears to be state-full giving corrupted results when reused.

Eliminated the need to parse the duration from a string.

sched/job_def_serializer.go
```diff
@@ -0,0 +1,136 @@
...
+// Binary behavior
+func (s binarySerializer) Deserialize(targetStruct thrift.TStruct, sourceBytes []byte) (err error) {
+	d := thrift.NewTDeserializer()
+	err = d.Read(targetStruct, sourceBytes)
+	return err
+}
+
+func (s binarySerializer) Serialize(sourceStruct thrift.TStruct) (b []byte, err error) {
+	d := thrift.NewTSerializer()
+	serializedValue, err := d.Write(sourceStruct)
```
dbentley: FWIW, you can just:

return d.Write(sourceStruct)

(Also, I might just rename sourceStruct to in)

dbentley:
LGTM

PR #99, opened 2016-09-08
runner/execer/fake: interceptor execer picks execer based on argv

dbentley:
InterceptorExecer looks at its first argument and, if it matches a condition, passes it to an execer (the interceptor) or if not passes it to the default.

This lets us include both sim and os execer in the same binary and then pick which to use at command runtime based n the command-line. So we can still simulate commands, or run them via the OS.

### DIFF
--- a/binaries/workerserver/main.go
+++ b/binaries/workerserver/main.go
@@ -8,7 +8,8 @@ import (
 
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/scootdev/scoot/common/endpoints"
-	"github.com/scootdev/scoot/runner/execer/fake"
+	"github.com/scootdev/scoot/runner/execer/execers"
+	osexec "github.com/scootdev/scoot/runner/execer/os"
 	localrunner "github.com/scootdev/scoot/runner/local"
 	fakesnaps "github.com/scootdev/scoot/snapshots/fake"
 	"github.com/scootdev/scoot/workerapi/server"
@@ -31,7 +32,9 @@ func main() {
 	if err != nil {
 		log.Fatal("Error creating OutputCreatorr: ", err)
 	}
-	run := localrunner.NewSimpleRunner(fake.NewSimExecer(nil), fakesnaps.MakeInvalidCheckouter(), outputCreator)
+
+	ex := execers.MakeSimExecerInterceptor(execers.NewSimExecer(nil), osexec.NewExecer())
+	run := localrunner.NewSimpleRunner(ex, fakesnaps.MakeInvalidCheckouter(), outputCreator)
 	version := func() string { return "" }
 	handler := server.NewHandler(stats, run, version)
 	err = server.Serve(handler, fmt.Sprintf("localhost:%d", *thriftPort), transportFactory, protocolFactory)
--- a/binaries/daemon/main.go
+++ b/binaries/daemon/main.go
@@ -6,7 +6,7 @@ import (
 
 	"github.com/scootdev/scoot/daemon/server"
 	"github.com/scootdev/scoot/runner/execer"
-	"github.com/scootdev/scoot/runner/execer/fake"
+	"github.com/scootdev/scoot/runner/execer/execers"
 	"github.com/scootdev/scoot/runner/execer/os"
 	"github.com/scootdev/scoot/runner/local"
 	fakesnaps "github.com/scootdev/scoot/snapshots/fake"
@@ -20,7 +20,7 @@ func main() {
 	var ex execer.Execer
 	switch *execerType {
 	case "sim":
-		ex = fake.NewSimExecer(nil)
+		ex = execers.NewSimExecer(nil)
 	case "os":
 		ex = os.NewExecer()
 	default:
--- a/runner/execer/execers/intercept.go
+++ b/runner/execer/execers/intercept.go
@@ -1,0 +1,38 @@
+package execers
+
+import (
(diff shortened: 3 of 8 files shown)


ryancouto:
LGTM

runner/execer/fake/intercept.go
```diff
@@ -0,0 +1,31 @@
+package fake
```
CaitieM20: is this really a fake, my understanding is this was intending to be ran side by side with OS Execer in Prod.  Fake implies test use only to me.

runner/execer/fake/intercept.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/scootdev/scoot/runner/execer"
+)
+
+type InterceptExecer struct {
```
CaitieM20: Comments Please on public methods

runner/execer/fake/intercept.go
```diff
@@ -0,0 +1,31 @@
...
+func (e *InterceptExecer) Exec(command execer.Command) (execer.Process, error) {
+	if e.Condition(command) {
+		return e.Interceptor.Exec(command)
+	} else {
+		return e.Default.Exec(command)
+	}
+}
+
+func StartsWithSimExecer(cmd execer.Command) bool {
+	return len(cmd.Argv) > 0 && cmd.Argv[0] == "#! sim execer"
```
CaitieM20: can we make this string a constant?  This way it can easily be referenced in say test cases or generators?

runner/execer/fake/intercept_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// block1 will block on wg1; block2 on wg2
+	block1 := execer.Command{
+		Argv: []string{"#! sim execer", "pause", "complete 0"},
+	}
+	block2 := execer.Command{
+		Argv: []string{"pause", "complete 0"},
+	}
+
+	wg2.Add(1)
```
CaitieM20: why are you using sync wait groups?  There aren't any go routines being executed right?  Everything is running synchronously in this test, or am I missing something?

runner/execer/fake/intercept.go
dbentley: Done.

CaitieM20:
LGTM once comments are addressed

runner/execer/fake/intercept_test.go
dbentley: Yeah, execer runs commands in the background.

This is commented in runner/execer/execer.go:
"	// Starts process to exec command in a new goroutine.
	Exec(command Command) (Process, error)"

CaitieM20: oh yes right! Thanks for the knowledge!

runner/execer/fake/intercept.go
dbentley: Done.

runner/execer/fake/intercept.go
dbentley: Yup, you're right. Ideas I have to solve it:
a) just accept it's a little odd
b) make a new package runner/execer/intercept that holds this
c) rename runner/execer/fake to a new package runner/execer/execers that holds the SimExecer and the InterceptExecer
d) something else better I haven't thought of (but hopefully you have!)

WDYT?

dbentley: The more I think about it, the more I like the general pattern to have:

bar/foo: the package that defines interfaces
bar/foo/foos: the package that contains simple implementations (fakes, mocks, composites, etc.)

This gets us two main benefits:
*) separation between interface and implementation
*) imports don't have lots of packages like "fake" or "memory" or "simple"

binaries/workerserver/main.go
```diff
@@ -8,7 +8,8 @@ import (
 
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/scootdev/scoot/common/endpoints"
-	"github.com/scootdev/scoot/runner/execer/fake"
+	"github.com/scootdev/scoot/runner/execer/execers"
+	osexec "github.com/scootdev/scoot/runner/execer/os"
```
jasonbs10: Did we decide against having everything in ./runner/execer, or the interface in that dir and *all* implementations in ./runner/execer/execers? While the asymmetry of having multiple execers in one dir and prod execer in another does signify intent, it doesn't truly seem necessary. Glad to be rid of 'fake' either way.

jasonbs10:
lgtm

binaries/workerserver/main.go
dbentley: In general, we *can't* have every implementation in any one package. And that's part of the point of golang: implementations don't have to be related to the interface.

But I think Caitie made the perfectly cromulent point that for many interfaces, there are simple impls that are not fake but want to live together. I think for these, it's good to lump them together.

I think execer/os is right on the line. But, e.g., I think a git-based implementation of snapshot is too complex, and should be in its own package.

WDYT?

dbentley: BTW, I'd suggest the positive example of https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executors.html , where Foos is used to store many Foos, but not all.

jasonbs10: I lean towards folding /os into the /execers since it's not all that
complicated. For more complicated impls, I imagine they'll live in an
entirely different directory (i.e. not in runner/execer), particularly due
to the twitter/oss divide. Ultimately it's preference so I don't have a
convincing argument for either way.

PR #100, opened 2016-09-08
workerapi/client: refactor how we create commands

dbentley:
Create a small interface cmd. Each cmd is responsible for:
*) holding its data (the struct fields)
*) registerFlags: registers flags (which will populate its struct fields)
*) run: does its job


dbentley:
This is lower priority, but will help improve workercl for running pants jobs.

CaitieM20:
LGTM

ryancouto:
LGTM

PR #101, opened 2016-09-13
Caitie/remove work queue

CaitieM20:
Removing Work Queue from Scheduler.  Now ScootAPI talks directly to the Scheduler
*  Deleted sched/Queue libraries
* Refactored ScootAPI so RunJob code is in a separate file.  This is a pattern set by Status API.  
* ScootAPI uses uuid generator to assign JobID
* ScootAPI now just takes a Scheduler in the future it can take a Scheduler Service (resolver string etc...) should we want to deploy ScootAPI and Scheduler separately
* Updated Dependencies to include UUID Library

This change looks big but its mostly just deleted files & updating vendor dependencies


binaries/scheduler/main.go
```diff
@@ -29,12 +28,15 @@ type servers struct {
...
 	http   *endpoints.TwitterServer
 	sched  scheduler.Scheduler
-	queue  queue.Queue
 	stat   stats.StatsReceiver
 }
 
-func makeServers(thrift thrift.TServer, http *endpoints.TwitterServer, sched scheduler.Scheduler, queue queue.Queue, stat stats.StatsReceiver) servers {
-	return servers{thrift, http, sched, queue, stat}
+func makeServers(
+	thrift thrift.TServer,
```
dbentley: Do you still need all of these args? e.g., do you still need the scheduler if the http server has a reference to it?

deps.sh
```diff
@@ -2,6 +2,9 @@
 # Add new direct and transitive dependencies as submodules.
 # Note: caller must validate/curate/commit changes to .gitmodules and 'vendor/'
 #
+echo "Note: caller must validate/curate/commit changes to .gitmodules and 'vendor/'"
```
dbentley: Is there a way to run make test in a way where it *won't* get/install what's needed so we'll get a loud break in TravisCI instead of "oh it silently works but without versioning"?

scootapi/server/runJob.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	job := sched.Job{
+		Id:  generateJobId(),
+		Def: jobDef,
+	}
+
+	err = scheduler.ScheduleJob(job)
```
dbentley: How do you think about generating the job ID in the scheduler vs. the server package?

I think it's weird for the scheduler to have to think about "what if a client gives me duplicate job ids", so I'd consider moving it into there. (Perhaps even into the saga log? Like, StartSaga could create the new ID, which might be totally natural within certain implementations of Saga)

scootapi/server/runJob.go
```diff
@@ -0,0 +1,93 @@
+package server
```
dbentley: change runJob.go to run_job.go ?

dbentley:
LGTM

Overall: very yay!

Biggest nit: I think it'd be nice to move the ID generation down so that the server package has less and less logic and is *just* the translation between thrift and scoot

Idle observation: it's annoying that our deps.sh means we pull in so damn many deps that we don't end up using... It makes sorting through our actual runtime dependencies kinda difficult...

jasonbs10:
There's a comment in .gitmodules alluded to by deps.sh:

# Everything after here is a transitive dependency that is not required to
use scoot.
# Note: this division should be curated whenever the dependencies change.
# See: github.com/pantsbuild/pants/issues/3606

So it's manual right now but there is a place listing direct deps.

binaries/scheduler/main.go
CaitieM20: fixed

deps.sh
CaitieM20: So if it doesn't add the dependency correctly TravisCI will break.  @jasonbs10 can probably answer this question better.  He asked me to add the echo so the comment actually gets seen when the script is ran.

scootapi/server/runJob.go
CaitieM20: Let me think about it, i think this is probably ok.  I just want to let it simmer.  Will add in a subsequent code review if that's the case.  Will be an easy refactor

scootapi/server/runJob.go
CaitieM20: done, and i will do another code review to make the codebase consistent with this naming convention, since its mostly me who messed it up :)

scootapi/server/runJob.go
CaitieM20: ok yes thought about this! I will move this into scheduler.  Will submit in subsequent code review since it'll involve modifying a bunch of files & the Scheduler interface

PR #102, opened 2016-09-15
workerapi: unify Client with Runner

dbentley:
sched/worker/rpc/thriftWorker conflated both calling the RPC and polling it.

Create sched/worker/workers:PollingWorker to do just the polling.

Then make Client extend runner.Runner

This required unifying the interfaces a bit more, so runner.Runner is now back to returning an error.


dbentley:
https://phabricator.twitter.biz/D8182

JeanetteBruno:
LGTM - as per our conversation, we need a test for polling.go.  Once you check in these changes I'll be able merge my changes to taskRunner - storing the info from ProcessStatus into saga.

PR #105, opened 2016-09-16
Rename files to be snake case

CaitieM20:
We had inconsistent file names.  We should be using snake_case to name our files.  Contains no code changes just file re-names


dbentley:
LGTM

PR #106, opened 2016-09-16
Jeanette/scoot135 expose saga message

JeanetteBruno:
exposing SagaMessage (for creating a matcher in testRunner) and moved inMemorySagaLog to  a new saga/logs package


saga/sagalogs/inMemorySagaLog.go
```diff
@@ -1,8 +1,9 @@
-package saga
+package sagalogs
```
dbentley: You'll probably want to rename this to inmemory_saga_log.go (and rectify with Caitie's concurrent change....)

dbentley:
LGTM

PR #107, opened 2016-09-16
scootapi/client: create test_targets command

ryancouto:
scootapi test_targets will run all pants tests we expect to pass


scootapi/client/targets.go
```diff
@@ -0,0 +1,7 @@
+package client
+
+// the path to any pants tests that should pass go here,
+// scootapi test_targets will test all of them.
```
dbentley: I'd add a TODO(rcouto): this is a big hack; move this into a file and read it (also, move this into a separate script)

scootapi/client/test_targets_cmd.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		return err
+	}
+	jobDef := scoot.NewJobDefinition()
+	jobDef.Tasks = make(map[string]*scoot.TaskDefinition)
+	// create a task for each target to test
+	for i, t := range getTargets() {
+		task := scoot.NewTaskDefinition()
+		task.Command = scoot.NewCommand()
+		task.Command.Argv = []string{"./pants", "test", t + ":"}
```
dbentley: Long story, but the : is part of the target name; eventually we'll find multiple targets in one package, so just put it in the list above

scootapi/client/test_targets_cmd.go
```diff
@@ -0,0 +1,48 @@
...
+		jobDef.Tasks["task"+strconv.Itoa(i)] = task
+	}
+	log.Println("Testing targets")
+	_, err = client.RunJob(jobDef)
```
dbentley: Capture the ID and print it so we can follow up on it later?

scootapi/client/test_targets_cmd.go
```diff
@@ -0,0 +1,48 @@
...
+	jobDef := scoot.NewJobDefinition()
+	jobDef.Tasks = make(map[string]*scoot.TaskDefinition)
+	// create a task for each target to test
+	for i, t := range getTargets() {
+		task := scoot.NewTaskDefinition()
+		task.Command = scoot.NewCommand()
+		task.Command.Argv = []string{"./pants", "test", t + ":"}
+		jobDef.Tasks["task"+strconv.Itoa(i)] = task
+	}
+	log.Println("Testing targets")
```
dbentley: Still useful? Maybe just print the jobDef if anything?

scootapi/client/test_targets_cmd.go
```diff
@@ -0,0 +1,48 @@
...
+		return err
+	}
+	jobDef := scoot.NewJobDefinition()
+	jobDef.Tasks = make(map[string]*scoot.TaskDefinition)
+	// create a task for each target to test
+	for i, t := range getTargets() {
+		task := scoot.NewTaskDefinition()
+		task.Command = scoot.NewCommand()
+		task.Command.Argv = []string{"./pants", "test", t + ":"}
+		jobDef.Tasks["task"+strconv.Itoa(i)] = task
```
dbentley: I'd just use the target name as the taskname, so, like:

jobDef.Tasks["task-" + test] = task

?

dbentley:
LGTM

PR #109, opened 2016-09-19
[scheduler] refactor - move jobId generation to scheduler

CaitieM20:
Currently the JobId is generated in ScootAPI.  Logically it makes more sense for the scheduler to own this and return an Id if the job was successfully scheduled.  No functionality changes occur in this code review

No functionality Changed
* Updated Scheduler Interface to take a JobDefinition and return a JobId 
* Updated Generators 
* Moved JobId generation from scootapi to statefulscheduler

### DIFF
--- a/sched/scheduler/stateful_scheduler.go
+++ b/sched/scheduler/stateful_scheduler.go
@@ -127,12 +127,15 @@ func (s *statefulScheduler) ScheduleJob(jobDef sched.JobDefinition) (string, err
 
 // generates a jobId using a random uuid
 func generateJobId() string {
-	id, err := uuid.NewV4()
-	for err != nil {
-		id, err = uuid.NewV4()
-	}
 
-	return id.String()
+	// uuid.NewV4() should never actually return an error the code uses
+	// rand.Read Api to generate the uuid, which according to golang docs
+	// "Read always returns ... a nil error" https://golang.org/pkg/math/rand/#Read
+	for {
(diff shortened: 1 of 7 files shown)


sched/scheduler/stateful_scheduler.go
```diff
@@ -99,25 +100,39 @@ type jobAddedMsg struct {
...
 	}
 
-	return nil
+	return job.Id, nil
+}
+
+// generates a jobId using a random uuid
+func generateJobId() string {
+	id, err := uuid.NewV4()
+	for err != nil {
```
dbentley: I must have missed this before, but why would it return a transient error?

Also, I'd consider just doing:

for {
  if id, err := uuid.NewV4(); err == nil {
    return id.String()
  }
}

CaitieM20: So this is an artifact of the golang [rand.Read API](https://golang.org/pkg/math/rand/#Read) which is called as part of uuid.NewV4()  

"Read generates len(p) random bytes from the default Source and writes them into p. It always returns len(p) and a nil error. Read, unlike the Rand.Read method, is safe for concurrent use."

So basically this should never return an error, but should they make a change to the API and it does?  I think its so Rand implements the [Reader interface](https://golang.org/pkg/io/#Reader) 

I'll add this in a comment since its super weird.  I also looked it up when initially implementing, and change the for loop

PR #111, opened 2016-09-19
scootapi: refactor swarmtest and scootapi to convey address

dbentley:
scootapi: refactor how we specify commands

Delete test_targets command (this is the wrong level for it)

Store the address of scootapi as host:port in ~/.cloudscootaddr . This lets us not need to pass it as an argument or env variable, so we can run it from wherever.

Refactor swarmtest so it can just setup cloud scoot instead of running smoke test too.


binaries/setup-cloud-scoot/main.go
```diff
@@ -0,0 +1,22 @@
+package main
```
CaitieM20: nit: just to be super clear change name to setup-local-cloud-scoot

dbentley: I think this just connects you to a cloud scoot, which if we had an open source cloud scoot that ran in the cloud, this would also find the address (using service discovery) to connect you to that, so the fact that it's local is just kinda temporary.

PR #112, opened 2016-09-19
Jeanette/scoot135 scoot api return task data

JeanetteBruno:
with this change scoot api returns the processStatus from the tasks that was stored in the sagalog end task messages.  It also includes job serialization refactored to use the thrifthelpers.thrift_serializer.go
Note: I have not validated this change in source yet - I will do so tomorrow morning.


sched/generators.go
```diff
@@ -50,3 +52,57 @@ func GenJobId() gopter.Gen {
...
+func makeDomainJobFromThriftJob(thriftJob *schedthrift.Job) *Job {
+
+	var domainJob = Job{}
+	domainJob.Id = thriftJob.ID
+	sagaLogJobDef := thriftJob.JobDefinition
+
+	var schedJobDef = JobDefinition{}
+	schedJobDef.JobType = sagaLogJobDef.JobType
+	schedJobDef.Tasks = make(map[string]TaskDefinition)
+	for taskName, sagaLogTaskDefTask := range sagaLogJobDef.Tasks {
```
dbentley: sagaLogTaskDefTask is a mouthful... maybe just task?

sched/generators.go
```diff
@@ -50,3 +52,57 @@ func GenJobId() gopter.Gen {
 		return gopter.NewGenResult(string(result), gopter.NoShrinker)
 	}
 }
+
+func makeDomainJobFromThriftJob(thriftJob *schedthrift.Job) *Job {
+
+	var domainJob = Job{}
```
dbentley: domainJob := &Job{
  Id: thriftJob.ID
}

?

sched/generators.go
```diff
@@ -50,3 +52,57 @@ func GenJobId() gopter.Gen {
...
+	schedJobDef.JobType = sagaLogJobDef.JobType
+	schedJobDef.Tasks = make(map[string]TaskDefinition)
+	for taskName, sagaLogTaskDefTask := range sagaLogJobDef.Tasks {
+		var argvs []string
+		argvs = sagaLogTaskDefTask.Command.Argv
+		var envVars map[string]string
+		envVars = sagaLogTaskDefTask.Command.EnvVars
+		timeout := time.Duration(sagaLogTaskDefTask.Command.Timeout)
+		snapshotId := sagaLogTaskDefTask.Command.SnapshotId
+		command := runner.NewCommand(argvs, envVars, timeout, snapshotId)
```
dbentley: You create all of these varaibles just to use them once...

Could you just do:
  cmd := task.Command
 runner.NewCommand(cmd.Argv, cmd.EnvVars, cmd.Timeout, cmd.SnapshotId)

?

sched/generators.go
```diff
@@ -50,3 +52,57 @@ func GenJobId() gopter.Gen {
...
+func makeThriftJobFromDomainJob(domainJob *Job) (*schedthrift.Job, error) {
+	if domainJob == nil {
+		return nil, nil
+	}
+
+	// allocate a thrift Job for saga log
+	thriftJob := schedthrift.NewJob()
+
+	// copy the sched JobDefinition properties to the thrift jobDefinition structure
+	(*thriftJob).ID = (*domainJob).Id
```
dbentley: Do you need the (*foo).Foo?

sched/generators.go
```diff
@@ -50,3 +52,57 @@ func GenJobId() gopter.Gen {
...
+
+	internalJobDef := schedthrift.NewJobDefinition()
+
+	jobDef := domainJob.Def
+	internalJobDef.JobType = jobDef.JobType
+	internalJobDef.Tasks = make(map[string]*schedthrift.TaskDefinition)
+	for taskName, taskDef := range jobDef.Tasks {
+		internalTaskDef := schedthrift.NewTaskDefinition()
```
dbentley: Same here with shorter names and fewer temporaries making more readable code?

scootapi/server/job_status.go
```diff
@@ -87,3 +92,24 @@ func convertSagaStateToJobStatus(sagaState *s.SagaState) *scoot.JobStatus {
...
+		return err
+	}
+
+	scootRunStatus.RunId = fmt.Sprintf("%s", workerRunStatus.RunId)
+	scootRunStatus.OutUri = workerRunStatus.OutUri
+	scootRunStatus.ErrUri = workerRunStatus.ErrUri
+	scootRunStatus.ExitCode = workerRunStatus.ExitCode
+	scootRunStatus.Error = workerRunStatus.Error
+
+	js.TaskData[id] = &scootRunStatus
```
dbentley: This looks like another Thrift <-> Domain translation layer, no?

I think it's easiest if those return values instead of setting. WDYT?

sched/generators.go
JeanetteBruno: done

sched/generators.go
JeanetteBruno: done

sched/generators.go
JeanetteBruno: done

sched/generators.go
JeanetteBruno: I split the difference using thriftTask vs domainTask

sched/generators.go
JeanetteBruno: done - thanks for pointing out these improvements

scootapi/server/job_status.go
JeanetteBruno: this is a thrift <-> thrift conversion due to thrift generation errors when use include.  I've updated the names to be more meaningful and added a comment.

dbentley:
LGTM

PR #113, opened 2016-09-20
snapshot/git: create checkouts with git clone --reference

dbentley:
Using git clone reference lets us have many clones without taking lots of disk space. This will let us do work in the background (eventually; not yet implemented).

### DIFF
--- a/snapshot/git/gitfiler/checkouter.go
+++ b/snapshot/git/gitfiler/checkouter.go
@@ -1,0 +1,111 @@
+package gitfiler
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"sync"
+
+	"github.com/scootdev/scoot/os/temp"
+	"github.com/scootdev/scoot/snapshot/git/repo"
+	"github.com/scootdev/scoot/snapshots"
+)
+
+// Utilities for Reference Repositories.
+// A Reference Repository is a way to clone repos locally so that the clone takes less time and disk space.
+// By passing --reference <local path> to a git clone, the clone will not copy the whole ODB but instead
+// hardlink. This means the clone is much faster and also takes very little extra hard disk space.
+// Cf. https://git-scm.com/docs/git-clone
+
+// RefRepoGetter lets a client get a Repository to use as a Reference Repository.
+type RefRepoGetter interface {
+	// Get gets the Repository to use as a reference repository.
+	Get() (*repo.Repository, error)
+}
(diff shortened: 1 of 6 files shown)


JeanetteBruno:
lgtm

snapshot/git/gitfiler/checkouter.go
```diff
@@ -0,0 +1,112 @@
...
+	log.Println("gitfiler.RefRepoCloningCheckouter.clone: Cloning", cmd)
+	err = cmd.Run()
+	if err != nil {
+		return nil, fmt.Errorf("gitfiler.RefRepoCloningCheckouter.clone: error cloning: %v", err)
+	}
+
+	return repo.NewRepository(cloneDir.Dir)
+}
+
+type RefRepoCloningCheckout struct {
```
dgassaway: This was a little confusing at first (RefRepoCloningCheckouter vs RefRepoCloningCheckout), could do something like RefRepoCloner/RefRepoCheckout. Also, seems like golang is big on single char variables, but 'repo' seems better than 'r' for *repo.Repository, unless that was intended to distinguish from member in RefRepoCloningCheckouter

snapshot/git/repo/repo.go
```diff
@@ -0,0 +1,72 @@
...
+	return validateSha(out)
+}
+
+// Checkout an id in r
+func (r *Repository) Checkout(id string) error {
+	_, err := r.Run("checkout", id)
+	return err
+}
+
+func validateSha(sha string) (string, error) {
```
dgassaway: Depending where sha could end up coming from, it might make sense to accept 41 w/ '\n' or 40 w/out

dbentley: Done.

snapshot/git/gitfiler/checkouter.go
dbentley: I hear what you're saying about Checkouter vs. Checkout. I think we should be consistent, so if we want to change it, we should consider changing it in snapshot/filer.go ; WDYT of that (preexisting) interface structure?

Single character: done.

dgassaway: Hmm. In that case I'm almost inclined to use something more generic like 'Loader' that indicates a snapshot is not necessarily a git wrapper.  I'm not strongly held on this, though, and definitely agree consistency is more important.  If no one else agrees I'd say just leave it.

PR #115, opened 2016-09-23
snapshot/git/gitfiler: refactor Checkouter

dbentley:
Break down gitfiler a bit more. 

Checkouter: Checks out a Snapshot ID by running git checkout
RepoGetter: interface to get a Repo
RefCloner: RepoGetter that gets a repo by running git clone --reference
RepoPool: holds Repos and manages concurrent access. Can get new ones if given a RepoGetter

By building these blocks together in different ways, a client can choose whether to always make a new clone (setup.go.NewRefRepoCloningCheckouter) or always use the same Repo (setup.go.NewSingleRepoCheckouter)

Document in gitfiler/doc.go

log in local.simpleRunner to make it easier to follow in workerServer's stderr

Use TempDir in more places to be more consistent about where we store output.

### DIFF
--- a/snapshot/git/gitfiler/checkouter.go
+++ b/snapshot/git/gitfiler/checkouter.go
@@ -2,57 +2,10 @@ package gitfiler
 
 import (
 	"fmt"
-	"io/ioutil"
-	"path"
 
-	"github.com/scootdev/scoot/os/temp"
 	"github.com/scootdev/scoot/snapshot"
 	"github.com/scootdev/scoot/snapshot/git/repo"
 )
-
-// Utilities for Reference Repositories.
-// A Reference Repository is a way to clone repos locally so that the clone takes less time and disk space.
-// By passing --reference <local path> to a git clone, the clone will not copy the whole ODB but instead
-// hardlink. This means the clone is much faster and also takes very little extra hard disk space.
-// Cf. https://git-scm.com/docs/git-clone
-
-// RepoGetter lets a client get a Repository to use as a Reference Repository.
-type RepoGetter interface {
-	// Get gets the Repository to use as a reference repository.
-	Get() (*repo.Repository, error)
-}
-
-func NewSingleRepoPool(repoGetter RepoGetter, doneCh chan struct{}) *RepoPool {
-	singlePool := NewRepoPool(nil, nil, doneCh)
-	go func() {
-		r, err := repoGetter.Get()
-		singlePool.Release(r, err)
-	}()
-	return singlePool
-}
-
-func NewSingleRepoCheckouter(repoGetter RepoGetter, doneCh chan struct{}) *Checkouter {
-	pool := NewSingleRepoPool(repoGetter, doneCh)
-	return NewCheckouter(pool)
-}
-
-func NewRefRepoCloningCheckouter(refRepoGetter RepoGetter, clonesDir *temp.TempDir, doneCh chan struct{}) *Checkouter {
-	refPool := NewSingleRepoPool(refRepoGetter, doneCh)
-
-	cloner := &refCloner{refPool: refPool, clonesDir: clonesDir}
-	var clones []*repo.Repository
-	fis, _ := ioutil.ReadDir(clonesDir.Dir)
-	for _, fi := range fis {
-		// TODO(dbentley): maybe we should check that these clones are in fact clones
-		// of the reference repo? Using some kind of git commands to determine its upstream?
-		if clone, err := repo.NewRepository(path.Join(clonesDir.Dir, fi.Name())); err == nil {
-			clones = append(clones, clone)
-		}
-	}
-
-	pool := NewRepoPool(cloner, clones, doneCh)
-	return NewCheckouter(pool)
-}
 
 func NewCheckouter(repos *RepoPool) *Checkouter {
 	return &Checkouter{repos: repos}
@@ -62,6 +15,8 @@ func NewCheckouter(repos *RepoPool) *Checkouter {
 type Checkouter struct {
 	repos *RepoPool
 }
+
+const DEFAULT_REV = "1dda9fbde682e4922a0d5709c5539f573db4cc54"
 
 // Checkout checks out id (a raw git sha) into a Checkout.
 // It does this by making a new clone (via reference) and checking out id.
@@ -79,6 +34,11 @@ func (c *Checkouter) Checkout(id string) (co snapshot.Checkout, err error) {
 			c.repos.Release(repo, repoErr)
 		}
 	}()
+
+	// TODO(dbentley): do more ot validate id. E.g., don't let "HEAD" or "master" slip through
+	if id == "" {
+		id = DEFAULT_REV
+	}
 
 	// -d removes directories. -x ignores gitignore and removes everything.
 	// -f is force. -f the second time removes directories even if they're git repos themselves
--- a/snapshot/git/gitfiler/checkouter.go
+++ b/snapshot/git/gitfiler/checkouter.go
@@ -16,60 +16,67 @@ import (
 // hardlink. This means the clone is much faster and also takes very little extra hard disk space.
 // Cf. https://git-scm.com/docs/git-clone
 
-// RefRepoGetter lets a client get a Repository to use as a Reference Repository.
-type RefRepoGetter interface {
+// RepoGetter lets a client get a Repository to use as a Reference Repository.
+type RepoGetter interface {
 	// Get gets the Repository to use as a reference repository.
 	Get() (*repo.Repository, error)
 }
 
-// RefRepoCloningCheckouter checks out by cloning a Ref Repo.
-type RefRepoCloningCheckouter struct {
-	clones *repoPool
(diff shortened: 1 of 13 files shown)


dgassaway:
lgtm wrt logic and approach

dgassaway:
From travis-ci:
os/temp/temp_dir.go:73: no formatting directive in Errorf call

dgassaway:
lgtm

snapshot/git/gitfiler/pool.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	go p.loop()
+	return p
+}
+
+// RepoPool lets clients Get a Repository for an operation, and then Release it when done
+// RepoPool can create a new Repository by using a supplied RepoGetter.
+// New repos can be added by the client by Release'ing a new Repository
+// Cf. sync's Pool
+type RepoPool struct {
```
jasonbs10: Can we instead get away with a pool without loop() that calls getter.Get() but simply destroys the repo when released? If reference/shared cloning is cheap then removing the directory wholesale might even be faster than running 'git clean' on every checkout. I mean this is a fairly standard pool impl so I may be overthinking this.

snapshot/git/gitfiler/checkouter.go
```diff
@@ -2,111 +2,76 @@ package gitfiler
...
-	return &RefRepoCloningCheckout{repo: clone, id: id, checkouter: c}, nil
-}
-
-func (c *RefRepoCloningCheckouter) clone() (*repo.Repository, error) {
-	cloneDir, err := c.tmp.TempDir("clone-")
-	if err != nil {
-		return nil, err
+	// -d removes directories. -x ignores gitignore and removes everything.
+	// -f is force. -f the second time removes directories even if they're git repos themselves
+	cmds := [][]string{
```
jasonbs10: This might be better pulled out to another func, that way the caller can clean/reuse without having to return to the pool. Also, see comment in pool.go about returning a clean repo.

snapshot/git/gitfiler/cloner.go
```diff
@@ -0,0 +1,45 @@
...
+	"fmt"
+	"log"
+	"os/exec"
+
+	"github.com/scootdev/scoot/os/temp"
+	"github.com/scootdev/scoot/snapshot/git/repo"
+)
+
+// A Reference Repository is a way to clone repos locally so that the clone takes less time and disk space.
+// By passing --reference <local path> to a git clone, the clone will not copy the whole ODB but instead
```
dbentley: git clone --reference uses hardlinks

So if you:

git clone --reference a a b

You'll end up with a and b. If you du a or du b, you'll see each is the same size. But if you du a b, you'll see a is big and b is tiny.

snapshot/git/gitfiler/cloner.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	cloneDir, err := c.clonesDir.TempDir("clone-")
+	if err != nil {
+		return nil, err
+	}
+
+	// We probably ought to use a separate git dir so that processes can't mess up .git
```
dbentley: I guess also so they can't look in .git. E.g., processes shouldn't be able to git log (or git checkout)

snapshot/git/gitfiler/pool.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		// Serve, either receiving or sending repos
+		select {
+		case reserveCh <- free:
+			p.freeList = p.freeList[1:]
+		case r := <-p.releaseCh:
+			p.freeList = append(p.freeList, r)
+		case r := <-p.getCh:
+			p.freeList = append(p.freeList, r)
```
dbentley: Because getCh is poorly named; it's the channel of what comes back from a get.

snapshot/git/gitfiler/setup.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Utilities for creating Checkouters.
+
+// RepoGetter lets a client get a Repository to use as a Reference Repository.
+type RepoGetter interface {
+	// Get gets the Repository to use as a reference repository.
+	Get() (*repo.Repository, error)
+}
+
+// A Pool that will only have a signle repo, populated by repoGetter, that serves as long as doneCh
```
dbentley: Err, I think I just wrote awkward English, because I thought I could omit the "as long as doneCh [is open]"

snapshot/git/gitfiler/checkouter.go
dbentley: What kind of caller would want to not return to the pool?

I worry that the caller shouldn't be making decisions about that. It's the responsibility of the Pool to allocate; the client shouldn't be second-guessing by reusing.

jasonbs10: A different question is if Checkout is required to strictly control Repository access. Once someone gets a resource its really up to them what to do with it and for how long. The Checkout type does and does not allow for that - the user can do Command{dir=Checkout.dir} to accomplish anything even if they don't have access to repo.Run(). My preference is to have fewer types and more flexibility.

snapshot/git/gitfiler/cloner.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// A Reference Repository is a way to clone repos locally so that the clone takes less time and disk space.
+// By passing --reference <local path> to a git clone, the clone will not copy the whole ODB but instead
+// hardlink. This means the clone is much faster and also takes very little extra hard disk space.
+// Cf. https://git-scm.com/docs/git-clone
+
+// cloner clones a repo using --reference based on a reference repo
+type refCloner struct {
+	refPool   *RepoPool
```
jasonbs10: Kind of, but after construction, the clones get mixed in with the reference repo, no? I think it's cleaner to have a reference repo which you only use for expensive operational things like 'git pull', and then a pool of clones for everything else.

snapshot/git/gitfiler/cloner.go
jasonbs10: We should use TODOs if we really want to revisit this code and add that restriction. Otherwise I know I won't remember and won't be able to grep for TODO/FIXME.

snapshot/git/gitfiler/pool.go
jasonbs10: I'm not following, don't we only want to log requests/actions? If nothing is happening without us making a request then what is there to log?

In terms of background work, we use the loop construct to enable concurrent requests for internal state change, but I don't think that actually enables parallelism. So if we have no internal state to protect and we're already not parallel, just do the work in the foreground?

This is kind of moot though since full checkouts are super expensive. That makes me wonder if we should improve parallelism for  new clones.

snapshot/git/gitfiler/setup.go
```diff
@@ -0,0 +1,52 @@
...
+	return NewCheckouter(pool)
+}
+
+// A Checkouter that creates a new repo with git clone --reference for each checkout
+func NewRefRepoCloningCheckouter(refRepoGetter RepoGetter, clonesDir *temp.TempDir, doneCh chan struct{}) *Checkouter {
+	refPool := NewSingleRepoPool(refRepoGetter, doneCh)
+
+	cloner := &refCloner{refPool: refPool, clonesDir: clonesDir}
+	var clones []*repo.Repository
+	fis, _ := ioutil.ReadDir(clonesDir.Dir)
```
jasonbs10: You're right, I was thinking could do a form of COW for the working directory but that's more magic FS territory. To clarify, much of this code goes away if/when we switch to magic FS?

dbentley: Yes; all of this goes away with Magic FS.

snapshot/git/gitfiler/pool.go
dbentley: When the Pool loop begins, it will start up a Get that will happen in the background. So if you start up a worker, you'll see it happening in the background before any requests arrive.

snapshot/git/gitfiler/checkouter.go
dbentley: Yup, if a client of this interface calls Release() and then continues to operate, we'll have bugs.

I think this is a reasonable compromise so that we can reuse directories.

We can't return a *Repository, because we don't want people to depend on it being a repository (hence the comment that we ought to use a separate git dir, so that what people get isn't even a repository).

snapshot/git/gitfiler/cloner.go
dbentley: In setup.go:NewRefRepoCloningCheckouter there are two pools:
  refPool holds the single reference repo
  pool holds the clones

I agree (with what I think you're saying, but maybe I'm misunderstanding?), and I think the code already does that.

snapshot/git/gitfiler/cloner.go
dbentley: I agree. My bad.

snapshot/git/gitfiler/cloner.go
jasonbs10: I think we need to revisit the names at some point, these are getting pretty long. I may be confused and I think requiring an [expandable] pool with a static single repo contributes to that.

PR #117, opened 2016-09-23
[Scheduler] Refactor Shared Service Code & Embed Config as Part of Binary

CaitieM20:
Refactored out ./binaries/scheduler/main.go code that is shared across repos to scootapi/server/setup.go.  This is a non breaking change since no functionality has been removed.
* Default returns the default ice bag for DI and the default json schema
* RunServer starts the API Server
* Defines common flags for the executable

./binaries/scheduler/main.go uses refactored server code to use server setup code

Config is no longer a flag, but is read from a file.  I used [go-bindata](https://github.com/jteeuwen/go-bindata) to create a binary representation of the file which is packaged and deployed as part of the scheduler binary.  

If this looks amenable to folks we can then refactor the workerserver the same way


binaries/scheduler/config/config.json
```diff
@@ -0,0 +1,6 @@
+{
```
dbentley: How does this work for starting a swarm test that uses the local cluster?

binaries/scheduler/config/local.json
```diff
@@ -0,0 +1,5 @@
+{
+  "Cluster": {
+    "Type": "local"
+  }
```
dbentley: Think you need Workers: {Type: rpc} as well

CaitieM20: yup you are right, whoops somehow messed that up checked in now

PR #122, opened 2016-09-27
scootapi/client: update output

dbentley:


ryancouto:
make vet failed, "scootapi/client/run_job_cmd.go:72: possible formatting directive in Println call." otherwise lgtm

PR #124, opened 2016-09-27
scootapi/client: include run status; add testcase

dbentley:


dbentley:
Note: because of revert shenanigans, just look at the second commit for now

dbentley:
OK, then I think I'll just close this because it's accidentally submitted.

PR #127, opened 2016-10-03
swarmtest: look at wait in Main, not InitOptions

dbentley:
Before this change, if a user of this library overloaded Run, we'd always run, not wait.


dbentley:
https://phabricator.twitter.biz/D9134

PR #128, opened 2016-10-03
runner/runner print formatting for process status

dgassaway:
Added String() methods to struct types: runner.ProcessStatus and runner.Command so these print human-readable.

Please note that this removed all uses of the vendor lib github.com/luci/go-render which has been removed - looking for feedback on this.


runner/runner.go
```diff
@@ -34,7 +34,7 @@ const (
 )
 
 func (p ProcessState) IsDone() bool {
-	return p == COMPLETE || p == FAILED || p == ABORTED || p == TIMEDOUT
+	return p == COMPLETE || p == FAILED || p == ABORTED || p == TIMEDOUT || p == BADREQUEST
```
dbentley: As discussed in person:

1) BADREQUEST should probably not be IsDone, because a retry could work
2) BADREQUEST is weird, and maybe we should just return an error instead of a ProcessStatus...

runner/runner.go
```diff
@@ -78,6 +78,24 @@ type Command struct {
...
 
+func (c Command) String() string {
+	sfmt := "Command\n\tSnapshot ID:\t%s\n"
+	sargs := []interface{}{c.SnapshotId}
+
+	sfmt += "\t%v\n\tTimeout:\t%v\n"
+	sargs = append(sargs, c.Argv, c.Timeout)
+
+	if len(c.EnvVars) > 0 {
+		sfmt += "\tEnv:\n"
```
dbentley: Adding to sfmt is weird... maybe just do multiple Sprintfs so you don't need to store the fmt string at all?

.gitmodules
```diff
@@ -34,10 +34,6 @@
 	path = vendor/golang.org/x/sys
 	url = https://go.googlesource.com/sys
 	branch = 62bee037599929a6e9146f29d10dd5208c43507d
-[submodule "vendor/github.com/luci/go-render"]
```
jasonbs10: I think we want to keep this. What was the rationale for removing this entirely?

runner/runner.go
jasonbs10: I'm kind of regretting BADREQUEST. I initially thought it would be a good idea in worker.thrift but didn't really want it in runner.go. We should consider converting it to an error as Dan said.

runner/runner_test.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/scootdev/scoot/runner"
+)
+
+func TestCommandStringSimple(t *testing.T) {
+	c := runner.Command{
+		Argv:       []string{"cat", "cats.in", "|", "cat.bin", ">", "cats.out"},
```
jasonbs10: If tests double as documentation, I don't think we want to pipe/redirect here. AFAIK that's not supported in golang unless it's a single arg passed to a bash binary.

runner/runner_test.go
```diff
@@ -0,0 +1,72 @@
+package runner_test
```
dbentley: Maybe move this to status_test.go?

Also, does it need to be in a separate package (runner_test instead of runner)?

runner/runner_test.go
dbentley: Good catch Jason; I agree.

dgassaway:
* Removed submodule removal for potential future reuse (was not sure if we would scrub unused dependencies).
* Refactored String() method string building style
* Moved runner_test to status_test per feedback, although I'm not sure about this one - this is directly testing code implemented in runner.go.  I went with runner_test as the package name vs. runner per cloud_test code and some community recs on black-box test pkg names

dgassaway:
* Updated command print test to be a valid command

dbentley:
Hrm, I think this is testing status, but maybe we should just move the status definition into status.go and make runner.go have the definition of just the runner?

I used to prefer the _test package style name, but found that they're kinda rarely used. As a team (but before you joined), we talked about it and said it felt like more overhead and we would prefer to not use them when not required. If you feel strongly, we can have this discussion again at our tech discussion meeting.

Also, I realize now looking at the code that maybe using a bytes.Buffer would be even simpler for what you're trying to do, of just keep writing things to an output. Cf. https://godoc.org/bytes#example-Buffer

PR #130, opened 2016-10-05
scootapi/setup: framework to setup a cloud scoot

dbentley:
scootapi/setup makes it easy to setup a cloud scoot. This framework should (eventually) replace swarmtest. It makes it easy to build a binary with an interface like:

bin/setup-cloud-scoot --strategy local.local
bin/setup-cloud-scoot --strategy local.local --workers 3 
bin/setup-cloud-scoot --strategy local.local --workers 3 run bin/scootapi run_smoke_test

I'd recommend looking at (in this order, and with annotations):
setup.go (entry point to library)
cmds.go (low-level plumbing used throughout to run commands)
sched.go (strategy to start a local scheduler)
workers.go (strategies for a local scheduler to talk to workers, either in-memory or running locally)
builder.go (how to build binaries)
ports.go (how to wait for a server to start, to prevent race conditions)
main.go (constructing the objects and then handing off control to setup.go)

This also required changing some other stuff:
1) scheduler: change how we read config
2) workerserver: take addresses instead of ports
3) cluster/local: change how we parse the output from ps


dbentley:
https://phabricator.twitter.biz/D9233

JeanetteBruno:
I'm not sure I'm understanding the changes correctly and this appears to be a significant chunk of functionality, so I started a confluence page to document (my understanding of) the design. (https://confluence.twitter.biz/display/EE/Cloud+Scoot+Test+Harness) to document the design (as I understand it).  I've put my comments on the bottom of that page.  Also, am I able to see the github repo/branch that has all the changes? (this would be easier than looking at diffs)

I like the refactoring's direction.

jasonbs10:
Keep in mind that we will be discussing the virtues of the original code vs the new code tomorrow at the tech meeting. I like the direction of the new code as well but I'm of the opinion that there is such a thing as too much modularization. Ultimately I want whatever changes the group wants. 






On Wed, Oct 5, 2016 at 7:50 AM -0700, "Jeanette Bruno" <notifications@github.com> wrote:












I'm not sure I'm understanding the changes correctly and this appears to be a significant chunk of functionality, so I started a confluence page to document (my understanding of) the design. (https://confluence.twitter.biz/display/EE/Cloud+Scoot+Test+Harness) to document the design (as I understand it).  I've put my comments on the bottom of that page.  Also, am I able to see the github repo/branch that has all the changes? (this would be easier than looking at diffs)



I like the refactoring's direction.

scootapi/server/setup.go
```diff
@@ -92,3 +95,21 @@ func RunServer(bag *ice.MagicBag, schema jsonconfig.Schema, config []byte) {
 	}()
 	log.Fatal("Error serving: ", <-errCh)
 }
+
+// GetConfigText finds the right text for a configFlag.
+// If configFlag looks like a filename (of the form foo.bar where foo and bar are just alphanumeric),
+// read it as an asset.
+// Otherwise, assume it's the literal json text.
+func GetConfigText(configFlag string, asset func(string) ([]byte, error)) ([]byte, error) {
```
CaitieM20: So I disagree with having this method in the Server code.  I think it belongs as a private method in the binary, or perhaps as a util function in the config package, since workerserver and other binaries may want to use this convention.

Putting it here couples how we read and get input for scootapi to the scootapi implementation.  

This code also needs test cases

scootapi/setup/setup.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/scootdev/scoot/scootapi"
+)
+
+// Main is the entry point for Cloud Scoot setup.
+// cmds is running commands
+// strategies maps strings to SchedulerStrategy
+// which is the name of the strategy to use
+// args is the remaining arguments
+func Main(cmds *Cmds, strategies map[string]SchedulerStrategy, which string, args []string) error {
```
CaitieM20: nit: I dont know what which means can we give this a name that is more descriptive. I think this is just the user input strategy.

scootapi/setup/setup.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+
+	"github.com/scootdev/scoot/scootapi"
+)
+
+// Main is the entry point for Cloud Scoot setup.
+// cmds is running commands
+// strategies maps strings to SchedulerStrategy
+// which is the name of the strategy to use
+// args is the remaining arguments
```
CaitieM20: can you add a comment defining what args are supported and what they do?

dbentley:
"move the setup code into the swarmtest folder / package to make it more clear that this is all code related only to this tool/script"

Do we want to keep the swarmtest name? I think we'll use this just at the start of the day for "let me talk to prod scoot", so I thought we'd move away from that. That's why I made the new package.

scootapi/setup/workers.go
```diff
@@ -0,0 +1,95 @@
+package setup
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+)
+
+// WorkersStrategy is a strategy to start workers and returns the config to pass to a scheduler to talk to them
+type WorkersStartup interface {
```
dgassaway: "WorkersStartup" or "WorkersStrategy"?

binaries/scheduler/main.go
```diff
@@ -17,18 +17,14 @@ import (
...
 func main() {
-
 	flag.Parse()
-	log.Printf("Scheduler: reading config %+v", *configFileName)
-	config, err := config.Asset(fmt.Sprintf("config/%v", *configFileName))
-
+	configText, err := server.GetConfigText(*configFlag, config.Asset)
 	if err != nil {
-		log.Fatalf("Error Loading Config File: %v, with Error: %v", configFileName, err)
+		log.Fatal(err)
```
dbentley: This error will have that context, so this change doesn't lose any info.

It's populated by scootapi/server/setup.go:GetConfigText, on line 109.

scootapi/setup/ports.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"fmt"
+	"log"
+	"os/exec"
+	"time"
+)
+
+// WaitForPort waits ~5 seconds for a process to listen to the port, and returns an error if
+// the port remains open
+func WaitForPort(port string) error {
```
dgassaway: this functionality also lives in closed source swarmtest.go, may want to make note to re-use this one later

binaries/workerserver/main.go
```diff
@@ -16,13 +15,13 @@ import (
 	"github.com/scootdev/scoot/workerapi/server"
 )
 
-var thriftPort = flag.Int("thrift_port", 9090, "port to serve thrift on")
-var httpPort = flag.Int("http_port", 9091, "port to serve http on")
+var thriftAddr = flag.String("thrift_addr", "localhost:9090", "port to serve thrift on")
```
dgassaway: Little out of scope but we may want to start defining and using ports via single point of truth package var

scootapi/server/setup.go
dbentley: Moved to config/scootconfig/config.go (and tested in config_test.go)

scootapi/setup/setup.go
dbentley: Done.

scootapi/setup/setup.go
dbentley: Done.

scootapi/server/setup.go
```diff
@@ -92,3 +95,21 @@ func RunServer(bag *ice.MagicBag, schema jsonconfig.Schema, config []byte) {
...
+	if matched, _ := regexp.Match(`[[:alnum:]]*\.[[:alnum:]]`, []byte(configFlag)); matched {
```
dbentley: I agree with being explicit in general.

In this case, I'm not concerned because text.text isn't valid JSON.

Maybe this step could *also* go into the server.RunServer (so it calls the function that's now in jsonconfig)?

scootapi/setup/cmds.go
```diff
@@ -0,0 +1,125 @@
...
+// If we really want to be correct, we have to start another process that will do
+// the babysitting.
+func (c *Cmds) Kill() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	for _, cmd := range c.watching {
+		if p := cmd.Process; p != nil {
+			log.Printf("Killing: %v\n", cmd.Path)
+			// TODO(dbentley): send a SIGINT and send a SIGKILL only if it hasn't finished later
```
dbentley: Update the TODO

scootapi/setup/workers.go
dbentley: Fixed; thanks.

scootapi/setup/ports.go
dbentley: I think swarmtest.go's can be deleted because their functionality now exsits here?

binaries/workerserver/main.go
dbentley: What's a single point of truth package var?

scootapi/server/setup.go
dgassaway: Assuming configFlag is a whole flattened json string, this will match '{"key": "text.text"}', which is valid.  This actually matches more subtle stuff like:
'{["key": "value"], ["key": ".value"]}', which is kind of what I am getting at

scootapi/server/setup.go
dbentley: Sorry, I misspoke: moved to config/jsonconfig/config.go

scootapi/setup/cmds.go
```diff
@@ -0,0 +1,125 @@
...
+
+// StartCmd starts a Cmd that was created by Command
+func (c *Cmds) StartCmd(cmd *exec.Cmd) error {
+	log.Println("Starting", cmd.Args)
+	err := cmd.Start()
+	if err == nil {
+		go func() {
+			cmd.Wait()
+			log.Printf("Cmd %v finished; killing", cmd.Path)
+			c.Kill()
```
dgassaway: Sure, but that is not clear from cmds.go and this has the same language as os/exec but differs in use

scootapi/setup/cmds.go
```diff
@@ -0,0 +1,125 @@
...
+	cmd := c.Command(path, arg...)
+	return c.RunCmd(cmd)
+}
+
+// stop watching a cmd (because it's done)
+func (c *Cmds) remove(cmd *exec.Cmd) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	for i, other := range c.watching {
+		if other == cmd {
```
dgassaway: Yeah, wishing go had a set.  This is small enough that it doesn't matter.

scootapi/setup/ports.go
dgassaway: sure

scootapi/server/setup.go
dbentley: Oh, yup, forgot to put in anchors.

binaries/workerserver/main.go
dgassaway: That sounds good to me

scootapi/setup/cmds.go
dgassaway: Sure, thank you

scootapi/setup/workers.go
```diff
@@ -0,0 +1,95 @@
...
+	return &InMemoryWorkersStrategy{workersFlag: workersFlag}
+}
+
+func (s *InMemoryWorkersStrategy) StartupWorkers() (string, error) {
+	log.Println("Using in-memory workers")
+	if s.workersFlag == "" {
+		return "local.memory", nil
+	}
+
+	return fmt.Sprintf(`{"Cluster": {"Type": "memory", "Count": %s}, "Workers": {"Type": "local"}}`, s.workersFlag), nil
```
jasonbs10: StartupWorkers can return two different types of config (strategyName or json) but they should both be representable as json right? I'd keep these consistent.

scootapi/setup/sched.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"log"
+)
+
+// SchedulerStrategy will startup a Scheduler (or setup a connection to one)
+type SchedulerStrategy interface {
+
+	// Startup starts up a Scheduler, returing the address of the server or an error
+	Startup() (string, error)
```
jasonbs10: This should be StartupScheduler() for consistency. I was also thinking about using a single StartupStrategy interface and commenting that the return value is config used by upstream components, sched returns host:port and workers returns schedConfig. WDYT?

scootapi/setup/setup.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	if len(args) == 0 {
+		return wait()
+	}
+	return run(cmds, args)
+}
+
+func startup(strategies map[string]SchedulerStrategy, strategyName string) error {
+	var keys []string
```
jasonbs10: Do this inside of the !ok?

scootapi/setup/setup.go
```diff
@@ -0,0 +1,57 @@
...
+// strategies maps strings to SchedulerStrategy
+// strategyName is the name of the strategy to use
+// args is the remaining arguments. These can be used to run a command
+func Main(cmds *Cmds, strategies map[string]SchedulerStrategy, strategyName string, args []string) error {
+	if err := startup(strategies, strategyName); err != nil {
+		return err
+	}
+
+	if len(args) == 0 {
+		return wait()
```
jasonbs10: I'd inline this or expand it to include main signal handling.

scootapi/setup/builder.go
```diff
@@ -0,0 +1,69 @@
...
+// NewGoBuilder creates a GoBuilder
+func NewGoBuilder(cmds *Cmds) *GoBuilder {
+	return &GoBuilder{
+		cmds: cmds,
+	}
+}
+
+// install runs go install and caches the results
+func (b *GoBuilder) install() {
+	if b.installed {
```
jasonbs10: It's fast enough not to matter, but it would be nice if this only compiled the minimum set of necessary binaries, or even just scheduler+worker.

scootapi/setup/builder.go
```diff
@@ -0,0 +1,69 @@
+package setup
```
dbentley: Yup; I agree. I thought about making there be an interface for just the command to run, but then felt like that was too much abstraction... WDYT? Is it worth unifying these?

(I'd tend to say "let's unify them if/when we see them drift apart")

scootapi/setup/builder.go
```diff
@@ -0,0 +1,69 @@
...
+	b.schedBin = path.Join(binDir, "scheduler")
+	b.workerBin = path.Join(binDir, "workerserver")
+}
+
+func (b *GoBuilder) Scheduler() (string, error) {
+	b.install()
+	return b.schedBin, b.err
+}
+
+func (b *GoBuilder) Worker() (string, error) {
```
dbentley: Will change.

jasonbs10:
(also, I agree with the weirdness of putting setup for workers in scootapi). Were you going to address that or would that involve scattering setup in multiple places?

scootapi/setup/workers.go
dbentley: It's not a strategyName, it's what gets passed to the scheduler per se, which is built-in to the scheduler (because of Caitie's config changes in the scheduler)

scootapi/setup/sched.go
dbentley: Collapsing these makes it easier to do the wrong thing... Using one as the other will lead to some kind of trouble, so why make it easier?

scootapi/setup/builder.go
jasonbs10: I'd like to see us go all in with code deduping since that was a large motivator for this PR.

dgassaway:
Is this waiting on anything?

PR #132, opened 2016-10-06
Un-latched stat.Counter, only latency/histogram should be latched.

jasonbs10:
Paired with go/phab/D9655


JeanetteBruno:
I'm still coming up to speed on these metrics/latching/non-latching concepts, but did notice the following:
stats_test.go, new version line 120 (a comment): change 'counter' to 'latency' 
stats.go, line 51 (a comment): the comment states that 'StatsRegistry does not support the Latency metric'... this doesn't appear to be correct.

common/stats/stats_test.go
```diff
@@ -94,16 +95,18 @@ func TestMarshal(t *testing.T) {
...
 func TestNonLatching(t *testing.T) {
 	stat := DefaultStatsReceiver().(*defaultStatsReceiver)
-	stat.Counter("counter").Inc(1)
+	stat.Latency("latency").Stop()
 
 	rendered := string(stat.Render(false))
-	if rendered != `{"counter":{"count":1}}` {
-		t.Fatal("Expected current stats in render", rendered)
+	matched, err := regexp.MatchString(`"count":1,`, rendered)
+	if !matched || err != nil {
```
dbentley: I suspect if err != nil, matched will always be false.

jasonbs10:
For the stats.go comment, I'm updating the comment to clarify that only our
finagleStatsRegistry supports latency while rcrowley's impl does not.

sched/scheduler/stateful_scheduler.go
```diff
@@ -238,11 +243,17 @@ func (s *statefulScheduler) scheduleTasks() {
...
 			func() error {
 				log.Println("Starting task", taskId, " command:", strings.Join(taskDef.Argv, " "))
-				return runTaskAndLog(
+				status, err := runTaskAndLog(
 					saga,
 					wf,
 					taskId,
 					taskDef)
+				if err == nil && status.State == runner.COMPLETE {
+					s.stat.Counter("schedCompletedTaskCounter").Inc(1)
```
CaitieM20: For code simplicity/maintainability we should log this closer to the code where the request is being made to the worker/runner.

Why not just log these in runTaskAndLog instead of propagating this information all the way back?

sched/scheduler/task_runner.go
```diff
@@ -10,23 +11,23 @@ import (
 // Run the task on the specified worker, and update the SagaLog appropriately.  Returns an error if one
 // occurs while running a task or writing to the SagaLog.  This method blocks until all saga messages
 // are logged and the task completes
-func runTaskAndLog(saga *saga.Saga, worker worker.Worker, taskId string, task sched.TaskDefinition) error {
+func runTaskAndLog(saga *saga.Saga, worker worker.Worker, taskId string, task sched.TaskDefinition) (runner.ProcessStatus, error) {
```
CaitieM20: suggest just passing the stat field in and logging stats in here. 

The intention is that this function handles everything associated with running a task.  Sending it to the worker, updating saga log, and IMO logging information about tasks.  

There's more granularity here about what went wrong we probably want to have a counter to Log when updating the Saga Fails to differentiate between when a Worker Fails etc... This is the info we'll need to debug the service in production

scootapi/server/serve.go
```diff
@@ -20,16 +24,45 @@ func MakeServer(handler scoot.CloudScoot,
...
+	stat               stats.StatsReceiver
+	runJobStatCount    int64
+	jobStatusStatCount int64
+}
+
+func (h *Handler) loop() {
+	tickerS := time.NewTicker(time.Second)
+	tickerM := time.NewTicker(time.Minute)
+	for {
+		// Manually latch and reset until we find a better way to do req/sec within viz's one minute window.
```
CaitieM20: Twitter's Observability Stack Doesn't support req/sec.  Is there a reason we want this other than nice to have? I suspect our req/sec and even req/min will fairly low initially.

If we do want req/sec I can show you the hack that Observability & Manhattan use which generates a stat for every second in the minute.  You end up with stats like
req_0
req_1
req_2
...
req_59

Then you still don't need latching the are just using counters and rates in Viz.  Implementing any kind of adhoc latching is going to make this difficult to reason about if most metrics aren't latched

scootapi/server/serve.go
```diff
@@ -20,16 +24,45 @@ func MakeServer(handler scoot.CloudScoot,
...
+			h.runJobStatCount = 0
+			h.jobStatusStatCount = 0
+		}
+	}
 }
 
 func (h *Handler) RunJob(def *scoot.JobDefinition) (*scoot.JobId, error) {
 	defer h.stat.Latency("runJobLatency_ms").Time().Stop()
-	return runJob(h.scheduler, def)
+	h.runJobStatCount++
```
CaitieM20: is this thread safe? you potentially have concurrent reads & writes of runJobStatCount in this method and in the loop?

Does this pass go test -race ./...

sched/scheduler/stateful_scheduler.go
jasonbs10: I think it's cleaner to have stats grouped together when possible rather than pass StatsReceiver everywhere. And it seems more reasonable to return a result from RunAndLog than not (I'm not a fan of void functions). That said, Dan and I were discussing moving more stats into Runner and further down the stack generally so we may revisit this.

sched/scheduler/task_runner.go
jasonbs10: Ok. I did want an overall failure rate but I suppose we can just add them up in our dashboard.

scootapi/server/serve.go
jasonbs10: Ok, I'll switch to RPM and we can revisit if that doesn't cut it.

scootapi/server/serve.go
jasonbs10: It does pass travis tests but you're right it's technically undefined behavior.

jasonbs10: Actually I think that we need to make sure all rpc calls are thread safe? We could have multiple RunJob() calls accessing the handler fields for instance. I wonder if the right thing is to either commit to using channels or AsyncRunner in our thrift handlers?

binaries/scheduler/main.go
```diff
@@ -5,17 +5,18 @@ package main
...
 	"github.com/scootdev/scoot/common/endpoints"
 	"github.com/scootdev/scoot/common/stats"
 	"github.com/scootdev/scoot/scootapi/server"
-	"log"
 )
 
 // Set Flags Needed by this Server
-var addr = flag.String("addr", "localhost:9090", "Bind address for api server.")
-var httpPort = flag.Int("http_port", 9091, "port to serve http on")
+var thriftAddr = flag.String("thrit_addr", "localhost:9090", "Bind address for api server.")
```
dbentley: thrift_addr?

PR #133, opened 2016-10-06
Mark property based tests with build tag

ryancouto:


Makefile
```diff
@@ -45,7 +45,7 @@ vet:
 
 test:
 	# Runs only unit tests
-	go test -race -tags=\!integration $$(go list ./... | grep -v /vendor/ | grep -v /cmd/)
+	go test -race -tags=\!integration\!property $$(go list ./... | grep -v /vendor/ | grep -v /cmd/)
```
jasonbs10: I'd use property_test or proptest to make this super clear in the test files that don't have _prop in the filename. Integration could be changed as well (but stands alone better than property).

CaitieM20: I would leave property based tests in here.  We want them to run as part of Travis CI.  This will make them no longer run

Makefile
```diff
@@ -45,7 +45,7 @@ vet:
...
 	sh testCoverage.sh
 
 test-integration:
```
CaitieM20: suggestion: add at test-unit and a test-property target for ease of use when developers only want to run one type locally

sched/job_def_serializer_test.go
```diff
@@ -1,3 +1,5 @@
+// +build property
```
CaitieM20: This file has property & unit tests in it.  adding this tag will make interpret all of them as property based tests. Can you split the property tests out into a separate file?

sched/scheduler/stateful_scheduler_test.go
```diff
@@ -1,3 +1,5 @@
+// +build property
```
CaitieM20: ditto this file has property and unit tests

scootapi/server/job_status_test.go
```diff
@@ -1,3 +1,5 @@
+// +build property
```
CaitieM20: both prop & unit tests in this file split out into two files for this tag to work properly

sched/job_def_serializer_test.go
jasonbs10: On second thought, what do we gain by distinguishing between unit/property tests? Anything at all?

Makefile
ryancouto: done

ryancouto: done

sched/job_def_serializer_test.go
ryancouto: reading https://jira.twitter.biz/browse/SCOOT-145, it seems there's some desire to be able to run unit tests without having to wait for property tests.

CaitieM20: yes this.  Sometimes property tests take longer to run, so you don't always want to run them when iterating on code

PR #134, opened 2016-10-10
Jeanette/scoot143 go1 7

JeanetteBruno:
Pull request for https://phabricator.twitter.biz/D9575

moved vendor libs that were embedded in stretchr/testify/vendor to scoots top level vendor dir.


dbentley:
Our other vendor'ed libs we've done using submodules, so they don't live in our repo per se. Do you mean to change that, or just the result of running a command?

JeanetteBruno:
I didn't realize that we used submodules.  I've removed the files from git and re-added them as submodules

.gitmodules
```diff
@@ -174,4 +174,13 @@
 [submodule "vendor/golang.org/x/text"]
 	path = vendor/golang.org/x/text
 	url = https://go.googlesource.com/text
-	branch = 1e65e9bf72c307081cea196f47ef37aed17eb316
\ No newline at end of file
+	branch = 1e65e9bf72c307081cea196f47ef37aed17eb316
+[submodule "vendor/github.com/stretchr/objx"]
+	path = vendor/github.com/stretchr/objx
+	url = https://github.com/stretchr/objx.git
```
jasonbs10: These should all specify the branch rev like above otherwise we're not really locking our dependencies in, submodule --init will pull in the latest.

PR #136, opened 2016-10-10
runner/local ensure checkout release in runner, cleanup

dgassaway:
- runner releases checkout resource if run cancelled before checkout is complete
- some cleanup wrt multiple uses of "run" in simple.go
- runner final updateStatus calls clearer, not relying on inner function logic
- test package rename per convention
- set stdout/stderr refs in run status sooner, and preserve them if status updated later

Main issue was IDd via code inspection - thinking about how to implement positive/negative tests to verify checkout.Release() calls

### DIFF
--- a/runner/local/simple.go
+++ b/runner/local/simple.go
@@ -107,10 +107,10 @@ func (r *simpleRunner) updateStatus(newStatus runner.ProcessStatus) (runner.Proc
 	}
 
 	if newStatus.State.IsDone() {
-		if oldStatus.StdoutRef != "" && newStatus.StdoutRef == "" {
+		if newStatus.StdoutRef == "" {
 			newStatus.StdoutRef = oldStatus.StdoutRef
 		}
-		if oldStatus.StderrRef != "" && newStatus.StderrRef == "" {
+		if newStatus.StderrRef == "" {
 			newStatus.StderrRef = oldStatus.StderrRef
 		}
 
@@ -140,7 +140,9 @@ func (r *simpleRunner) run(cmd *runner.Command, runId runner.RunId, doneCh chan 
 	case <-doneCh:
 		go func() {
 			<-checkoutDone
-			checkout.Release()
+			if checkout != nil {
+				checkout.Release()
+			}
 		}()
 		return
 	case <-checkoutDone:
--- a/runner/local/simple.go
+++ b/runner/local/simple.go
@@ -93,37 +93,37 @@ func (r *simpleRunner) Erase(runId runner.RunId) error {
 	return nil
 }
 
-func (r *simpleRunner) updateStatus(new runner.ProcessStatus) (runner.ProcessStatus, error) {
+func (r *simpleRunner) updateStatus(newStatus runner.ProcessStatus) (runner.ProcessStatus, error) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
-	old, ok := r.runs[new.RunId]
+	oldStatus, ok := r.runs[newStatus.RunId]
 	if !ok {
-		return runner.ProcessStatus{}, fmt.Errorf("cannot find run %v", new.RunId)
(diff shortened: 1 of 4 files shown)


runner/local/simple.go
```diff
@@ -88,47 +79,57 @@ func (r *simpleRunner) StatusAll() ([]runner.ProcessStatus, error) {
...
 		// We are ending the running task.
 		// depend on the invariant that there is at most 1 run with !state.IsDone(),
 		// so if we're changing a Process from not Done to Done it must be running
 		log.Printf("local.simpleRunner: run done. %+v", new)
 		close(r.running.doneCh)
 		r.running = nil
 	}
+
+	r.runs[new.RunId] = new
 	return new, nil
```
jasonbs10: As part of this cleanup could you rename 'new' to something else?

runner/local/simple.go
```diff
@@ -137,12 +138,16 @@ func (r *simpleRunner) run(cmd *runner.Command, runId runner.RunId, doneCh chan
 	// Wait for checkout or cancel
 	select {
 	case <-doneCh:
+		go func() {
```
jasonbs10: Now that I look at it, it seems like we would get a nil pointer exception if doneCh closes and 'defer checkout.Release()' gets called before checkout completes. The simplest solution would be to 'defer func() { if checkout != nil { checkout.Release() } }()

runner/local/simple.go
dgassaway: Yep, done!

runner/local/simple.go
dgassaway: I think I'm missing something :( As I'm reading it: if doneCh is closed, we start a goroutine that waits on checkoutDone and calls Release.  If doneCh is not closed, we just block until checkoutDone is closed and then defer the Release statement, so in both cases waiting for checkoutDone to be closed.

jasonbs10: Nope you're not missing anything, I misread.

runner/local/simple.go
```diff
@@ -88,47 +79,57 @@ func (r *simpleRunner) StatusAll() ([]runner.ProcessStatus, error) {
 	return statuses, nil
 }
 
-func (r *simpleRunner) Abort(run runner.RunId) (runner.ProcessStatus, error) {
-	return r.updateStatus(runner.AbortStatus(run))
+func (r *simpleRunner) Abort(runId runner.RunId) (runner.ProcessStatus, error) {
```
dbentley: Ugh, we've done fooId everywhere up til now, but I think Go style says to do fooID. Not sure if we should be consistent, or start doing the right thing and go back and change at some point... WDYT?

runner/local/simple.go
```diff
@@ -88,47 +79,57 @@ func (r *simpleRunner) StatusAll() ([]runner.ProcessStatus, error) {
...
-		return old, nil
+
+	if oldStatus.State.IsDone() {
+		return oldStatus, nil
 	}
-	r.runs[new.RunId] = new
-	if new.State.IsDone() {
+
+	if newStatus.State.IsDone() {
+		if oldStatus.StdoutRef != "" && newStatus.StdoutRef == "" {
```
dbentley: You can skip the oldStatus.StdoutRef != ""... same result, but it's less code...

runner/local/simple.go
dbentley: But what about if checkout is nil? Won't this panic?

runner/local/simple.go
dgassaway: Seems like the Go way is to prefer "ID" so I will go with that for files touched in this review.  If we want to revert files to sue that I can make a ticket, or we can start requesting changes to other reviews as it comes up.

runner/local/simple.go
dgassaway: Going to add some nil checks

runner/local/simple.go
dgassaway: will do

PR #139, opened 2016-10-18
[WorkerServer] Refactor Common Startup Code & Add Config

CaitieM20:
Config and WorkerServer Refactor to mimic [Scheduler](https://github.com/scootdev/scoot/commit/bafe23b1cc396762b297f2cd6233a04d82e9ce69)
- Added config file to ./binaries/workerserver/main.go, we don't have config now but will
- Refactored common startup code out of ./binaries/workerserver/main.go and put it into workerserver/setup.go
- Made WorkerServer use the Magic Bag to initialize services so that we don't have to pass http_addr and thrift_addr through as individual params

tested via 


workerapi/server/server.go
```diff
@@ -13,8 +13,24 @@ import (
...
+	return thrift.NewTSimpleServer4(
+		worker.NewWorkerProcessor(handler),
+		transport,
+		transportFactory,
+		protocolFactory)
+}
+
 // Called by a main binary. Blocks until the connection is terminated.
-func Serve(handler worker.Worker, addr string, transportFactory thrift.TTransportFactory, protocolFactory thrift.TProtocolFactory) error {
+func Serve(
```
jasonbs10: I'm missing it, where does Serve() get used?

workerapi/server/setup.go
```diff
@@ -0,0 +1,90 @@
...
+	schema := jsonconfig.Schema(make(map[string]jsonconfig.Implementations))
+	return bag, schema
+}
+
+// Starts the Server based on the MagicBag and config schema provided
+// this method blocks until the server completes running or an
+// exception occurs.
+func RunServer(
+	bag *ice.MagicBag,
+	schema jsonconfig.Schema,
```
jasonbs10: We need to parse the schema here otherwise it gets dropped on the floor.

workerapi/server/server.go
CaitieM20: it doesn't right now, but closed source does still, so deleting this now would cause a breaking change.

Until I merge this and change that code base I can't delete this yet.  There will be a follow up pull request removing this when all of the changes in Twitter Source are done.

workerapi/server/setup.go
CaitieM20: yup, added, I totally missed that, because config isn't being used yet, but once it is we don't want to have to change this again

workerapi/server/server.go
CaitieM20: its not right now, but removing this would cause a breaking change in Twitter Source.  

My plan is to merge this, then update Twitter Source.  Then submit a final OSS pull request removing this method.  That way we don't have to expertly time merges.

PR #140, opened 2016-10-18
Remove SwarmTest Binary

CaitieM20:
Removing the SwarmTest application from Binaries.  This is no longer the ideal way to launch this after @dbentley refactor.

Preferred way to run smoke test

Which I updated on the readme yesterday.  Deleting the old binary to avoid confusion in the future


jasonbs10:
I thought the preferred way (wrt replacing the original binary) was to pass "run ..." into setup-cloud-scoot? If we have to run two separate commands I'd hold off on deleting the original binary just yet.

JeanetteBruno:
also - swarmtest's setup_then_wait was convenient for setting up schedulers
and workers that are waiting for an ad-hoc command via scootapi.   Have
these features been ported to the new testing?  We'll also need to update
the quick reference and other instructions.

CaitieM20:
@JeanetteBruno The way @dbentley said we should be doing the setup_then_wait is just to run.
```
```
This does the same thing.  Then if you want to run a command or a smoke test against it you can.  I updated the Readme.  I don't think @dbentley changes are merged into Twitter Source yet? So I didn't update the go/scoot instructions.  Ideally @dbentley should do this when he checks in.  

I encountered this yesterday when trying to run swarmtest and the old instructions in the readme were broken, based on his changes.

dgassaway:
I think closed source has all of the changes, setup-cloud-scoot --strategy x.x works there

dbentley:
setup-cloud-scoot exists and should work in both repos

Yes, one command should suffice for both use cases.

Concatenating "run <argv>" to the command-line should run that argv.

PR #142, opened 2016-10-19
[workerserver] magic bag updates

CaitieM20:
Updating the default magic bag to include all defaults, main.go should override the ones it want's to use.  

tested via 

### DIFF
--- a/workerapi/server/setup.go
+++ b/workerapi/server/setup.go
@@ -5,14 +5,17 @@ import (
 
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/scootdev/scoot/common/endpoints"
+	"github.com/scootdev/scoot/common/stats"
 	"github.com/scootdev/scoot/config/jsonconfig"
 	"github.com/scootdev/scoot/ice"
+	"github.com/scootdev/scoot/os/temp"
 	"github.com/scootdev/scoot/runner"
 	"github.com/scootdev/scoot/runner/execer"
 	"github.com/scootdev/scoot/runner/execer/execers"
 	osexec "github.com/scootdev/scoot/runner/execer/os"
 	localrunner "github.com/scootdev/scoot/runner/local"
 	"github.com/scootdev/scoot/snapshot"
+	"github.com/scootdev/scoot/snapshot/snapshots"
 )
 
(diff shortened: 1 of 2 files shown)


jasonbs10:
I don't see anything that restricts us from overriding magic bag defaults... it looks like you can just call bag.Put(fn)? And if this doesn't work for some reason it seems like we'd want to make it work. Can you describe what's failing exactly?

CaitieM20:
Oh you are right, just tried it with tmp dir defined in both places it does appear to override.  I thought @dbentley had mentioned that you couldn't put things with the same type in the bag or it would cause problems.  I can update this, will still need a bit of a refactor to make runner.OutputCreateor and snapshot.Checkouter their own functions that can be overridden

workerapi/server/setup.go
```diff
@@ -43,19 +47,27 @@ func Defaults() (*ice.MagicBag, jsonconfig.Schema) {
 		endpoints.MakeStatsReceiver,
```
jasonbs10: Actually while we're here I'm realizing that I greatly prefer when we explicitly state the inputs/outputs for each func. Do you think it's too noisy to write functions that delegate to something with the same signature?

workerapi/server/setup.go
```diff
@@ -26,15 +28,17 @@ func makeServers(thrift thrift.TServer, http *endpoints.TwitterServer) servers {
 }
 
 // Creates the default MagicBag and JsonSchema for this Server and
-// returns them.  These can be modified before calling RunServer
-// Magic Bag does not include:
-// - thrift.TServerTransport
-// - *endpoints.TwitterServer
-// - *temp.TempDir
-// these should be added by callers before invoking RunServer
+// returns them.  These functions can be overriddne before calling RunServer
```
jasonbs10: spelling: overriddne

workerapi/server/setup.go
CaitieM20: nope i really like it too!  Makes it much more readable in my opinion,  I can update the rest of the functions to match!

workerapi/server/setup.go
CaitieM20: fixed

PR #144, opened 2016-10-19
Jeanette/scoot 199 daemon finding all runners busy

JeanetteBruno:
created api for scoot daemon and implemented Run() logic to return an error when all runners are busy.


snapshot/snapshots/fake_checkouter.go
```diff
@@ -15,6 +15,12 @@ type noopCheckouter struct {
 	path string
 }
 
+// have to have a valid path in the checkouter otherwise exec'ing a command
+// from simple.go doesn't work
+func MakeNoOpCheckouter() snapshot.Checkouter {
```
jasonbs10: I think TempCheckouter can be used for this case (and we probably don't want a Checkouter pointing to cwd).

daemon/server/local_scoot.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func NewRunnerManager(numRunners int, exec execer.Execer, checkouter snapshot.Checkouter, outputCreator runner.OutputCreator) *RunnerManager {
+	newRunnerManager := RunnerManager{}
+
+	newRunnerManager.runners = make([]runner.Runner, numRunners)
+	newRunnerManager.runIdToRunnerMap = make(map[runner.RunId]runner.Runner)
+
+	for i := 0; i < numRunners; i++ {
+		newRunnerManager.runners[i] = (runner.Runner)(local.NewSimpleRunner(exec, checkouter, outputCreator))
```
jasonbs10: You shouldn't need to cast this?

daemon/server/local_scoot.go
```diff
@@ -0,0 +1,56 @@
+package server
```
jasonbs10: I'd put RunManager somewhere in scoot/runner (and rename it) unless there's going to be more daemon specific stuff added.

daemon/server/local_scoot.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/scootdev/scoot/runner"
+	"github.com/scootdev/scoot/runner/execer"
+	"github.com/scootdev/scoot/runner/local"
+	"github.com/scootdev/scoot/snapshot"
+
+	"fmt"
+	"time"
+)
+
+const maxAccumulatedWaitTime time.Duration = 1 * time.Minute //TODO parameterize this
```
jasonbs10: Can you comment on this, I don't see where 'accumulated' comes into play.

daemon/server/local_scoot.go
```diff
@@ -0,0 +1,56 @@
...
+		runnerCommand := runner.NewCommand(cmd, env, maxAccumulatedWaitTime, snapshotId)
+
+		// start the command
+		status, err := runnerCandidate.Run(runnerCommand)
+
+		if err == nil {
+			// if it started successfully, stop looking
+			r.runIdToRunnerMap[status.RunId] = runnerCandidate
+			return status.RunId, nil
+		}
```
jasonbs10: You'll want an else here to return the relevant error rather than falling through to "no runners avail'

daemon/server/local_scoot.go
```diff
@@ -0,0 +1,56 @@
...
+	for i := 0; i < numRunners; i++ {
+		newRunnerManager.runners[i] = (runner.Runner)(local.NewSimpleRunner(exec, checkouter, outputCreator))
+	}
+
+	return &newRunnerManager
+}
+
+// find a runner that can run the requested command.  If all runners are currently processing commands
+// return an error, otherwise have the runner run the requested command.
+func (r *RunnerManager) Run(snapshotId string, cmd RunCommand, outputStrategy OutputStrategy) (runner.RunId, error) {
```
jasonbs10: My inclination would be to make this implement runner.Runner and perhaps fold OutputStrategy into runner.Command.

daemon/server/local_scoot_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/scootdev/scoot/snapshot/snapshots"
+
+	"strings"
+	"testing"
+)
+
+func TestNoRunners(t *testing.T) {
+
+	var exec execer.Execer = os.NewExecer()
+	var noopCheckouter snapshot.Checkouter = snapshots.MakeNoOpCheckouter()
```
jasonbs10: We've been preferring shorter names in general since that's golang convention (although I'm not a huge fan of single letter vars). Also ':=' should suffice here.

daemon/server/scootAPI.go
```diff
@@ -0,0 +1,16 @@
+package server
```
jasonbs10: This file doesn't appear to be necessary. We want the interface to be defined in daemon/protocol/daemon.proto (or thrift if we choose to switch), and to implement it using already existing components like runner.Runner (which will happen to be a RunnerManager in the daemon).

daemon/server/local_scoot.go
JeanetteBruno: the runner needs a timeout value with each command.  I'm also projecting that the daemon's Wait() feature will also have a timeout.  Both can use the same timeout.  I've updated the code to parameterize this with the setup.

daemon/server/local_scoot.go
JeanetteBruno: thx - fixed

daemon/server/local_scoot_test.go
JeanetteBruno: done - but would like to discuss the name length convention - shorter names do lead to more confusion.

daemon/server/local_scoot.go
jasonbs10: The command could fail immediately and return an exitCode (ex: if a binary doesn't exist). That command would never succeed and isn't really a "no runners" error.

PR #145, opened 2016-10-20
[Scheduler] Adding timeout to tasks

CaitieM20:
We are running into problems testing targets in source.  Some go off and never complete.  Adding a configurable timeout via the scheduler on tasks.  Once the timeout is reached the scheduler will mark the task as completed, but it will be marked with Status TIMEDOUT.  

* I also made the Polling Period Configurable while I was in that code.  
* fixed default config name in scheduler main.go since that never got updated with @dbentley changes.
* Renamed MakeRPCWorkerFactory to Create() to be consistent with the rest of the config objects

### DIFF
--- a/config/scootconfig/modules.go
+++ b/config/scootconfig/modules.go
@@ -45,20 +45,48 @@ func (c *ClusterLocalConfig) Create() (*cluster.Cluster, error) {
 }
 
 type WorkersThriftConfig struct {
-	Type string
+	Type               string
+	PollingPeriod      string // will be parsed to a time.Duration
+	EnforceTaskTimeout bool
+	TaskTimeout        string // will be parsed to a time.Duration
 }
 
-const pollingPeriod = time.Duration(250) * time.Millisecond
+const defaultPollingPeriod = time.Duration(250) * time.Millisecond
+const defaultTaskTimeout = time.Duration(30) * time.Minute
 
-func MakeRpcWorkerFactory(tf thrift.TTransportFactory, pf thrift.TProtocolFactory) worker.WorkerFactory {
-	return func(node cluster.Node) worker.Worker {
-		c := client.NewClient(tf, pf, string(node.Id()))
-		return workers.NewPollingWorker(c, pollingPeriod)
+func (c *WorkersThriftConfig) Create(
+	tf thrift.TTransportFactory,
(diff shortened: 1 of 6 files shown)


sched/worker/workers/polling_test.go
```diff
@@ -106,6 +106,53 @@ func TestPollingWorker_ErrorPolling(t *testing.T) {
...
+	ex := execers.NewSimExecer(&wg)
+	r := local.NewSimpleRunner(ex, snapshots.MakeInvalidCheckouter(), runners.NewNullOutputCreator())
+
+	w := NewPollingWorkerWithTimeout(
+		r,
+		time.Duration(10)*time.Microsecond,
+		false,
+		time.Duration(10)*time.Microsecond)
+
+	status, err := w.RunAndWait(task("sleep 50", "complete 43"))
```
dgassaway: Will this test take 50 seconds to run?

config/scootconfig/modules.go
```diff
@@ -45,20 +45,48 @@ func (c *ClusterLocalConfig) Create() (*cluster.Cluster, error) {
...
-		c := client.NewClient(tf, pf, string(node.Id()))
-		return workers.NewPollingWorker(c, pollingPeriod)
+	if c.TaskTimeout != "" {
+		tt, err = time.ParseDuration(c.TaskTimeout)
+		if err != nil {
+			return nil, err
+		}
 	}
+
+	wf := func(node cluster.Node) worker.Worker {
```
dgassaway: Maybe log here or in NewPollingWorker what the worker is being initialized with

sched/worker/workers/polling_test.go
CaitieM20: nope its milliseconds. 
Running go test -v ./sched/worker/workers shows it takes .5s
--- PASS: TestPollingWorker_TimeoutDisabled (0.05s)

config/scootconfig/modules.go
CaitieM20: So this function gets called every time a task is scheduled.  I'm worried that it'll be vary spammy to the logs.   Is there some specific information you are looking for we could add a metric?

dgassaway: Oh, based on the name I was assuming this was more related to workerserver startup, not task run. If we did log levels this might be a debug or trace log, but nbd for now.

PR #147, opened 2016-10-24
Updated daemon protocol to reflect the desired pants api.

jasonbs10:
Updated daemon protocol to reflect the desired pants api. 
Added an example test to exercise daemon as pants would initially.


daemon/protocol/daemon.proto
```diff
@@ -8,17 +8,20 @@ syntax = "proto3";
...
 
 service ScootDaemon {
   rpc Echo (EchoRequest) returns (EchoReply) {}
 
-  rpc Run(Command) returns (ProcessStatus) {}
-  rpc Status(StatusQuery) returns (ProcessStatus) {}
+  rpc CreateSnapshot(CreateSnapshotRequest) returns (CreateSnapshotReply) {}
+  rpc CheckoutSnapshot(CheckoutSnapshotRequest) returns (CheckoutSnapshotReply) {}
+  rpc Run(RunRequest) returns (RunReply) {}
+  rpc Poll(PollRequest) returns (PollReply) {}
```
JeanetteBruno: should this be 'Status' instead of 'Poll'?  (It doesn't wait till the runs reach a certain state does it?  Doesn't it return immediately with their current status?)

daemon/protocol/daemon.proto
```diff
@@ -27,31 +30,70 @@ message EchoReply {
...
+message PollReply {
+  enum State {
+    UNKNOWN = 0;
+    PENDING = 1;
+    RUNNING = 2;
+    COMPLETED = 3;
+    FAILED = 4;
+  }
+  message Result {
+    State state = 1;
```
JeanetteBruno: recommend putting run_id in this structure.

daemon/protocol/daemon.proto
```diff
@@ -27,31 +30,70 @@ message EchoReply {
...
+  string path = 1;
 }
 
-message Command {
-  repeated string argv = 1;
-  map<string, string> env = 2;
-  // Timeout in nanoseconds
-  int64 timeout = 3;
+message CreateSnapshotReply {
+  string error = 1;
```
dbentley: Does it include the created ID?

daemon/protocol/daemon.proto
```diff
@@ -27,31 +30,70 @@ message EchoReply {
...
-  ProcessState state = 2;
-  string stdout_ref = 3;
-  string stderr_ref = 4;
-  int32 exit_code = 5;
-  string error = 6;
+
+// Checkout snapshot.
+message CheckoutSnapshotRequest {
+  string snapshot_id = 1;
+  string path = 2;
```
dbentley: FWIW, I could go either way on this... Maybe the path should just be part of the reply? WDYT?

daemon/protocol/daemon.proto
```diff
@@ -27,31 +30,70 @@ message EchoReply {
...
+message RunRequest {
+  message Command {
+    repeated string argv = 1;
+    map<string, string> env = 2;
+    int64 timeout_ns = 3; //TODO: consistent special values for timeouts like PollRequest.timeout_ns?
+  }
+  message OutputPlan {
+    bool stdout = 1;
+    bool stderr = 2;
+    map<string, string> src_to_dest_paths = 3;
```
dbentley: Oh I haven't seen this. Maybe comment here?

daemon/protocol/daemon.proto
jasonbs10: This is a unix-style poll, described by PollRequest - <0 for indefinite, 0 for immediate, >0 for timeout.

daemon/protocol/daemon.proto
jasonbs10: It definitely needs to.

daemon/protocol/daemon.proto
jasonbs10: Unless we're planning to delete/gc checked out dirs (I don't think we want to), it's probably best to have the caller control this path.

daemon/protocol/daemon.proto
jasonbs10: Done.

daemon/protocol/daemon.proto
```diff
@@ -27,31 +31,86 @@ message EchoReply {
...
+  string snapshot_id = 1;
+  Command cmd = 2;
+  OutputPlan plan = 3;
+}
+
+message RunReply {
   string run_id = 1;
-}
\ No newline at end of file
+  string error = 2;
```
JeanetteBruno: LGTM 
-one thought: do we want to return a status in RunReply?  or will the status always either be 'error' or 'starting'...
can the status from Run() be 'running'/'preparing' - it is starting immediately,  'pending' - it is queued behind other run requests?

jasonbs10: It makes sense for PREPARING to be in there. For RunReply though, we only care if the scheduler accepted the run or not imo.

jasonbs10:
The latest change implements the daemon proto and plumbs it through the server code. If this looks ok I'll wrap up by updating the daemon client and adding tests.

daemon/protocol/daemon.proto
```diff
@@ -27,31 +31,93 @@ message EchoReply {
...
+  Command cmd = 2;
+  OutputPlan plan = 3;
 }
 
-message StatusQuery {
+message RunReply {
   string run_id = 1;
-}
\ No newline at end of file
+  string error = 2;
```
JeanetteBruno: I'm wondering if users will need to be able to get the status of the runs (regardless of whether any have finished),  and do we also need to provide 'wait till all are done'?

Poll can be extended to provide: 'get status', 'wait for at least one to finish', 'wait for all to finish' semantics if we add a 'selection' flag: AllRuns,  FinishedRuns

daemon/server/server.go
```diff
@@ -6,27 +6,35 @@ import (
...
+type Server interface {
+	ListenAndServe() error
+	Stop()
+}
+
+type server struct {
 	runner runner.Runner
+	filer  snapshot.Filer
 
 	server *grpc.Server
```
dgassaway: Can we rename some of these please, as this becomes difficult to read. package server, type server, type field server, etc

jasonbs10:
I was talking to Dan about confusing naming like this on Monday. It's a problem. If you have a contained set of specific suggestions I can do those here. Otherwise I suggest we cut a ticket and get fresh eyes on it for this upcoming sprint. 






On Wed, Oct 26, 2016 at 8:23 AM -0700, "Drew Gassaway" <notifications@github.com> wrote:












@dgassaway commented on this pull request.





In daemon/server/server.go:
>  	runner runner.Runner
+	filer  snapshot.Filer
 
 	server *grpc.Server



Can we rename some of these please, as this becomes difficult to read. package server, type server, type field server, etc

dgassaway:
Agree it should probably be a ticket also.  In this case I would suggest:
- leave "server" as package name
- for the struct server type, use simpleServer or similar (basicServer, serverStruct, whatever)
- for the *grpc.Server field in the struct use grpcServer (maybe gRPCServer is more golang-style?)

daemon/integration/integration_test.go
```diff
@@ -55,9 +56,8 @@ func TestMain(m *testing.M) {
 		log.Fatal("could not make server", err)
 	}
 
-	l, err := server.Listen()
 	go func() {
-		s.Serve(l)
+		s.ListenAndServe()
```
dbentley: I found this could lead to a race condition, because the test could start (and try to connect to the port) before this goroutine is able to listen. WDYT?

daemon/protocol/daemon.proto
```diff
@@ -27,31 +31,93 @@ message EchoReply {
...
+    // This copies the src files & dirs to the corresponding destination path=dir/base.
+    // The sources and destinations are all relative paths within the snapshot.
+    // An empty map will result in an empty snapshot (except for stdout/stderr).
+    // Behavior is undefined for duplicate entries within a destination.
+    // Sources may include the '*' wildcard. If they do, corresponding destinations are treated as parent dirs.
+    //
+    // Note: snapshots will always contain root STDOUT and STDERR text files.
+    map<string, string> src_paths_to_dest_dirs = 3;
+  }
+  string snapshot_id = 1;
```
dbentley: In runner.go, snapshot ID is part of command, but here it's separate. What's the right way to decide what goes in a command vs. a runrequest?

daemon/server/handler.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (h *Handler) Run(cmd *runner.Command) (status runner.ProcessStatus, err error) {
+	// Direct delegation to underlying runner.
+	return h.runner.Run(cmd)
+}
+
+func (h *Handler) Poll(runIds []runner.RunId, timeout time.Duration, returnAll bool) (statuses []runner.ProcessStatus) {
+	// Set up pollTicker to periodically query runner for status.
+	// Set up callerTimer to handle user-specified timeout.
+	pollTicker := time.NewTicker(h.pollInterval)
```
dbentley: Does this ever get closed?

daemon/server/handler.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	return err
+}
+
+func (h *Handler) Run(cmd *runner.Command) (status runner.ProcessStatus, err error) {
+	// Direct delegation to underlying runner.
+	return h.runner.Run(cmd)
+}
+
+func (h *Handler) Poll(runIds []runner.RunId, timeout time.Duration, returnAll bool) (statuses []runner.ProcessStatus) {
```
dbentley: Should this method (maybe not now, but eventually) just go on Runner? Or be a utility function?

It's been nice when our handlers are very little code other than translating and serializing/deserializing.

snapshot/filer.go
```diff
@@ -25,12 +25,19 @@ type Checkout interface {
...
 
 // Ingester creates a Snapshot from a path in the local filesystem.
 type Ingester interface {
-	// Takes a file or dir path and stores the contents for later Checkout by id.
+	// Takes a file or dir path and stores the contents in a snpashot which may then be checked out by id.
 	Ingest(path string) (id string, err error)
+
+	// Takes a mapping of source paths to be copied into corresponding destination directories.
+	// Source paths are absolute, and destination directories are relative to Checkout root.
+	IngestMap(srcToDest map[string]string) (id string, err error)
```
dbentley: This seems like a significant change to our interface. Is this blocking anything else? Do we need it right now? If not, it seems like the kind of change worth breaking out (so that we can keep changes small and review interface changes first)

snapshot/filer.go
```diff
@@ -25,12 +25,19 @@ type Checkout interface {
 	ID() string
 
 	// Releases this Checkout, allowing the Checkouter to clean/recycle this checkout.
-	// After Release(), the client may not look at files under Path()
+	// After Release(), the client may not look at files under Path().
 	Release() error
+
+	// Relocates the checkout contents to a caller controlled dir and then calls Release().
+	Disown(newAbsDir string) error
```
dbentley: See below

jasonbs10:
@dbentley I believe I addressed all of these as discussed except for IngestMany. I'm keeping that for now since it's the simplest/easiest way to take the stdout/stderr filepaths for now (and I commented out all the OutputPlan stuff).

binaries/daemon/main.go
```diff
@@ -32,13 +34,16 @@ func main() {
 	if err != nil {
 		log.Fatal("error creating temp dir: ", err)
 	}
+	defer os.RemoveAll(tempDir.Dir)
```
dbentley: This'll make it hard to debug; I guess I was trusting that temp gets reaped... WDYT?

daemon/server/handler.go
dbentley: ping

runner/local/simple.go
```diff
@@ -198,7 +197,24 @@ func (r *simpleRunner) run(cmd *runner.Command, runId runner.RunId, doneCh chan
 
 	switch st.State {
 	case execer.COMPLETE:
-		r.updateStatus(runner.CompleteStatus(runId, st.ExitCode))
+		snapshotId := ""
+		srcToDest := map[string]string{
```
dbentley: (Feel free to ignore; happy to do this in another PR)

Here's a way you could accomplish this without IngestMap (in pseudocode; excuse typos/wrong method names):

outputSnapDir := tempDir.NewTempDir()
ioutil.Cp(stdout.AsFile(), path.Join(outputSnapDir, "stdout"))
ioutil.Cp(stderr.AsFile(), path.Join(outputSnapDir, "stderr"))

snapID := r.filer.Ingest(outputSnapDir)

snapshot/filer.go
```diff
@@ -13,6 +13,9 @@ type Filer interface {
 type Checkouter interface {
 	// Checkout checks out the Snapshot identified by id, or an error if it fails.
 	Checkout(id string) (Checkout, error)
+
+	// Create checkout in a caller controlled dir.
+	CheckoutAt(id string, dir string) (Checkout, error)
```
dbentley: Sweet!

runner/local/simple.go
jasonbs10: We can address this in a new PR, and it's not super clean to require two copies of a file to ingest.

PR #148, opened 2016-10-25
scootapi,workerapi/client - client and dialer refactoring

dgassaway:
Refactor the api clients somewhat to support use of a dialer interface to manage thrift connection. Separate client from cli in workerapi, scootapi too commingled for now. This is primarily to support custom dialer implementations in closed source and to decouple client/cli/connection logic.

- separate workerapi client and cli client
- move dialer to common and use with both clients
- distinguish package, var, struct field names


binaries/scootapi/main.go
```diff
@@ -1,26 +1,26 @@
...
 import (
+	"log"
+
 	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/scootdev/scoot/common/dialer"
 	"github.com/scootdev/scoot/scootapi/client"
-	"log"
 )
 
-// Command line tool to talk to CloudScoot Server
```
jasonbs10: Can you replace these comments with something better rather than just delete?

dgassaway: Sure. Intent was preventing comments from being incomplete/out of date, but that is a premature optimization

scootapi/client/smoke_test_cmd.go
```diff
@@ -67,15 +67,15 @@ func (r *smokeTestRunner) run(numJobs int, timeout time.Duration) error {
...
 
 func (r *smokeTestRunner) generateAndStartJob() (string, error) {
-	client, err := r.cl.Dial()
+	err := r.cl.Dial()
 	if err != nil {
 		return "", err
 	}
 	rng := rand.New(rand.NewSource(time.Now().UnixNano()))
 
 	// We just want the JobDefinition here Id doesn't matter
```
dgassaway: yeah I don't really get this comment

PR #149, opened 2016-10-25
Jeanette/scoot199 queueing runner

JeanetteBruno:
Initial check in of QueueingRunner to kick-off reviewing the implementation approach.  I will be adding more tests.


runner/local/queueing_runner.go
```diff
@@ -0,0 +1,228 @@
...
+	// make a channel for the 'on queue' response
+	requestOnQueueCh := make(chan requestResponse)
+
+	// put the request on the process requests channel
+	requestContent := runRequestEvent{cmd: c, onQueueResponseCh: requestOnQueueCh}
+	qr.runRequestsCh <- requestContent
+
+	// get and return the 'on queue' response from this request's channel
+	requestQueued := <-requestOnQueueCh
+	close(requestOnQueueCh)
```
jasonbs10: Typically the sender closes channels, it doesn't particularly matter here though (and we could probably just leave this to be garbage collected automatically without closing?)

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,228 @@
...
+// status indicating if the request has been successfully queued or if the request has been denied (due to queue
+// length exceeded or some other error).
+
+// The status of a request is obtained with the run id.
+
+const QueueFullMsg = "No resources available. Please try later."
+const UnspportedFeatureMsg = "Not implemented.  This feature is not available for QueueingRunner"
+
+// The request queue entries are commandAndId structs
+type commandAndId struct {
```
jasonbs10: We can use runner.Command directly if we get rid of qRunId.

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,228 @@
...
+			qr.addRequestToQueue(nextRequest)
+
+		case <-qr.notifyRunnerAvailCh:
+			qr.runnerIsAvailable = true
+
+		case qRunId := <-qr.statusRequestsCh:
+			runnerRunId := qr.queueIdToRunnerIdMap[qRunId]
+			qr.getRunStatus(runner.RunId(runnerRunId), runner.RunId(qRunId))
+
+		default:
```
jasonbs10: This seems like it'll peg the cpu. We likely don't want a default case and can just move qr.RunNextCommandInQueue directly into case <-qr.notifyRunnerAvailCh.

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,228 @@
...
+
+	// map the queueing runner run ids to the runner's run ids. We need runner's run id for Status()
+	queueIdToRunnerIdMap map[string]string
+}
+
+// This must be used to initialize QueueingRunner properly
+func NewQueuingRunner(context context.Context,
+	runner runner.Runner,
+	maxQueueLen int,
+	runnerAvailableCh chan struct{}) runner.Runner {
```
jasonbs10: Btw my comments about taking vs returning a channel in simple.go don't apply here, but I wonder if we should make the runnerAvailableCh available directly from runner so we don't have to pass it around separately.

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,228 @@
...
+	qr.statusRequestsCh <- string(qRunId) // block next status request
+
+	statusResponse := <-qr.statusResponseCh // wait till get the status
+
+	return *statusResponse.status, statusResponse.err
+}
+
+// the event loop is triggering getting the status
+func (qr *QueueingRunner) getRunStatus(runnerRunId, qRunId runner.RunId) {
+	if strings.Compare("", string(runnerRunId)) == 0 {
```
JeanetteBruno: gone

runner/local/queueing_runner.go
JeanetteBruno: removed default case

runner/local/queueing_runner.go
JeanetteBruno: removed

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,228 @@
...
+		runner:               runner,
+		done:                 false,
+		maxQueueLen:          maxQueueLen,
+		nextRunId:            0,
+		runnerIsAvailable:    true,
+		notifyRunnerAvailCh:  runnerAvailableCh,
+		runRequestsCh:        make(chan runRequestEvent),
+		statusRequestsCh:     make(chan string),
+		statusResponseCh:     make(chan requestResponse),
+		queueIdToRunnerIdMap: make(map[string]string),
```
JeanetteBruno: see comment above about queueingRunner becoming PoolRunner

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,228 @@
...
+	runner runner.Runner,
+	maxQueueLen int,
+	runnerAvailableCh chan struct{}) runner.Runner {
+
+	qRunner := &QueueingRunner{
+		runner:               runner,
+		done:                 false,
+		maxQueueLen:          maxQueueLen,
+		nextRunId:            0,
+		runnerIsAvailable:    true,
```
JeanetteBruno: removed done

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,228 @@
...
+func NewQueuingRunner(context context.Context,
+	runner runner.Runner,
+	maxQueueLen int,
+	runnerAvailableCh chan struct{}) runner.Runner {
+
+	qRunner := &QueueingRunner{
+		runner:               runner,
+		done:                 false,
+		maxQueueLen:          maxQueueLen,
+		nextRunId:            0,
```
JeanetteBruno: we should talk about this... IMHO QueueingRunner should become PoolRunner.  we should not have queueing in runners that really run the commands.

runner/local/queueing_runner.go
JeanetteBruno: see prior comment on runnerAvaiableCh - we should have a ftf conversation about this.

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,228 @@
...
+	maxQueueLen       int
+	nextRunId         int // used to assign unique runIds to the run requests
+	runQueue          []commandAndId
+	runnerIsAvailable bool
+
+	notifyRunnerAvailCh chan struct{} // the runner will use this channel to signal it is available
+
+	runRequestsCh chan runRequestEvent // channel for synchronizing calls to Run()
+
+	statusRequestsCh chan string          // channel for synchronizing status requests
```
JeanetteBruno: done - was able to just make it runner.RunId once I renamed the runner field to aRunner.

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,228 @@
...
+type requestResponse struct {
+	status *runner.ProcessStatus
+	err    error
+}
+
+//TODO - do we really need a new channel for each run request?
+// This struct to associates a response channel to each Run() request
+type runRequestEvent struct {
+	cmd               *runner.Command
+	onQueueResponseCh chan requestResponse // chan to use to indicate that the request is queued
```
JeanetteBruno: I was trying to have the channel name reflect it's purpose. (My natural inclination for more expressive names ex: lockRunRequests, unlockRunRequests - but steered away from that since the it seems shorter names are preferred.)  naming the channel commandCh doesn't convey that it's blocking the next run request.

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,228 @@
...
+
+// requestResponse structs are returned in the Run() and Status() response channels
+type requestResponse struct {
+	status *runner.ProcessStatus
+	err    error
+}
+
+//TODO - do we really need a new channel for each run request?
+// This struct to associates a response channel to each Run() request
+type runRequestEvent struct {
```
JeanetteBruno: done

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,228 @@
...
+	cmd *runner.Command
+}
+
+// requestResponse structs are returned in the Run() and Status() response channels
+type requestResponse struct {
+	status *runner.ProcessStatus
+	err    error
+}
+
+//TODO - do we really need a new channel for each run request?
```
JeanetteBruno: Great! done (wasn't sure if I was missing something)

runner/local/simple.go
```diff
@@ -195,10 +209,18 @@ func (r *simpleRunner) run(cmd *runner.Command, runId runner.RunId, doneCh chan
...
 	case execer.FAILED:
-		r.updateStatus(runner.ErrorStatus(runId, fmt.Errorf("error execing: %v", st.Error)))
+		r.updateStatus(runner.ErrorStatus(runId, fmt.Errorf("error execing: %v", st.Error)), r.runnerAvailableCh)
 	default:
-		r.updateStatus(runner.ErrorStatus(runId, fmt.Errorf("unexpected exec state: %v", st.State)))
+		r.updateStatus(runner.ErrorStatus(runId, fmt.Errorf("unexpected exec state: %v", st.State)), r.runnerAvailableCh)
+	}
+}
+
+func reportRunnerAvailable(runnerAvailableCh chan struct{}) {
```
JeanetteBruno: Glad you mentioned this - I forgot a //TODO message:  should reportRunnerAvailable() be called when updateStatus() doesn't find the runId (line113) - if yes, then I would leave reportRunnerAvailable() in place (it hits my 'don't copy >=3 loc rule)

runner/local/simple.go
```diff
@@ -93,7 +105,7 @@ func (r *simpleRunner) Erase(runId runner.RunId) error {
 	return nil
 }
 
-func (r *simpleRunner) updateStatus(newStatus runner.ProcessStatus) (runner.ProcessStatus, error) {
+func (r *simpleRunner) updateStatus(newStatus runner.ProcessStatus, runnerAvailableCh chan struct{}) (runner.ProcessStatus, error) {
```
JeanetteBruno: done - thx for catching this

runner/local/simple.go
```diff
@@ -13,22 +13,34 @@ import (
 
 func NewSimpleRunner(exec execer.Execer, checkouter snapshot.Checkouter, outputCreator runner.OutputCreator) runner.Runner {
```
JeanetteBruno: My natural inclination was that the caller took ownership of the channel: 'I need a runner and want it to report back to me when it is available'... this whole reporting back channel design has the following weaknesses: 
-the runner is passed in to queueingRunner and queueingRunner assumes that the runner initially is avaialable
- also, the isAvailable concept breaks down if/when the runner becomes a service - possibly shared by multiple callers (do we ever envision this architecture?).  

A stronger solution (IMHO) would be to simply send the Run() request and let the runner reject it if it is not available.  I think I started out with this approach but after conversations with Dan and in my eagerness to try using channels as mutex I may have gone overboard!  I think going back to this design (Run() rejects when not available - which is what runner does already) would actually simplify the QueueingRunner code.

jasonbs10: As per our discussion, let's leave this as is unless we can find a solution without any drawbacks.

runner/local/simple.go
jasonbs10: In that case, yeah, we could keep this function and then handle line 113. Alternatively, we could make this an anonymous function within UpdateStatus.

runner/local/queueing_runner.go
jasonbs10: My mistake, QueueRunner needs to generate its own runIds so it can return a valid run handle without actually running the command.

JeanetteBruno:
replaced by pull request 153

PR #151, opened 2016-10-31
python lib

jasonbs10:
Tested python daemon client lib.


daemon/protocol/locate.go
```diff
@@ -7,6 +7,7 @@ import (
 )
 
 // Utilities to deal with Scoot instances.
+// Replicated in # Copied from daemon/protocol/client_lib.py
```
dbentley: typo?

daemon/protocol/README.md
```diff
@@ -1,3 +1,16 @@
 The protocol betweeen Scoot Local Server and Scoot Local Clients.
 
-This includes the protocol buffers, and any other utilities that should be shared by client and server.
\ No newline at end of file
+This includes the protocol buffers, and any other utilities that should be shared by client and server.
+
+// TODO(dbentley): integrate this into our go build
```
dbentley: Copied comment now out of date?

daemon/protocol/client_lib.py
```diff
@@ -0,0 +1,228 @@
...
+        self.exit_code = exit_code
+        self.error = error
+
+
+def start():
+    """
+    """
+    global _client, _domain_sock
+    if _domain_sock is None:
+        raise ScootExceptoin("Could not find domain socket path.")
```
dbentley: Typo?

jasonbs10:
@dbentley, I'd like feedback about whether this looks as you expected, if we want to start automating the code generation, where the example python client should live and if there's anything wrong with calling it from a go test script (currently in __main__ right now), and any thoughts on how best to create a pex for distribution.

dbentley:
This looks a lot like I expected.

Is there generated code now? I thought grpc gave us a lot of code generation, but not the "how to find the socket" code.

Where it should live: I'm fine with here, but a new repo might make sense too.

Calling it from a go test script sounds great.

Pex for distribution: I thought that Pex was for binaries, and if we want this as a library then it should maybe be in pypi, not a pex? But I don't know about that, and I bet Kris Wilson would help.

daemon/protocol/client_lib.py
```diff
@@ -0,0 +1,228 @@
...
+    return None
+
+def _locate_socket():
+    scootdir = _locate_scoot_dir()
+    if scootdir is None:
+        return None
+    return os.path.join(scootdir, "socket")
+
+
+# Singleton variables.
```
dgassaway: Prefer an implementation of the client as a class, where these vars are data members and the start/stop/run/etc functions are methods.  This will let us ditch the global/singleton pattern, and be able to use `__init__` to simplify the "if is_started" checks in all of the functions. This should make the client here super easy to use as a lib or call from separate test files, although I'm not sure what our desired lib vs reference implementation separation is. (Maybe all of the daemon_pb2 calls can be factored out into non-class functions that will eventually make up the library portion in their own file).  I think this would be more flexible either way, WDYT?

daemon/protocol/client_lib.py
```diff
@@ -0,0 +1,228 @@
...
+def checkout_snapshot(snapshot_id, dirpath):
+    """
+    """
+    global _client
+    if not is_started():
+        raise ScootExceptoin("Not started.")
+    req = daemon_pb2.CheckoutSnapshotRequest(snapshot_id=snapshot_id, dir=dirpath)
+    try: resp = _client.CheckoutSnapshot(req)
+    except Exception as e: raise ScootException(e)
+    if resp.error != "":
```
dgassaway: ^ same

daemon/protocol/client_lib.py
```diff
@@ -0,0 +1,228 @@
...
+def create_snapshot(path):
+    """
+    """
+    global _client, _context
+    if not is_started():
+        raise ScootException("Not started.")
+    req = daemon_pb2.CreateSnapshotRequest(path=path)
+    try: resp = _client.CreateSnapshot(req, _context)
+    except Exception as e: raise #ScootException(e)
+    if resp.error != "":
```
dgassaway: "if resp.error:" should work

daemon/protocol/client_lib.py
```diff
@@ -0,0 +1,228 @@
...
+        ok_id = create_snapshot(os.path.join(tmpdir, "*"))
+        fail_id = create_snapshot(script)
+
+	# Run scripts serially in their respective snapshots. Block until each run finishes.
+	ok_run_id = run(argv=["./script.sh"], timeout_ns=500*1e6, snapshotId=ok_id)
+        ok_statuses = poll(run_ids=[ok_run_id], timeout_ns=500*1e6)
+        if len(ok_statuses) != 1:
+            raise ScootException("expected one poll result for ok_run_id.")
+
+	fail_run_id = run(argv=["./script.sh"], timeout_ns=500*1e6, snapshotId=fail_id)
```
kwlzn: +1 - it'd be great to use 2 space indent too (vs 4, currently) to match Twitter/Pants/Google/etc style.

daemon/protocol/client_lib.py
```diff
@@ -0,0 +1,228 @@
...
+
+# Singleton variables.
+_domain_sock = _locate_socket()
+_client = None
+_context = None
+
+class ScootException(Exception):
+    """
+    """
+    def __init__(self, caught):
```
kwlzn: it'd probably make sense to implement `__repr__` and `__str__` for this as well, so that the repr'd/string representation of the exception clearly shows the wrapped exception. e.g.

```
>>> class ScootException(Exception):
...   def __init__(self, caught):
...     self.caught = caught
... 
>>> e = ScootException(Exception('blah'))
>>> e
ScootException()
```

vs

```
>>> class ScootException(Exception):
...   def __init__(self, caught):
...     self.caught = caught
...   def __repr__(self):
...     return '{}({!r})'.format(self.__class__.__name__, self.caught)
...   
... 
>>> e = ScootException(Exception('blah'))
>>> e
ScootException(Exception('blah',))
```

and `__str__` could just return `repr(self)` - or you could also call `super(Exception, self).__init__(...)` to define the string representation using the base `Exception` mechanics.

daemon/protocol/client_lib.py
```diff
@@ -0,0 +1,228 @@
...
+        open(script, 'w').write("#!/bin/sh\nls resource.txt")
+        open(resource, 'w').write("content")
+
+	# Ingest scripts into their own snapshots. The 'fail' snapshot will be missing resource.txt.
+        ok_id = create_snapshot(os.path.join(tmpdir, "*"))
+        fail_id = create_snapshot(script)
+
+	# Run scripts serially in their respective snapshots. Block until each run finishes.
+	ok_run_id = run(argv=["./script.sh"], timeout_ns=500*1e6, snapshotId=ok_id)
+        ok_statuses = poll(run_ids=[ok_run_id], timeout_ns=500*1e6)
```
kwlzn: seems like `timeout_ns` could stand to be a constant?

daemon/protocol/client_lib.py
```diff
@@ -0,0 +1,228 @@
...
+    def domain(status):
+        return ScootStatus(run_id=status.run_id,
+                           state=status.state,
+                           snapshot_id=status.snapshot_id,
+                           exit_code=status.exit_code,
+                           error=status.error)
+    return [domain(x) for x in resp.status]
+
+
+if __name__ == '__main__':
```
kwlzn: at a minimum, it'd be good to move this into a function defn (which at least nets a function scope vs global) and then do:

```
if __name__ == '__main__':
  run_test()
```

daemon/protocol/client_lib.py
kwlzn: here and elsewhere, should this actually be:

`raise ScootException(Exception("Could not find domain socket path"))`

given that `ScootException` expects to wrap an exception instance vs a string?

daemon/protocol/client_lib.py
```diff
@@ -0,0 +1,228 @@
...
+
+def create_snapshot(path):
+    """
+    """
+    global _client, _context
+    if not is_started():
+        raise ScootException("Not started.")
+    req = daemon_pb2.CreateSnapshotRequest(path=path)
+    try: resp = _client.CreateSnapshot(req, _context)
+    except Exception as e: raise #ScootException(e)
```
kwlzn: seems like this should also wrap with `ScootException`? maybe this was commented out for debugging?

daemon/protocol/client_lib.py
dgassaway: Any reason we want any exception class logic at all over:
`class ScootException(Exception):
    pass`
?

.travis.yml
```diff
@@ -1,19 +1,26 @@
...
-  - make format
-  - make vet
+#before_script:
+#  - make format
+#  - make vet
 
 script:
-  - make test
+  - pip install grpcio
+  - echo -e "import sys\nprint(sys.path)\nimport grpc" | python
```
dbentley: Are these changes needed?

daemon/integration/python/integration_py_test.go
```diff
@@ -0,0 +1,73 @@
+package integration_test_py
```
dbentley: package nam mismatch?

dgassaway:
I like the test refactor.  I still strongly believe this should be made as a class for code quality and to avoid using the global paradigm to enforce the singleton client concept, but I won't block it for that if we're fine moving forward.

PR #152, opened 2016-10-31
Caitie/max retry task

CaitieM20:
Adding a configurable value to the scheduler MaxRetryTasks.  This value is the maximum number of times a task will be retried before being marked as completed.  This prevents a bad task from taking down the entire cluster.  Total times tried is 1 + MaxRetryTasks.

* Once a Task is Retried maximum number of times and has still failed the Task is marked as completed and the error is recorded in the Saga.  This allows the rest of the Job to progress and try to execute.  
* Add SchedulerConfig to DI/config system to ease the pain & reduce the amount of code changes necessary for adding config values in the future.  
* There is a separate SchedulerConfig object in the Scheduler package to avoid coupling our config implementation to the Scheduler implementation.  This requires some extra code, but I think is the right separation of concerns in the future.  
* additional unit test cases added
* verified by running swarmtest on a local cluster.


sched/scheduler/task_runner.go
```diff
@@ -1,17 +1,34 @@
...
+	"log"
+
 	"github.com/scootdev/scoot/common/stats"
 	"github.com/scootdev/scoot/saga"
 	"github.com/scootdev/scoot/sched"
 	"github.com/scootdev/scoot/sched/worker"
 	"github.com/scootdev/scoot/workerapi"
 )
 
+const DeadLetterExitCode = -100
```
dbentley: I think Unix exit codes are -128 to 127, so I might pick something like -200 just to be safely outside one byte?

sched/scheduler/stateful_scheduler.go
```diff
@@ -240,6 +248,8 @@ func (s *statefulScheduler) scheduleTasks() {
 		jobState := s.inProgressJobs[jobId]
 		nodeId := ta.node.Id()
 
+		markCompleteOnFailure := bool(ta.task.NumTimesTried >= s.maxRetriesPerTask)
```
CaitieM20: yup i think that's a better description changed

PR #153, opened 2016-11-01
runner/local: QueueingRunner queues runs

dbentley:
QueueingRunner is a runner that queues runs (up to a capacity) and delegates runs to a delegate.

This isn't yet wired up into any server, but should let us make the RunMany test case pass.

rename simExecer to SimExecer and replace injected WaitGroup with a Resume method

(Based on lots of work by Jeanette)


runner/execer/execers/sim.go
```diff
@@ -11,24 +11,28 @@ import (
...
 
-// simExecer execs by simulating running argv.
+// SimExecer execs by simulating running argv.
 // each arg in command.argv is simulated in order.
 // valid args are:
 // complete <exitcode int>
 //   complete with exitcode
 // pause
-//   pause until the simExecer's WaitGroup is Done
+//   pause until the SimExecer's WaitGroup is Done
```
JeanetteBruno: I think this comment needs to be changed?  (// pause until the SimExecer's Abort() or Resume() is called.)?

runner/execer/execers/sim.go
```diff
@@ -151,11 +159,19 @@ func (s *completeStep) run(status execer.ProcessStatus, p *simProcess) execer.Pr
...
 type pauseStep struct {
-	wait *sync.WaitGroup
+	ch chan struct{}
 }
 
 func (s *pauseStep) run(status execer.ProcessStatus, p *simProcess) execer.ProcessStatus {
-	s.wait.Wait()
+	waitCh := make(chan struct{})
+	go func() {
+		p.Wait()
```
JeanetteBruno: a comment here would be helpful: 
p.Wait() // pauses until Abort() is called

runner/execer/execers/sim.go
```diff
@@ -151,11 +159,19 @@ func (s *completeStep) run(status execer.ProcessStatus, p *simProcess) execer.Pr
...
+		close(waitCh)
+	}()
+	select {
+	case <-waitCh:
```
JeanetteBruno: again comments:
case <-waitCh:  // is triggered by calling Abort(), which causes Wait() to close waitCh
case <-s.ch: // is triggered by calling Resume()

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,339 @@
...
+	// qToDel maps IDs in our namespace to IDs in the delegate's namespace
+	// delToQ is vice versa
+	qToDel map[runner.RunId]runner.RunId
+	delToQ map[runner.RunId]runner.RunId
+}
+
+// NewQueueingRunner creates a new QueueingRunner.
+// delegate will be used to run.
+// capacity is the maximum entries to hold in the queue
+// runnerAvailableCh should send a struct whenever the runner is available
```
JeanetteBruno: change to 
// the delegate runner should send a struct{}{} on runnerAvailableCh when the delegate runner becomes available

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,339 @@
...
+	// start the request event processor
+	go qRunner.eventLoop()
+
+	return qRunner
+}
+
+// Implement Runner's methods by sending a request to qr.reqCh and waiting
+
+func (qr *QueueingRunner) Run(cmd *runner.Command) (runner.ProcessStatus, error) {
+	statusCh, errCh := make(chan runner.ProcessStatus), make(chan error)
```
JeanetteBruno: Do we need to make a new set of channels for each request? If we had one channel per command (Run(), Status(), StatusAll(), etc.) would that also work?   Am I missing an implementation concept that requires a new channel for each request?

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,339 @@
...
+		reqCh: make(chan interface{}),
+
+		delegate:  delegate,
+		capacity:  capacity,
+		nextRunId: 0,
+
+		runnerAvail: true,
+
+		q:       nil,
+		qToDel:  make(map[runner.RunId]runner.RunId),
```
JeanetteBruno: This is not a big deal, but my mind tends to stutter over the 'del' abbreviation - 'delete' is the first expansion that comes to my mind, then I have to remember this use is 'delegate' not 'delete'.

runner/local/queueing_runner_test.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+}
+
+func TestUnknownRunIdInStatusRequest(t *testing.T) {
+	env := setup(4, t)
+	defer env.teardown()
+	qr := env.qr
+
+	st, err := qr.Status(runner.RunId("not a real run id"))
+	if err == nil {
```
JeanetteBruno: do we want to verify that the error message was due to not finding the run id? (add: '|| !strings.Contains(err.Error(), runner.UnknownRunIdMsg')

runner/local/simple_test.go
```diff
@@ -112,6 +114,10 @@ func aborted() runner.ProcessStatus {
 	return runner.AbortStatus(runner.RunId(""))
 }
 
+func badRequest(errorText string) runner.ProcessStatus {
+	return runner.BadRequestStatus(runner.RunId(""), fmt.Errorf(errorText))
+}
+
 func assertRun(t *testing.T, r runner.Runner, expected runner.ProcessStatus, args ...string) runner.RunId {
```
JeanetteBruno: I've found it useful for commonly used test assertion routines to have a 'tag' parameter that the test can set to better identify which use of the assertion is error-ing.

runner/execer/execers/sim.go
dgassaway: Can we also add a comment here about how to invoke the SimExecer (the "#! simexec" or whatever)

dgassaway:
Does this supersede #149 ?

JeanetteBruno:
yes - this pull request replaces 149

runner/execer/execers/sim.go
dbentley: Done.

runner/execer/execers/sim.go
dbentley: Done.

runner/local/queueing_runner.go
dbentley: Done.

runner/local/queueing_runner.go
dbentley: Discussed offline.

We don't need to per se, but I think it's easier to reason about so we don't have to convince ourselves.

I call this the "self-addressed, stamped envelope" pattern. A request includes the destination.

runner/local/queueing_runner_test.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	// resume
+	env.sim.Resume()
+
+	// drain
+	for _, id := range runIDs {
+		assertWait(t, qr, id, complete(0), "n/a")
+	}
+	runIDs = nil
```
dbentley: Done.

runner/local/queueing_runner_test.go
dbentley: Done.

runner/local/simple_test.go
dbentley: Ooh; I like that pattern.

FWIW, since writing this code I've heard that you're supposed to only t.Fatal on the outermost level.... so instead of assert, we should test and then if err != nil t.Fatal... but that's so much ode.

Would it be nicer to include the full stack trace? That's at least populated automatically?

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,339 @@
...
+func (qr *QueueingRunner) statusAll() ([]runner.ProcessStatus, error) {
+	r, err := qr.delegate.StatusAll()
+	if err != nil {
+		return nil, err
+	}
+
+	// translate the RunID from the delegate's namespace to ours
+	// (this is why we need the reverse mapping as well)
+	for i, st := range r {
+		qID, ok := qr.delToQ[st.RunId]
```
jasonbs10: This block right here, I think it's officially too many short var names.

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,339 @@
...
+		runnerAvail: true,
+
+		q:       nil,
+		qToDel:  make(map[runner.RunId]runner.RunId),
+		delToQ:  make(map[runner.RunId]runner.RunId),
+		errored: make(map[runner.RunId]runner.ProcessStatus),
+	}
+
+	// push runner available notifications into our request channel
+	go func() {
```
jasonbs10: It's a little awkward to have this apart from the event loop.

runner/local/simple_test.go
JeanetteBruno: Let's move this conversation to our Testing Best Practices page (https://confluence.twitter.biz/display/EE/Testing+Best+Practices)

runner/local/queueing_runner.go
```diff
@@ -0,0 +1,339 @@
...
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/scootdev/scoot/runner"
+)
+
+// TODO(dbentley): should this be in queue.go
+// TODO(dbentley): should this whole package move to runner/runners?
```
dbentley: OK; I'll do that in a new PR.

runner/local/queueing_runner.go
dbentley: Yes; I could fix it, but it would make the event loop look weird (because now it has to read from two channels, not just one). WDYT?

runner/local/queueing_runner.go
dbentley: Explicated; how's that?

PR #154, opened 2016-11-01
Scoot200 - created 2 integration tests and fixed a couple of issues w…

JeanetteBruno:
…ith returning correct status values.

### DIFF
--- a/daemon/integration/integration_test.go
+++ b/daemon/integration/integration_test.go
@@ -2,15 +2,23 @@ package integration_test
 
 import (
 	"flag"
+	"io/ioutil"
+	"os"
+	"path"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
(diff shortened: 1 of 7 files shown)


daemon/integration/integration_test.go
```diff
@@ -25,6 +33,63 @@ func TestEcho(t *testing.T) {
...
+}
+
+// send 2 run commands, where the first one sleeps for a short while to block the second one
+// expected results: first request runs, second request returns an error
+func TestRun2Commands(t *testing.T) {
+	// run the first command
+	var statusReq []string = []string{"run", "sleep 10"}
+	runId := assertRun("ignore", "running", "nil", statusReq[:], t)
+
+	// run the second command, it should get runner busy message
```
dbentley: I'd also add a TODO of "we want this to enqueue and then succeed"

JeanetteBruno: done

PR #155, opened 2016-11-01
snapshot/git/gitfiler - add capacity option for repo pools, update travis-ci

dgassaway:
- Added a capacity spec for repo pools and used for Cloning Checkouter. Allows for limit on creation of unnecessary expensive repos (e.g. git clone source)
- Make daemon integration test timings safer
- Update travis-ci golang version to 1.7, fix Readme widget
Notes
1. daemon integration test failed in master and in some branches. Hard to reproduce, suspect race cond.
2. Repopool w/ cloning checkouter of monorepo would create additional clone that could not be used, at cost of ~3GB and i/o contention


snapshot/git/gitfiler/pool.go
```diff
@@ -80,6 +92,7 @@ func (p *RepoPool) loop() {
 			p.freeList = append(p.freeList, r)
```
dbentley: should the size increase here, too?

dgassaway: This is strange wrt size vs freeList - size is intended to indicate "allocated" repos, vs "reserved" repos.  The double use of "Get()" is confusing here also.  In the cloner case we first Get() the clone, which is more of an expensive allocation operation. Then the user of the pool Get()s the repo, which is more like saying "hey I'm going to use that one."  So I this introduces a repo allocation vs reservation/use issue.

To be clear, size should only increase when: we initialize the pool with existing repos, or in the pool loop we run out of free repos, have spare capacity, and call RepoGetter.Get().

jasonbs10: I would find that update helpful. Wrt to the confusing 'size', I might prefer initited or numInited.

daemon/integration/integration_test.go
```diff
@@ -39,7 +41,7 @@ func TestRunSimpleCommand(t *testing.T) {
 	runId := assertRun("ignore", "running", "nil", statusReq[:], t)
 
 	// give the command time to finish running
-	time.Sleep(30 * time.Millisecond)
+	time.Sleep(100 * time.Millisecond)
```
jasonbs10: These changes seem completely unrelated to the other changes? Was this failing for you/travis-ci?

dgassaway: Yes, I had the TestRun2Commands test fail randomly a couple of times locally and saw some failed builds in travis due to it.  I'm just increasing the time spacing on these for now to make them very unlikely.

PR #156, opened 2016-11-02
runner: create Subscriber interface

dbentley:
Subscriber is a new interface to replace polling (and especially polling that has to just call StatusAll)

Eventually, this should be folded into Runner. It might even (eventually) replace Erase() and Clear() because the Semantics of Subscribing are clearer.

This is just the interface for now (as we discussed)

### DIFF
--- a/runner/subscriber.go
+++ b/runner/subscriber.go
@@ -1,0 +1,81 @@
+package runner
+
+// SubscriptionID identifies a subscription as an opaque string.
+type SubscriptionID string
+
+// SubscriptionDefinition defines what is watched by a Subscription.
+type SubscriptionDefinition struct {
+	// IDs of Runs to watch, or empty to watch all
+	Runs []RunId
+	// States to watch. To watch state n, set the nth bit in StateMask.
+	StateMask int
+}
+
+// An Event is an ID and the status that was recorded.
+type Event struct {
+	ID     EventID
+	Status ProcessStatus
+}
+
+// An EventID identifies an Event; it's Subscription and a Sequence number.
+type EventID struct {
+	Sub SubscriptionID
+	Seq int
+}
+
+// PollOpts describes options for our Poll
+type PollOpts struct {
+	// How long to wait for new Events
+	Timeout time.Duration
+
+	// We might add:
+	// maximum number of evens to return
+	// MaxEvents int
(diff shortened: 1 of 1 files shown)


JeanetteBruno:
This is a good concept to explore,  need more background on the use case...  I don't think this is the right venue to discuss my questions... can we cover this in a vc?

In the meantime, here is a summary of my understanding and questions.
I'm interpreting the use case as follows:
The user wants the ability to continuously receive the status of run(s).  
They will do so by 
1. defining a subscription as a list of Scoot Daemon runIds, with an optional state filter (mask)
2. subscribing (Subscribe()) to the definition
3. actually receive subscribed events in one of 2 ways:
	a. via a channel 
		i.  requesting that the daemon start streaming run status events, starting at an event sequence number (Stream())
		ii. loop pulling events from this channel
	b. via a Poll request
		i. receive a list of events matching the subscription definition once the subscribed runs have met the state filter or a timeout parameter is satisfied
4. unsubscribing when they no longer want to receive events

 questions:
 - What is the user's motivation? - why do they need to continuously receive the state of runs?
- The users would have to write a loop to pull values off the Stream() channel - isn't it simpler for them to just poll each time and not deal with setting up stream?
 - How does the user obtain the starting event sequence number? (from prior streaming?)
 - Does unsubscribe only apply with Stream()?
 - The users would have to write a loop to pull values off the Stream() channel - isn't it simpler for them to just poll each time and not deal with setting up stream?
The users would have to write a loop to pull values off the Stream() channel - isn't it simpler for them to just poll each time and not deal with setting up stream?

dgassaway:
I like the subscription idea from the perspective that it could simplify poll-and-wait (or add flexibility to that).  I'm not totally sold on how this would address that, though. Echoing some of Jeanette's comments:

* Presumably the subscriber is on the client side - how are they getting the events off of a channel?  Is the subscriber implementation polling the server and populating it?  Or is this the server side logic that will need a corresponding client interface later?
* Stream and Poll being based on Event seems odd (plus this seems to be a chicken-and-egg problem where we need an event to start streaming events) - why not use the subscription ID? The "server-side" could keep track of what was sent to a subscriber, and drop events as they are consumed (prevents rewinding).

PR #157, opened 2016-11-03
snapshot/gitfiler - add support for checkouter stats and small fixes

dgassaway:
Adds some basic metrics - uptime and passes the stat handler into the checkouter install functions in magic bag so that checkouters and repository init can be instrumented with metrics.  If there is a better approach feel free to suggest.

- Pass stats receiver into snapshot.Checkouter Installer
- Pass stats receiver into RepoPool, RepoIniter interface
- Record latency (elapsed) time metric for git cloner init
- Uptime metric for workerserver
- Stats Installer sets receiver to ms precision. Default is ns, but our current metric emissions are labelled ms (and are off by 1 million)
- Normalize use of CurrentStatsReceiver vs NilStatsReceiver() in tests

https://phabricator.twitter.biz/D11349


sched/scheduler/stateful_scheduler_test.go
```diff
@@ -46,7 +46,7 @@ func makeStatefulSchedulerDeps(deps *schedulerDeps) *statefulScheduler {
 		deps.sc,
 		deps.wf,
 		deps.config,
-		stats.NilStatsReceiver(),
+		stats.CurrentStatsReceiver,
```
CaitieM20: do you need to change this? are you vetting stats in the test cases?

scootapi/server/setup.go
```diff
@@ -32,7 +34,9 @@ func makeServers(
 func Defaults() (*ice.MagicBag, jsonconfig.Schema) {
 	bag := ice.NewMagicBag()
 	bag.PutMany(
-		endpoints.MakeStatsReceiver,
+		func(scope endpoints.StatScope) stats.StatsReceiver {
```
CaitieM20: Twitter Obs only supports time series that are at the minute granularity.  There is no way to get smaller

snapshot/git/gitfiler/checkouter_test.go
```diff
@@ -24,7 +25,7 @@ func TestCheckouter(t *testing.T) {
 	}
 	doneCh := make(chan struct{})
 	defer close(doneCh)
-	checkouter := NewRefRepoCloningCheckouter(&ConstantIniter{repo}, tmp, doneCh, 0)
+	checkouter := NewRefRepoCloningCheckouter(&ConstantIniter{repo}, stats.CurrentStatsReceiver, tmp, doneCh, 0)
```
jasonbs10: Same comment as Caitie, we should explicitly use NilStatsReceiver in tests.

snapshot/git/gitfiler/pool.go
```diff
@@ -40,7 +43,8 @@ func NewRepoPool(initer RepoIniter,
 // Cf. sync's Pool
 // Supports maximum pool capacity - 0 is unlimited
 type RepoPool struct {
-	initer RepoIniter
+	initer  RepoIniter
+	statRec stats.StatsReceiver
```
jasonbs10: I'd like to stick with 'stat' as the name since that's what is is elsewhere.

workerapi/server/server.go
```diff
@@ -69,6 +70,7 @@ func (h *handler) stats() {
 			h.stat.Gauge("failedCachedRunsGauge").Update(numFailed)
 			h.stat.Gauge("endedCachedRunsGauge").Update(int64(len(processes)) - numActive)
 			h.stat.Gauge("timeSinceLastContactGauge_ms").Update(timeSinceLastContact_ms)
+			h.stat.Gauge("uptime_ms").Update(int64(time.Since(startTime) / time.Millisecond))
```
jasonbs10: Can you rename to match the variables above?

sched/scheduler/stateful_scheduler_test.go
dgassaway: This is semantic only, for now CurrentStatsReceiver==NilStatsReceiver().  However, in all other test cases we use Current and not Nil; just something I noticed for consistency.

scootapi/server/setup.go
dgassaway: Right - this is for how the internal stats receiver is handing time/numeric values.  We are currently reporting stuff like "X took 1000000 ms" when it really took 1ms.

snapshot/git/gitfiler/pool.go
dgassaway: Makes sense, was trying to avoid var name conflict, but not really an issue in this case

workerapi/server/server.go
dgassaway: will do

dgassaway:
request review of updates

snapshot/git/gitfiler/setup.go
```diff
@@ -4,49 +4,52 @@ import (
...
-	singlePool := NewRepoPool(nil, nil, doneCh, 1)
-	go func() {
-		r, err := repoIniter.Init()
-		singlePool.Release(r, err)
-	}()
+// Util for creating a Pool that will only have a single repo
+func NewSingleRepoPool(repoIniter RepoIniter,
+	stat stats.StatsReceiver,
+	doneCh <-chan struct{}) *RepoPool {
+	singlePool := NewRepoPool(repoIniter, stat, []*repo.Repository{}, doneCh, 1)
```
jasonbs10: It may have made more sense to have a comment before, but it could also be useful now - why are we no longer calling repoIniter.Init() here?

dgassaway: The previous implementation of a pool with only a single resource was a hack - the initer and getter funcs were nil-ed out so the pool could not initialize the resources, and then this manually initialized it.  With the capacity feature, we just give the pool a cap of 1 and it will initialize a single repo on its own.

PR #158, opened 2016-11-07
runner: StatusQuery allows waiting (WIP)

dbentley:
Every user of Runner wanted polling, so add polling.

runner/statuser.go is the new interface

Merge runner/local into runner/runners

runner/runners/README.md describes the new strategy: refactor the file f/k/a simple.go, which was doing three things:
*) should I run this -> SingleController (runner/runners/single.go)
*) how do I run this -> Invoker (runner/runners/invoke.go)
*) status database -> Statuses (runner/runners/statuses.go)

This refactor allows QueueingRunner to melt into QueueController (queue.go)

Extracts controller_test.go from simple_test.go

Awkwardly copy/update the SnapshotID due to merge conflicts (and remove its own type)

Currently, lots of code is broken because this changes an interface but not every implementation. If we like this, I'll go add the implementation (and start removing the polling code this obviates)


runner/runners/README.md
```diff
@@ -0,0 +1,29 @@
+# Runner Impl Design #
+This package contains implementations of Runner.
```
CaitieM20: Add a sentence about what Runner does.  If you don't know this doc provides no context

runner/runners/README.md
```diff
@@ -0,0 +1,29 @@
...
+* Controller
+* Invoker
+
+## Statuses (statuses.go) ##
+Statuses is a database of ProcessStatus. It has two facets:
+* Read Statuses (by implementing runner.Statuser). Allows reading the current state and waiting for new results.
+* Write Statuses. What writes Statuses?
+
+## Controller (single.go and queue.go) ##
+Controller offers Run() and Abort(). It offers mutual exclusion (don't run two at a time) and a queue (in queue.go; single.go doesn't). It writes to a Statuses.
```
CaitieM20: Add a sentence to describe what Controller is

runner/runners/README.md
```diff
@@ -0,0 +1,29 @@
+# Runner Impl Design #
+This package contains implementations of Runner.
+
+There are three main components of our Runner Impl:
+* Statuses
+* Controller
+* Invoker
+
+## Statuses (statuses.go) ##
+Statuses is a database of ProcessStatus. It has two facets:
```
CaitieM20: Is it a database?  Are these durably stored?

runner/runners/README.md
```diff
@@ -0,0 +1,29 @@
...
+* Read Statuses (by implementing runner.Statuser). Allows reading the current state and waiting for new results.
+* Write Statuses. What writes Statuses?
```
CaitieM20: add a TODO if you don't have an answer for this question

runner/runners/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Controller (single.go and queue.go) ##
+Controller offers Run() and Abort(). It offers mutual exclusion (don't run two at a time) and a queue (in queue.go; single.go doesn't). It writes to a Statuses.
+
+To actually start a process, it uses an Invoker.
+
+## Invoker (invoke.go) ##
+Invoker doesn't implement any interface; it's just a struct with useful methods. Invoker's Run method lowers a Scoot command to an Execer command. This involves:
```
CaitieM20: Add a sentence about what Invoker does first before describing implementations.  Struct with useful methods tells me nothing.  Its a struct that with methods for (... list things methods do...)

What does lower mean in this context? Sends a command to the Execer?

runner/runners/combine.go
```diff
@@ -0,0 +1,44 @@
+package runners
```
CaitieM20: file name should be the same as the struct it implements.  This makes finding this when browsing code much easier.

runner/runners/combine.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"github.com/scootdev/scoot/runner"
+)
+
+// Create a new ControllerAndStatuserRunner than uses cont for the controller and stat for the Statuser
+func NewControllerAndStatuserRunner(cont runner.Controller, stat runner.Statuser) *ControllerAndStatuserRunner {
```
CaitieM20: Please add docs to all these methods.

runner/runners/invoker.go
```diff
@@ -0,0 +1,137 @@
+package runners
+
```
CaitieM20: tests?

runner/runners/output.go
```diff
@@ -1,7 +1,15 @@
 package runners
```
CaitieM20: Recommend putting different implementations in different files

runner/runners/output.go
```diff
@@ -32,3 +40,61 @@ func (o *nullOutput) Write(p []byte) (int, error) {
 func (o *nullOutput) Close() error {
 	return nil
 }
+
+type localOutputCreator struct {
+	hostname string
+}
+
+// Create a new OutputCreator
+func NewOutputCreator(tmp *temp.TempDir) (runner.OutputCreator, error) {
```
CaitieM20: Should this be called NewLocalOutputCreator?  It creates a specific implementation of the interface.  Suggest changing the name and update comments

runner/runners/output.go
```diff
@@ -32,3 +40,61 @@ func (o *nullOutput) Write(p []byte) (int, error) {
...
+	hostname, err := os.Hostname()
+	if err != nil {
+		return nil, err
+	}
+	return &localOutputCreator{hostname: hostname}, nil
```
CaitieM20: why not use Create method?  This will simplify maintaining this code in the future

runner/runners/output.go
```diff
@@ -32,3 +40,61 @@ func (o *nullOutput) Write(p []byte) (int, error) {
...
+
+func (o *localOutput) Write(p []byte) (n int, err error) {
+	return o.f.Write(p)
+}
+
+func (o *localOutput) Close() error {
+	return o.f.Close()
+}
+
+// Return an underlying Writer. Why? Because some methods type assert to
```
CaitieM20: Please add Comments for all Public methods

runner/runners/queue.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func NewQueueRunner(exec execer.Execer, filer snapshot.Filer, outputCreator runner.OutputCreator, capacity int) runner.Runner {
+	statuses := NewStatuses()
+	invoker := NewInvoker(exec, filer, outputCreator)
+	controller := &QueueController{statuses: statuses, invoker: invoker, capacity: capacity}
+	return NewControllerAndStatuserRunner(controller, statuses)
+}
+
+// TODO(dbentley): QueueController and SingleController are so similar...
```
CaitieM20: Suggest Removing Comment:  These two controllers are semantically different right?  Forcing an abstraction on them to save code, will result in code that is harder to maintain / modify down the line

runner/statuser.go
```diff
@@ -0,0 +1,131 @@
...
+	// If negative, wait until there are results
+	Timeout time.Duration
+
+	// We might add:
+	// maximum number of evens to return
+	// MaxEvents int
+}
+
+// Statuser allows a client to read ProcessStatus'es
+type Statuser interface {
```
CaitieM20: This name is a bit odd since you added -er to a noun.  Suggest StatusChecker for more clarity

CaitieM20:
I question whether this batch query interface should be in the Runner codebase.  

If runner is supposed to be common code shared between the Worker nodes and the Scoot Local Daemon, this will only ever be used by the later.  To me this suggests moving it up a layer above the runner.

In addition it seems like formulating query that checks the status of multiple Jobs would be useful at the CloudExec Level too, which also suggests that this code should live at a higher level, and perhaps not use the RunId as the query mechanism, but the JobId/TaskId.  

A suggestion would be to modify the current CloudExec api to support querying the GetStatus API and supporting taking a list of TaskIds to query for or the masks.  This will be generally useful and greatly reduce network traffic.

runner/runners/queue.go
dbentley: Now that I think of it, SingleController is the same as QueueController with capacity=0... Should I just remove SingleController then?

runner/runners/output.go
dbentley: Yes; done.

runner/runners/output.go
dbentley: Create takes a localOutputCreator to return a localOutput; this function is the one to create a localOutputCreator to begin with.

runner/runners/output.go
dbentley: Yup; done.

runner/runners/invoker.go
dbentley: Right now it's tested as part of the runner end-to-end tests. I'm happy to add tests if we think this is the right structure/decomposition.

runner/runners/combine.go
dbentley: Done.

runner/runners/README.md
dbentley: Rephrased; how's that?

runner/runners/README.md
dbentley: How's that?

runner/runners/README.md
dbentley: Ugh, I was trying to lead into the next paragraph but it was unclear. Rephrased. How's that?

runner/runners/README.md
dbentley: It's in-memory only; rephrased.

runner/runners/README.md
dbentley: Done.

dbentley:
I agree that CloudExec API needs something similar-ish, but in terms of its API.

The motivation here is that right now we use the Runner in two places:
*) Daemon
*) Worker

And despite having two uses, we actually have *three* implementations of polling:
*) sched/worker/polling.go
*) daemon/server/handler.go
*) runner/local/simple_test.go

So, the reason for this is more for the Polling/Waiting than the Batching.

The Batching will be useful for the Daemon API right now, which could make sense moving it up. But I suspect it will make sense for the Worker eventually, when we are running multiple Tasks simultaneously on one Worker (which won't be for a long time!).

runner/runners/combine.go
dgassaway: Suggest a name that conveys that is notable about this Runner.  Namely it's pluggable or controllable, so maybe "PluggableRunner", "ControllableRunner", or "FlexibleRunner", and then matching the file name.

runner/runners/queue.go
```diff
@@ -0,0 +1,116 @@
...
+func (c *QueueController) Run(cmd *runner.Command) (runner.ProcessStatus, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if len(c.queue) >= c.capacity {
+		return runner.ProcessStatus{}, fmt.Errorf(QueueFullMsg)
+	}
+	st := c.statuses.NewRun()
+	if c.runningID == runner.RunId("") {
+		return c.start(cmd, st.RunId)
```
dgassaway: It seems odd to have the "start" function return a status, and to have this function potentially return different states (new run vs something returned by start()). In either case the caller will have to poll for state later anyway for completeness, would simplify by not returning anything.

runner/runners/queue.go
```diff
@@ -0,0 +1,116 @@
...
+
+	go func() {
+		st := c.invoker.Run(cmd, c.runningID, c.abortCh, updateCh)
+		close(updateCh)
+		c.finish(st)
+	}()
+	return c.statuses.Status(id)
+}
+
+// finish is a callback to be called when a process is finished with its final state
```
dgassaway: Hard to follow at first that the queue pop & run logic is in here. Any opinion on having an "appendAndStart" function that can be called from Run() and finish() that handles the queue logic here, and could also replace the "if runnerID == RunID(""); start()" logic in Run().

runner/statuser.go
dgassaway: +1, I read this as stat-user

runner/runners/README.md
```diff
@@ -0,0 +1,31 @@
+# Runner Impl Design #
+package runner defines Runner, which lets clients start, stop, and get the status of Scoot Commands. This package contains implementations of Runner.
+
+There are three main components of our Runner Impl:
+* Statuses
```
dgassaway: The relationship between these is confusing. Not clear which are interfaces (controller, statuser), implementations (queuingcontroller, statuses), and just structs (invoker).  Seems like at a high level, that status and controller are most important, and invoker is an implementation detail of our particular controller.

runner/runners/combine.go
```diff
@@ -0,0 +1,51 @@
...
+	return &ControllerAndStatuserRunner{cont, stat}
+}
+
+// ControllerAndStatuserRunner combines a Controller and a Statuser to make a Runner.
+type ControllerAndStatuserRunner struct {
+	cont runner.Controller
+	stat runner.Statuser
```
jasonbs10: Can't you just make these embedded interfaces and delegate automatically?

runner/runners/invoker.go
```diff
@@ -0,0 +1,139 @@
...
+type checkoutAndError struct {
+	checkout snapshot.Checkout
+	err      error
+}
+
+// Run runs cmd as run id returning the final ProcessStatus
+// Run will send updates the process is running to updateCh.
+// Run will enforce cmd's Timeout, and will abort cmd if abortCh is signaled.
+// Run will not return until the process is not running.
+func (in *Invoker) Run(cmd *runner.Command, id runner.RunId, abortCh chan struct{}, updateCh chan runner.ProcessStatus) runner.ProcessStatus {
```
jasonbs10: This will eventually need some comments, it's mostly a wall of text right now.

runner/runners/invoker.go
```diff
@@ -0,0 +1,139 @@
...
+	"log"
+	"time"
+
+	"github.com/scootdev/scoot/runner"
+	"github.com/scootdev/scoot/runner/execer"
+	"github.com/scootdev/scoot/snapshot"
+)
+
+// NewInvoker creates a NewExecer that will use the supplied helpers
+func NewInvoker(exec execer.Execer, filer snapshot.Filer, outputCreator runner.OutputCreator) *Invoker {
```
jasonbs10: I don't love the having three similar sounding types, Execer/Invoker/Runner, but I'm ok with how they're split up I guess. Can we do Execer/ExecerController/Runner, or something else that doesn't reuse 'controller'?

runner/runner.go
```diff
@@ -53,25 +53,23 @@ func NewCommand(argv []string, env map[string]string, timeout time.Duration, sna
 	return &Command{Argv: argv, EnvVars: env, Timeout: timeout, SnapshotId: snapshotId}
 }
 
-type Runner interface {
+// Controller lets clients Control commands.
+type Controller interface {
```
jasonbs10: The name hierarchy could be clearer, maybe something like:
```
Runner {
  CmdController {
    StatusRecorder
    CmdExecerAdapter {
      Execer
    }
  }
  StatusChecker
}
```
This isn't great either but it's clearer I think.

runner/runners/README.md
```diff
@@ -0,0 +1,31 @@
+# Runner Impl Design #
```
jasonbs10: It would be great if the interface/struct hierarchy could be outlined visually in one block of text near the top.

runner/runners/combine.go
JeanetteBruno: pattern I've used is to have
- X'ers - providing entry points to X
- X'erHelper - providing utility routines for implementing X (hidden to all users of X)
- X'erManager - typically provides X entry points, plus additional API, as necessary, for handling request volume and multiple Xers and using inner X'ers to actually do the work
So we would have: 
RunnerAndStatus
	QueuingRunner/RunnerManager (I like either name)
		Runner
			RunnerHelper
				Execer
					OsExecer
	Status
I've put a sample architecture diagram and component list on https://docs.google.com/document/d/1fKU6Pp8kmXKYuzV-pH47ODJwTn3Isdp64KY32AQammA/edit# (this content would be put in the readme.md)

runner/statuser.go
JeanetteBruno: I would just go with Status for the interface and InMemoryStatus for the implementation (replace Statuses).

Also, shouldn't the interface have the full CRUD operations? I see Retrieve and Delete, what about Create and Update?

runner/statuser.go
```diff
@@ -0,0 +1,138 @@
...
+	StatusAll() ([]ProcessStatus, error)
+
+	// Prunes the run history so StatusAll() can return a reasonable number of runs.
+	Erase(run RunId) error
+}
+
+// Helper Functions to create StatusQuery's
+
+// Query for when id is done
+func RunDone(id RunId) StatusQuery {
```
JeanetteBruno: these function names are easy to mis-read.... perhaps: MakeDoneQuery(), MakeCurrentQuery(), MakeStateQuery(),  MakeCurrentPollOpts(), MakeWaitPollOpts() (not sure we need the word 'Run' in the function names since they are functions within runner.)

runner/runners/queue.go
```diff
@@ -0,0 +1,116 @@
...
+		for i, cmdAndID := range c.queue {
+			if runId == cmdAndID.id {
+				c.queue = append(c.queue[:i], c.queue[i+1:]...)
+				c.statuses.Update(runner.AbortStatus(runId))
+			}
+		}
+	}
+
+	// Unlock so the abort can call finish()
+	c.mu.Unlock()
```
JeanetteBruno: Having the unlock here is dangerous ... what happens if c.statuses.Update() panics?

Does Abort() have to return a ProcessStatus? Is error sufficient?

What are the semantics for all cases: abort if the runId is running or pending, error if the runId is already done?

runner/runners/queue.go
```diff
@@ -0,0 +1,116 @@
...
+	statuses *Statuses
+	invoker  *Invoker
+	capacity int
+
+	runningID runner.RunId
+	abortCh   chan struct{}
+	queue     []commandAndID
+	mu        sync.Mutex
+}
+
```
JeanetteBruno: It seems odd to have status control the runId - I understand that it makes sense when thinking about runId as a db key, but it just makes me pause.

CaitieM20:
Dan and I discussed this today via hangouts:  

The fundamental problem here is the Runner Interface as it is today does too much.  The original goal of the Runner was to encapsulate shared logic need to run a task.  This can be shared between the Scoot Cloud Workers and the Local Scoot Daemon. 

One of the core problems is the async nature of Run function.  There is no reason to do this asynchronously since the context which this library runs in is always a local one.  It should just block until the task is completed, and potentially provide an update channel.  This removes the need for polling completely.  

The Worker code, will need to provide an API for polling for failure detection between itself and the scheduler, but this is because its running as part of a distributed system.  This is not core Runner functionality.

The Local Scoot Daemon will need to provide code which controls what tasks get sent to the Runner and to Scoot Cloud Exec.  But the runner in this context is not part of a distributed system, failures are correlated since they are on the same machine, and doesn't need to provide a polling API.  Ideally we want these code bases that use the Runner to evolve separately.  Their contexts are different (remote execution vs local), multi-tenancy concerns are different (cloud workers are uniform, whereas resources on dev laptops are not) etc... They look very similar now but will diverge in the future, so we should not try to make a common abstraction here.  The common shared code should be what runs the task, and this is all the Runner interface should supply.  

The proposal we came up with was to replace the Runner Interface with a slimmed down version.  The Run command, returns when the task has completed.  It can also take a channel which updates are sent on.  This way callers of this method are notified immediately when a task completes instead of having to poll a local process.

dbentley:
Closing for now until it's ready again

PR #159, opened 2016-11-08
runner: decompose runner.Runner (WIP)

dbentley:
It seems like maybe the decomposition of runner/local can be moved up and just be offered as individual building blocks.

Document the components of Runner and the design in runner/README.md


runner/controller.go
```diff
@@ -0,0 +1,16 @@
+package runner
+
+// Controller controls runs, implementing tenancy and queuing (or not).
+type Controller interface {
```
CaitieM20: So this is the piece that i'm suggesting should be fundamentally different, whether its running on the worker or via local scoot daemon.  And if you want this interface for the Local Scoot Daemon it should only be part of that code base.  

The worker isn't going to wait a short period of time for the command to finish.  I think this is an extra layer that is not needed, and should be up to the worker code & scoot daemon code independently to decide how to call the runner. 

I suggest common functionality like checking if a command is well formed should be pulled out into a utility library.  (this could go in the runner package).

runner/invoker.go
```diff
@@ -0,0 +1,14 @@
+package runner
+
+// Invoker runs a Scoot command. This involves:
+//   * setup (check out a snapshot)
+//   * post-processing (save output)
+//   * babysitting (implement timeout and abort)
+//   * updates while running
+type Invoker interface {
```
CaitieM20: this is great I like this!

runner/statuses.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Statuses allows Querying the status of Processes
+type Statuses interface {
+	// QueryWait queries the statuses for statuses that match q, waiting if none match now.
+	// Returns the ProcessStatus'es that match the Query.
+	// (i.e., if the query is for 3 RunIds but only one matches the StateMask,
+	// only one will be returned)
+	// If timeout is 0, same as QuerySync
+	// If positive, wait up to that long
+	Query(q StatusQuery, timeout time.Duration) ([]ProcessStatus, error)
```
CaitieM20: why do we need to Query if we just wait for the function to return now?  The Local Scoot Daemon can just run a task and directly get notified.

runner/statuses.go
```diff
@@ -0,0 +1,70 @@
...
+
+	StatusesSync
+}
+
+// StatusWriter allows updating Statuses
+type StatusWriter interface {
+	// NewRun creates a new run, with a new RunId in status PENDING
+	NewRun() (ProcessStatus, error)
```
CaitieM20: i don't understand how this fits in with just calling Run?

runner/statuses.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// Update writes a new status.
+	// It enforces several rules (not errors)
+	//   cannot change a status once it is Done
+	//   cannot erase Stdout/Stderr Refs (but can update to new values)
+	Update(ProcessStatus) error
```
CaitieM20: Why would you call update aren't you just pushing the updates to the channel supplied when calling Run?

runner/statuses.go
CaitieM20: I dont think we should attempt to create a unifying abstraction around getting the status of a local call vs a remote call, at least right now

These are fundamentally different and doing this is not trivial nor a good idea see ([A Note on Distributed Computing](http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.41.7628))

runner/runner.go
```diff
@@ -53,25 +53,10 @@ func NewCommand(argv []string, env map[string]string, timeout time.Duration, sna
...
-	// Prunes the run history so StatusAll() can return a reasonable number of runs.
-	Erase(run RunId) error
-}
+// TODO(dbentley): do we want to have Runner interface?
+// It combines a few things that are related, but different responsibilities.
+// Perhaps we could rename it something like "RunService", to make clear it's both ends?
+// type Runner interface {
+// 	Controller
+// 	LegacyStatuses
+// }
```
jasonbs10: I'd want something like a RunService interface plus a good default impl or two so any interaction between Controller & Statuses can be encapsulated at the right level. There is no interaction now but there could be...

JeanetteBruno:
I thought we would want the daemon to run commands async - to take advantage of the multi-core features on our laptops. I agree that the distributed scoot needs some additional features such as multi-tenancy, security and more persistent persistence, but it seems to me that cloud scoot and local scoot can share a high-level pattern of a single entry point 'controller/scheduler' (accepting run and status requests), (multiple) 'workers/runners' to actually run the commands and a 'status module/saga' to manage the states of the runs (typically with CRUD operations).

I'm not sure we all understand/agree on the high level architecture ... critiquing individual lines of code does not seem to be the most effective way to do achieve this.  If we are going to do 'async group design' perhaps we should focus on coming to agreement on the readme contents first.  Another option is for all of us meet to come to consensus on the design (and produce an agreed on readme that defines the components their scope).

runner/README.md
```diff
@@ -0,0 +1,41 @@
+package runner is structs and interfaces to run Scoot processes.
+
+## Summary ##
+* Invoker runs one Command using Execer (and snapshot.Filer, etc.)
+* Controller controls when to Run something (using an Invoker)
+* Statuses queries Process history
+  * StatusesSync doesn't offer waiting, so requires client polling
```
JeanetteBruno: we should describe features in terms of what they do and only secondary in terms of what they don't do....
StatusesSync - returns the current status of requested runs.  This feature is exposed to the user (via entry points in the controller?)
LegacyStatuses - (is this really the name for the entry point?) - what does it do (what did the old Read API do?)
StatusWriter - creates and updates status entries. (be careful of the word 'Process' - not sure its well-defined at this point).  This feature is used by the Controller.

what about erasing statuses?
also - we are no longer offering the users a (blocking) 'wait for state' feature?  (I'm fine with that, just want to make sure)

runner/README.md
```diff
@@ -0,0 +1,41 @@
...
+Invoker runs a Scoot command. This involves:
+* setup (check out a snapshot)
+* post-processing (save output)
+* babysitting (implement timeout and abort)
+* updates while running
+
+Conceptually, it lowers a Scoot Command from a high-level abstraction (Snapshots, etc.) to an Execer command (about directories that are paths in the local filesystem).
+
+## Controller (controller.go) ##
+Controller controls the Runs, starting a new one (Run()) or ending one (Abort()). It implements:
```
JeanetteBruno: IMHO - Controller should support running multiple commands concurrently in both daemon and cloud scoot
and yes Controller should support accepting a command even if there are no resources to run it immediately.

runner/README.md
```diff
@@ -0,0 +1,41 @@
...
+* tenancy (run one at a time? Or multiple? Fixed, or based on system utilization?)
+* queueing (should we accept work we can't start right now)
+
+When a Controller wants to start a Run of a Command, it will use an Invoker.
+
+## Statuses (statuses.go) ##
+Statuses allows querying for Process Status.
```
JeanetteBruno: we should describe what a 'Process Status' ... if its the status of a 'Run of a Command' why don't we call it Run Status or Command Status - where did the word 'Process' come from?

runner/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+StatusesSync is a subset of Statuses that doesn't offer waiting. This is necessary because RPC transports like Thrift don't like waiting to respond. (We expect to offer a client-side polling library that can offer Statuses on top of only a StatusesSync).
```
JeanetteBruno: could we call this CurrentStatus?

Also - perhaps we should have a readme in Daemon/client that describes the API exposed to users.... if a client-side 'status wait' (polling) feature is being offered there, that is where it should be defined.

runner/README.md
```diff
@@ -0,0 +1,41 @@
...
+StatusWriter allows creating new Process's and updating existing Process's Status.
+
+LegacyStatuses offers a Legacy query API. Deprecated, and we intend to remove it.
+
+## Relation to Execer ##
+Execer is lower-level than runner. Runner deals with Scoot Abstractions; but Execer is just about Unix abstractions: files, directories, etc.
+
+OSExecer calls os/exec.
+
+SimExecer simulates behavior based on the args passed to it. This lets a caller script the behavior of the Execer.
```
JeanetteBruno: SimExecer, DoneExecer, InterceptExecer are alternative execers used in various test scenarios.
SimExecer - lets the test script the execer behavior
DoneExecer - immediately sets all run requests to completed
InterceptExecer - allows the use/injection of 2 different execers base on a test-defined boolean function on the run command.

runner/README.md
CaitieM20: agree i think this is a long term goal

dbentley:
Closing for now until it's ready again

PR #160, opened 2016-11-08
Jeanette/scoot200 integration test

JeanetteBruno:
adding a go, protobuf-agnostic, client for the daemon (client_lib.go) and integration tests using the client, server, queueing runner, simple runner and os execer.

The first objective was to add an integration test of the client, server, queueing runner and os execer that ran 'many' run commands and exercised the full suite of daemon APIs.  (note: but not all error conditions.)

The existing client code (conn.go) only implemented a subset of the interface and the existing integration-test code exercised this subset via the command line entry to the daemon.   Conn also required the caller to issue a Dial() command to get a connection (conn), before they could issue the API commands via the conn.  

The client_lib eliminates the caller having to issue the Dial() command.  They get a ClientLib (NewClientLib()) and use the API commands directly - (Dialing is hidden within ClientLib)

conn and client_lib should be merged and the missing command line entries need to be added (I think?).  I prefer exposing the term ClientLib to users (as opposed to Conn).  Looking for feedback.


jasonbs10:
Is this all necessary prior to trying out daemon w/thrift, and even then it only appears useful for A/B testing and not after we choose one protocol over the other? Can you add a motivation paragraph to the summary (or arrange a VC if that format is better)?

PR #162, opened 2016-11-10
Makefile: run swarmtest in Travis CI

dbentley:
run swarmtest in Travis CI

This uses setup-cloud-scoot which lets us run smoke test against a swarm without a specific binary.

Delete the old binary (well, test) in tests/swarmtest_test.go

Change .travis.yml a bit to run the new Make target "travis".

Also update .gitignore to exclude emacs auto-save files while I'm in this section of code (i.e., project config files at the root)


Makefile
```diff
@@ -58,10 +58,16 @@ test-integration:
```
ryancouto: tests/swarmtest/swarmtest.go was the only file with an integration build tag - think we can remove test-integration Make target. Otherwise lgtm

dbentley: I was thinking of leaving the integration test tag because it seems like it could be useful. But, I will add a comment saying we don't currently use it.

Makefile
```diff
@@ -54,14 +54,21 @@ test-unit:
...
 test-integration:
 	# Runs all tests including integration and property tests
+	# We don't currently have any integration tests, but we're leaving this so we can add more.
 	go test -race -tags="integration property_test" $$(go list ./... | grep -v /vendor/ | grep -v /cmd/)
 
 testlocal: generate test
 
+swarmtest:
+	go install github.com/scootdev/scoot/binaries/setup-cloud-scoot
+	${GOPATH}/bin/setup-cloud-scoot --strategy local.local run ${GOPATH}/bin/scootapi run_smoke_test 10 1m
```
jasonbs10: We have to be careful with GOPATH here and elsewhere. I have multiple paths in mine. Why do we need the go install bit anyway?

dbentley: Good catch. Thanks.

go run seems weird because you give it go files, not go packages.

So I figured you'd want to build the binary and then run it... WDYT? How can I do this right?

Makefile
```diff
@@ -54,14 +54,20 @@ test-unit:
 
 test-integration:
 	# Runs all tests including integration and property tests
+	# We don't currently have any integration tests, but we're leaving this so we can add more.
 	go test -race -tags="integration property_test" $$(go list ./... | grep -v /vendor/ | grep -v /cmd/)
 
 testlocal: generate test
 
+swarmtest:
```
jasonbs10: I'd comment on what this line is doing / what the various args are.

PR #166, opened 2016-11-12
Jschiller/python

jasonbs10:


JieGhost:
Can you extract all the python files that will be in the wheel into a separate dir? A layout like this:
scoot/
  client/
    python/
      bin/
      pyscoot/
      tests/
      setup.py
      ...
I think it will be more clear and easier to build a wheel from.
What do you think?

jasonbs10:
Without knowing much about python best practices, I think this change is overkill for a largely non-python project. I'm not sure why bin/ or tests/ is important for creating a wheel (are wheels supposed to contain tests?) and the current structure, which puts all the required files in daemon/protocol/python, seems sufficient.

Can you elaborate on why this change would be impactful?

JieGhost:
I don't think wheels contains tests dir. 
The structure I am suggesting does not have functional impact but will have structural benefits.
It is very common for a multi-language project to isolate files of different languages if possible. 

You put all required files in daemon/protocol/python, which I think is good. But there are 2 things I want to bring up:
1. Will you have clients for additional languages other than python in the future? If so, scoot/client/ or scoot/daemon/client/ may be a better place for those language specific code. daemon/protocol is more about daemon instead of clients IMHO.
2. I prefer to have python test files in the same dir with python source files (like I suggested above). It is less messy and developers can easily find what they want if they want to make changes to python code. It will be even better if you can separate go files and python files completely. I understand you have integration_py_test.go as a way to run python test under go test framework, but even if they are not in the same dir, you should still be able to invoke the integration test, right?

I personally feel more strongly about 2 than 1, but neither of them are blockers. Feel free to ignore them if you want.

Thanks for dealing with pypi stuff!

PR #167, opened 2016-11-15
runner: decompose runner.Runner

dbentley:
This is API/structure only change, as we've discussed.

Describe design in README.md

This breaks the project overall, in order to keep the number of files changed small.

If we like this, then I'll add implementations.


runner/status_rw.go
```diff
@@ -0,0 +1,116 @@
...
+type StatusQuerier interface {
+	// Query returns all RunStatus'es matching q, waiting as described by w
+	Query(q Query, w Wait) ([]RunStatus, error)
+
+	StatusQueryNower
+}
+
+// StatusQueryNower allows Query'ing Statuses but with no Waiting
+// We will implement a PollingQueuer that wraps a StatusQueryNower and satisfies QueryNower
+type StatusQueryNower interface {
```
JeanetteBruno: Not sure why we have StatusQueryNower and don't just have QueryNow() in StatusQuerier interface

dbentley: Expanded the comment; how's that?

runner/status_rw.go
```diff
@@ -0,0 +1,116 @@
...
+
+// Helper Function to create StateMask that matches exactly state
+func MaskForState(state ProcessState) StateMask {
+	return 1 << uint(state)
+}
+
+// Query describes a query for RunStatuses.
+// The Runs and States are and'ed: a RunStatus matches a Query
+// if its ID is in q.Runs (or q.AllRuns) and its state is in q.States
+type Query struct {
```
dbentley: I thought about that, but it seemed weird to have this magic value, and clearer (with little cost) to have an extra, specific, field. WDYT?

runner/status_rw.go
```diff
@@ -0,0 +1,116 @@
...
+// We will implement a PollingQueuer that wraps a StatusQueryNower and satisfies QueryNower
+type StatusQueryNower interface {
+	// QueryNow returns all RunStatus'es matching q in their current state
+	QueryNow(q Query) ([]RunStatus, error)
+}
+
+// Convenience methods for common queries in terms of the more general Query interface.
+
+// Status returns the current status of id from q.
+func Status(q StatusQueryNower, id RunID) (RunStatus, error) {
```
dbentley: They are both in LegacyStatusReader, below.

This Status is a concrete function that's a convenience function that adapts the legacy interface to the general, minimal interface.

runner/README.md
```diff
@@ -0,0 +1,57 @@
+# Runner Design #
+package runner is structs and interfaces to run Scoot processes. package runners contains implementations. This doc describes both.
```
JeanetteBruno: we should establish key terms at the top of the readme and use them consistently throughout (eg: command, runRequest, etc.).

Package 'runner' is structs and interfaces to run a user defined (unix) command.  Package 'runners' contains an implementation.  This doc describes both

runner/status_rw.go
```diff
@@ -0,0 +1,116 @@
...
+// Wait describes how to Wait
+type Wait struct {
+	// How long to wait for Statuses
+	Timeout time.Duration
+
+	// We might add whether to return as soon as one status matches, or waiting until all
+}
+
+// StatusQuerier allows reading Status by Query'ing.
+type StatusQuerier interface {
```
dbentley: Now that I think about it, I think eventually it will just be StatusReader, once we've shaken out the LegacyStatusReader usages and can collapse them...

runner/README.md
```diff
@@ -0,0 +1,57 @@
...
+  * implements Abort and Timeout
+* runners.Statuses (struct) offers read and write of RunStatus
+  * satisfies interfaces StatusWriter and StatusReader
+
+## Invoker (runner/runners/invoker.go) ##
+Invoker runs a Scoot command. This involves:
+* setup (check out a snapshot)
+* post-processing (save output)
+* babysitting (implement timeout and abort)
+* status updates while running/once finished
```
JeanetteBruno: I don't understand 'status updates while running/once finished'

runner/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Conceptually, it lowers a Scoot Command from a high-level abstraction (Snapshots, etc.) to an Execer command (about directories that are paths in the local filesystem).
```
JeanetteBruno: 'Conceptually, it installs the command's snapshot files, then uses an Execer to run the command.  When the command has finished, it created a snapshot of the command's output.'  ?better?

runner/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Controller (runner/controller.go) ##
+Controller controls the Runs, starting a new one (Run()) or ending one (Abort()). It implements:
```
JeanetteBruno: chand Runs to Run requests?
tenancy - prevents multiple run requests from impacting each other.  Currently only runs one run request at a time.
queue - manages multiple run requests while the invoker is busy

dbentley: Is there a difference between a run request and a Command?

runner/controller.go
```diff
@@ -0,0 +1,16 @@
...
+
+type Controller interface {
+	// Run prepares cmd. It returns its status and any errors.
+	// Run may:
+	// enqueue cmd (leading to state PENDING)
+	// run cmd immediately (leading to state RUNNING)
+	// check if cmd is well-formed, and reject it if not (leading to state FAILED)
+	// wait a very short period of time for cmd to start (leading to any state)
+	// Run may not wait indefinitely for cmd to finish. This is an async API.
+	Run(cmd *Command) (RunStatus, error)
```
JeanetteBruno: based on prior conversation with Dan.... should we just return RunStatus (eliminate error) - put errors from the scoot platform in the RunStatus's .error field?  This applies to all APIs that return a status and an error value

runner/runner.go
```diff
@@ -34,8 +34,8 @@ type Command struct {
 
 func (c Command) String() string {
 	var b bytes.Buffer
-	fmt.Fprintf(&b, "Command\n\tSnapshot ID:\t%s\n\tArgv:\t%q\n\tTimeout:\t%v\n",
-		c.SnapshotId,
+	fmt.Fprintf(&b, "Command\n\tSnapshotID:\t%s\n\tArgv:\t%q\n\tTimeout:\t%v\n",
```
JeanetteBruno: change Argv to Command?

runner/runner.go
```diff
@@ -49,29 +49,22 @@ func (c Command) String() string {
 	return b.String()
 }
 
-func NewCommand(argv []string, env map[string]string, timeout time.Duration, snapshotId string) *Command {
-	return &Command{Argv: argv, EnvVars: env, Timeout: timeout, SnapshotId: snapshotId}
+func NewCommand(argv []string, env map[string]string, timeout time.Duration) *Command {
```
JeanetteBruno: NewRunRequest(....)

also - don't we need to provide SnapshotId?

runner/runners/facade.go
```diff
@@ -0,0 +1,21 @@
...
+// NewQueueRunner creates a new Service that uses a Queue
+// (This is what the Daemon and Worker will call for now, and it will take an Execer, Filer, &c.)
+func NewQueueRunner() runner.Service {
+	controller := NewQueueController()
+	statuses := NewStatuses()
+	return &ServiceFacade{controller, statuses, statuses}
+}
+
+// ServiceFacade makes a Service from component parts.
+type ServiceFacade struct {
```
JeanetteBruno: I don't understand why we need ServiceFacade - how/where would we use it?

runner/controller.go
dbentley: We tried this before and I think it went poorly, so I think we'd rather go the other way. E.g., if the network is down, you want to get an error.

Maybe I should add that we want to get rid of RunState "BADREQUEST" altogether...

runner/runner.go
dbentley: This is the argv field within Command, so I think it's good to say Argv.

runner/runners/facade.go
dbentley: It's the only thing that satisfies Service. So everywhere we pass around Runner today (that really needs the whole Service, not just the Controller or the StatusReader), we'd pass around a Facade.

runner/runners/invoke.go
```diff
@@ -0,0 +1,31 @@
+package runners
+
+import (
+	"github.com/scootdev/scoot/runner"
+	"github.com/scootdev/scoot/runner/execer"
+	"github.com/scootdev/scoot/snapshot"
+)
+
+// NewInvoker creates an Invoker that will use the supplied helpers
+func NewInvoker() *Invoker {
```
dbentley: Yup, I was going to add that as part of the implementation.

runner/runner.go
dbentley: Oh, are you saying we should change Command to RunRequest?

I think it turns out that aybe we don't need this function at all, cause you can just create a Command.

runner/runner.go
```diff
@@ -49,29 +49,22 @@ func (c Command) String() string {
...
-	// Run may not wait indefinitely for cmd to finish. This is an async API.
-	Run(cmd *Command) (ProcessStatus, error)
-
-	// Status checks the status of run.
-	Status(run RunId) (ProcessStatus, error)
-
-	// Current status of all runs, running and finished, excepting any Erase()'s runs.
-	StatusAll() ([]ProcessStatus, error)
+// Service allows starting/abort'ing runs and checking on their status.
+type Service interface {
```
dgassaway: Agree. Also, am not a big fan of the alias, unless we are really planning on cleaning it up soon, this is the type of thing that is easy to deprioritize forever and increase cruft.

runner/runners/queue.go
```diff
@@ -0,0 +1,38 @@
...
+// NewQueueController creates a new QueueController
+func NewQueueController() *QueueController {
+	return &QueueController{}
+}
+
+// QueueController maintains a queue of commands to run (up to capacity).
+type QueueController struct {
+	inv          *Invoker
+	queue        []cmdAndID
+	int          capacity
```
dgassaway: switch type/varName

runner/status_rw.go
dgassaway: Given naming of "status_rw.go" I think it makes sense to use "*Reader" and "*Writer" where appropriate

runner/status_rw.go
```diff
@@ -0,0 +1,120 @@
...
+	// Query returns all RunStatus'es matching q, waiting as described by w
+	Query(q Query, w Wait) ([]RunStatus, error)
+
+	StatusQueryNower
```
dgassaway: I almost want to suggest reversing the naming between QueryNower and Querier, and instead have something like PollingQuerier or WaitingQuerier and just Querier underneath.  I do buy in to the interface separation, though.

runner/status_rw.go
dgassaway: Can we distance naming between wrapper and core interface more - it makes reading confusing.  Suggest GetStatus or QueryStatus

runner/status_rw.go
jasonbs10: If we want to keep AllRuns we should elaborate on the interaction between that and Runs.

runner/status_rw.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// StatusQueryNower allows Query'ing Statuses but with no Waiting
+// This is separate from StatusQuerier because talking to the Worker, e.g., we will be able to
+// QueryNow easily, but because Thrift doesn't like long waits on RPCs, it can't do
```
jasonbs10: How do we know thrift doesn't like long RPC waits? And do we think callers would want a raw StatusQueryNower vs the more comprehensive StatusQuerier, i.e. do we have to expose Nower as a separate iface?

And the thing about special values for runner.Wait is that I'll look to find out what those are when reading through code, but there's no corresponding flag indicating that Query() has a QueryNow() cousin.

runner/runner.go
```diff
@@ -21,7 +21,7 @@ type Command struct {
 	Timeout time.Duration
 
 	// Runner can optionally use this to run against a particular snapshot. Empty value is ignored.
-	SnapshotId string
+	SnapshotID string
```
CaitieM20: Nit: Is changing the capitalization necessary?  Is this going to cause a bunch of breaking changes in twitter source?

runner/runner.go
CaitieM20: Nit: godoc comments please!

runner/runners/chaos.go
```diff
@@ -49,39 +49,39 @@ func (r *ChaosRunner) delay() error {
 }
 
 // Implement Runner
```
CaitieM20: nit: is this comment still correct?  It is implemented?

PR #168, opened 2016-11-15
snapshot/snapshots: FilerFacade combines Checkouter and Ingester to make a Filer

dbentley:
This is needed internally, where we have a separate Checkouter that we want to combine with a Filer.


workerapi/server/setup.go
```diff
@@ -59,10 +58,6 @@ func Defaults() (*ice.MagicBag, jsonconfig.Schema) {
 
 		func() (*temp.TempDir, error) { return temp.TempDirDefault() },
```
dgassaway: Just curious, why'd we remove the default?

dbentley: Because we need to override it in Twitter (and right now there's no way to override a default in ice)

snapshot/snapshots/fake_filer.go
```diff
@@ -87,32 +87,38 @@ func (t *tempFiler) CheckoutAt(id string, dir string) (snapshot.Checkout, error)
...
+type FilerFacade struct {
+	snapshot.Checkouter
+	snapshot.Ingester
 }
 
-// Make a Filer that can Checkout() but does a noop Ingest().
-func MakeTempCheckouterFiler(tmp *temp.TempDir) snapshot.Filer {
-	return &tempCheckouterFiler{Checkouter: MakeTempCheckouter(tmp)}
+// Make a Filer from a Checkouter and Ingester
+func MakeFilerFacade(checkouter snapshot.Checkouter, ingester snapshot.Ingester) *FilerFacade {
```
jasonbs10: As in the other PR, I'd prefer SimpleFiler or some such.

workerapi/server/setup.go
jasonbs10: Why do you need to override prior to returning from Defaults() when you can just override the returned values?

PR #170, opened 2016-11-16
scoot: godocs and readmes

dgassaway:
Some general notes:
* No code changes here
* Goal is to make scoot code easier to understand, more discoverable, and set skeleton for more complete docs going forward
* This is not a total doc-style overhaul. But hopefully is useful.
* In general we have a lot of inconsistencies in where we use exported types/funcs, and interfaces vs exported types

Doc philosophy used here:
* Packages and key exported types and functions should have a godoc-compatible comment
* If the package doc is really long or there's no obvious file for it, throw it in a doc.go file
* Broader diatribes on high level functionality or interaction better suited to readmes
* Probably fewer, more detailed readmes is better than many tiny readmes nested around


daemon/protocol/README.md
```diff
@@ -1,16 +1,34 @@
...
+```
+
+#### Compiling for Python (daemon_pb2.py)
+
+* Install latest protobuf and grpc+tools packages:
+```sh
+pip install grpcio-tools
+```
+
+_Editor's note_: we should think about using a _virtualenv_ for a fully portable process, and to
```
jasonbs10: I think this may more overhead than it's worth. We currently require grpc 1.0.1 but we could also say it's a trustworthy package and accept anything greater than that. That wouldn't be too onerous...

proc/proc.go
```diff
@@ -1,3 +1,6 @@
+// Package proc includes definitions for Scoot's
+// Job, Task and associated structures.
+// TODO this seems deprecated - Is this real?? DO I EXIST??
```
jasonbs10: +1

runner/runner.go
```diff
@@ -1,3 +1,7 @@
+// Package runner provides for execution of Scoot work and retrieval
+// of the results of that work. The core of this is the Runner interface.
+// This package includes the a standard runner as well as test and simulation
```
jasonbs10: the a -> a

saga/saga.go
```diff
@@ -1,3 +1,5 @@
+// Package saga provides a generic implementation of the Saga pattern
+// that can be used as a durability mechanism for the Scoot scheduler.
```
jasonbs10: Might be a good place to link to external saga docs (caitie's presentation link or otherwise).

sched/README.md
```diff
@@ -0,0 +1,14 @@
+# Scoot Scheduler
+
+The Scoot Scheduler is responsible for receiving job requests from the Cloud
+API, and distributing the jobs to workers and maintaining and communicating
+the state of jobs run.
+
+The scheduler code contains the following packages:
+* __sched__ - definitions for jobs, tasks, and states
+  * __thrift definitions for the above__
+* __schedthrift__ - generated code from the thrift definitions
```
jasonbs10: Is the indentation correct here? It doesn't quite look right.

scootapi/README.md
```diff
@@ -0,0 +1,61 @@
...
+* __SchedulerStrategy__ - definition of the logic for invoking the scheduler, including WorkersStrategy and Builder
+
+Other Concepts:
+* __WorkersStrategy__ - defines the logic for how to invoke the Workers
+* __Builder__ - interface that allows specification of how to create the scheduler and worker binaries Cloud Setup will invoke
+
+## ScootAPI Thrift Code
+
+__Dependencies for installing thrift__
+* Install Thrift manually (version >= 0.9.3). macOS:
```
jasonbs10: Seems like this thrift setup stuff should be moved up a level, only the actual proto generation is specific to scootapi.

scootapi/setup/setup.go
```diff
@@ -1,3 +1,6 @@
+// Package setup provides Cloud Scoot Setup, which is a mechanism for
```
jasonbs10: It would also help if we documented some of our conventions. Like flesh this out a little bit and make it applicable to the other components that do setup type stuff.

snapshot/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+Snapshots are Scoot's representation of file system states. Snapshots are immutable,
+and are labelled by a unique ID.
+
+The snapshot code consists of the main Snapshot interface, and various implementations
+for dealing with snapshots in different filesystem paradigms - simple directories, git
+repositories, etc.
+
+Other key concepts here include:
+* Filer
```
jasonbs10: We'll want to elaborate on things like this since it's not exactly intuitive what ex: Filer represents.

snapshot/git/repo/repo.go
```diff
@@ -1,3 +1,6 @@
+// Pacakge repo provides utilities for operating on a git repo.
```
jasonbs10: Pacakge -> Package.

It would be nice if our code editors caught this for us... I'm pretty sure emacs can spellcheck comments but I haven't ever bothered to set that up.

snapshot/snapshots/fake_checkouter.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/scootdev/scoot/snapshot"
 )
 
+// Create a Checkouter that always fails due to invalid path
```
jasonbs10: It doesn't actually fail. I'm not sure why we say it's a noop struct and then say MakeInvalid...

jasonbs10:
I'm still not clear on what's better as a doc.go vs a README.md. Do we need doc.go? Is there a clear rule we can use?

README.md
```diff
@@ -0,0 +1,48 @@
...
+
+Run a series of randomly generated tests against the local scheduler and workers:
+
+```sh
+go run ./binaries/scootapi/main.go run_smoke_test
+```
+
+### TBD
+Scoot contains many pluggable/overridable interfaces & implementations
+
```
JeanetteBruno: Add a summary of 3rd party stuff that needs to be installed (with instructions):
-thrift
-protobuf
-python grpc
-?python docopt?
?(others)

daemon/protocol/README.md
```diff
@@ -1,16 +1,34 @@
...
-// To compile into daemon_pb2.py:
-//  pip install grpcio-tools # Install latest protobuf and grpc+tools packages.
-//  brew uninstall protobuf # Be sure to remove any conflicting/older versions of protobuf.
-//  python -m grpc.tools.protoc -I. --python_out=. --grpc_python_out=. daemon.proto
-//
+__This still needs to be integrated into the build__
+
+#### Compiling for Go (daemon.pb.go)
+
+* Install a protoc with libproto version >= 3.0.0 from https://github.com/google/protobuf
```
JeanetteBruno: reference installation instructions in scoot/readme.md
(leave the instructions for running the code generation here?)

daemon/README.md
```diff
@@ -1,5 +1,26 @@
...
+* collecting results of runs
+
+The daemon is a local persistent process that has a Protobuf listening
+interface, servers the Scoot Daemon API.
+
+### Scoot Daemon API
+
+The API supports the following actions. See the .proto or client files for more info:
+
+* __Echo__ - echo test to verify the daemon is available
```
JeanetteBruno: I would leave it in - its useful to verify that the daemon is running.

sched/README.md
JeanetteBruno: * __sched__ scheduler go objects: jobs, tasks and states, and thrift version of these objects
^* __schedthrift__ - generated code from the thrift definitions
^* __scheduler__ - interfaces and implementations for job scheduling
^* __worker__ - interface for communication between the scheduler and workers
^* __workers__ - implementations of the worker interface (polling worker, etc). Note that this is not the implementation of the Scoot Worker itself
 (^ is a space)
 (also - 'Note that this is not the implementation of the Scoot Worker itself' - we need to make this more clear (don't block PR with this))

proc/proc.go
dbentley: Yeah, remove is great.

snapshot/snapshots/fake_checkouter.go
dbentley: The checkouter is a noop checkouter because it does a noop on checkout, it just returns a static path (which happens to be an invalid one). You can also create a noopCheckouter with a valid path.

snapshot/snapshots/fake_checkouter.go
```diff
@@ -72,6 +73,7 @@ type Initer interface {
 	Init() error
 }
 
+// Creates a checkouter that waits for an Initer to be done Initing before checking out.
 func MakeInitingCheckouter(path string, initer Initer) snapshot.Checkouter {
```
dbentley: We can probably get rid of initingCheckouter... I think it's dead code? (Feel free to TODO)

workerapi/server/server.go
```diff
@@ -83,8 +86,7 @@ func (h *handler) updateTimeLastRpc() {
 	h.mu.Unlock()
 }
 
-// Implement thrift worker.Worker interface.
-//
+// Implements thrift worker.Worker interface
```
dbentley: For these comments, it might be better to say "implement worker.Worker.QueryWorker" or even better a summary of what it is. (If that's too much work for this awesome revamp, then maybe TODO?)

scootapi/setup/setup.go
dgassaway: This is the only thing I didn't directly address, wasn't sure how to go about this.

dgassaway:
Tried to address everything. Opened a new issue in JIRA (229) to track handling all of the 3rd party dep/codegen docs.
On doc.go - I think these are best ad-hoc, when we have more than a few lines for a package comment but don't want to have the code file have a giant comment header. README.md vs doc.go is more of a gray area... I see doc.go as package-specific developer details, and readme as a higher level guide.

PR #171, opened 2016-11-17
Jeanette/scoot175 daemon api

JeanetteBruno:
Looking for early feedback on this approach for a python command line inteface to the local daemon.

daemon_cli_test.py runs in my eclipse environment, but does not run via command line ($python -m unittest daemon_cli_test) - it fails because it cannot start the daemon.

I've checked in the docopt.py code for now (so folks don't have to install it, till we agree we want to use this tool).

This is not ready to merge to master yet - looking for feedback on:
- the use of docopt
- the syntax for the daemon commands
- 'pythonic' coding practices that I'm not using


daemon/protocol/python/scoot/daemon_cli.py
```diff
@@ -0,0 +1,156 @@
+"""
+Usage:
+  daemon_cli.py snapshot create <srcDir>
+  daemon_cli.py snapshot checkout <snapshotId> <destDir>
+  daemon_cli.py exec run <command> [--snapshotId=<sid>] [--timeout=<ns>]
+  daemon_cli.py exec poll <runIds> [--wait=<wait>] [--all]
```
dbentley: I think poll is weird; I'd suggest (though it may be for a different time) status

daemon/protocol/python/scoot/daemon_cli.py
```diff
@@ -0,0 +1,156 @@
...
+        raise docopt.DocoptExit(e)
+    try:
+#         print("starting client")
+        proto.start()
+    except proto.ScootException as e:
+        if re.search("UNAVAILABLE", e) > 0:
+            raise docopt.DocoptExit("Scoot Daemon is not running!\n")
+        raise docopt.DocoptExit(e)          
+        
+    if cmd['snapshot'] and cmd['create']:
```
dbentley: I'd start decomposing into functions, so that you can do:
 if cmd['snapshot']:
   snapshotCmd(cmd)
  elif cmd['exec']:
    execCmd(cmd)

Might make the structure easier to follow.

daemon/protocol/python/scoot/daemon_cli.py
```diff
@@ -0,0 +1,156 @@
+"""
+Usage:
+  daemon_cli.py snapshot create <srcDir>
```
dbentley: I think we should consider making this the command-line tool named "scoot". Just scoot. It's how we expect users to use it...

daemon/protocol/python/scoot/daemon_cli.py
```diff
@@ -0,0 +1,156 @@
...
+import sys
+import re
+from scoot.docopt import DocoptExit
+
+sys.path.append(os.path.join(os.path.dirname(__file__), '../../protocol'))
+import client_lib as proto
+
+def display_statuses(statuses):
+    for status in statuses:
+        print("\n\{RunId:{0},\n State:{1},\n ExitCode:{2},\n {Error:{3},\n Snapshot:{5}\n\}".format(status.runId, proto.ScootStatus.State(status.state), status.ExitCode, status.Error, status.Snapshot))
```
dgassaway: Snapshot: {5} pointing at 6th arg, but only 5 args passed to format. The indices in format are optional, I actually prefer without to avoid this kind of thing, i.e.:
print '{} - {} - {}'.format(first, second, third)
The parens around the print string are unnecessary also - print '<stuff>' vs print('<stuff>')

daemon/protocol/python/scoot/daemon_cli.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+if __name__ == '__main__':
+    try:
+        cmd = docopt.docopt(__doc__)
+    except Exception as e:
```
dgassaway: This try/except doesn't really do anything. Catching all Exceptions is usually a bad sign. If we aren't catching anything specific and are just re-raising, we should just let this explode

daemon/protocol/python/scoot/daemon_cli.py
```diff
@@ -0,0 +1,156 @@
...
+        print("exception:{0}".format(e))
+        raise docopt.DocoptExit(e)
+    try:
+#         print("starting client")
+        proto.start()
+    except proto.ScootException as e:
+        if re.search("UNAVAILABLE", e) > 0:
+            raise docopt.DocoptExit("Scoot Daemon is not running!\n")
+        raise docopt.DocoptExit(e)
```
dgassaway: Again not sure why we would re-raise as a docopt error.  I would think DocoptExit would be for command parsing errors, this is a scoot error

daemon/protocol/python/scoot/daemon_cli.py
JeanetteBruno: perhaps we should have a quick huddle to review the current command line and discuss what we want as final state.

daemon/protocol/python/scoot/daemon_cli.py
JeanetteBruno: done

daemon/protocol/python/tests/daemon_cli_test.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python
+
+import os
+import re
+import shutil
+import subprocess
+import tempfile
+import time
+import unittest
+from Carbon.Aliases import false
```
dgassaway: What is this for?

daemon/protocol/python/scoot/daemon_cli.py
JeanetteBruno: done

daemon/protocol/python/tests/daemon_cli_test.py
```diff
@@ -0,0 +1,120 @@
...
+cliRef = ['python', pypath+'/daemon_cli.py']
+createS = ['snapshot', 'create']
+checkoutS = ['snapshot', 'checkout']
+runE = ['exec', 'run']
+pollE = ['exec', 'poll']
+
+echo = ['echo']
+
+class TestCliCommands(unittest.TestCase):
+    daemonProcess = None
```
dgassaway: nit: vars in this file are goStyle not python_style

daemon/protocol/python/tests/daemon_cli_test.py
```diff
@@ -0,0 +1,120 @@
...
+    pypath = ''
+      
+    def setUp(self):
+        self.tmpdir = tempfile.mkdtemp()
+
+        # Note: the following does not work - the process from the pool does not have GOPATH defined so it can't find the binary
+        try:
+            self.daemonProcess = subprocess.Popen(['{0}/bin/daemon'.format(gopath), '-execer_type', 'os'])
+        except Exception as e:
```
dgassaway: What specific exception are we trying to catch here? Catch-alls are unpythonic, if we should make it specific like OSError, or ditch the try block altogether.

daemon/protocol/python/tests/daemon_cli_test.py
```diff
@@ -0,0 +1,120 @@
...
+            self.fail('Fail:{0}'.format(e))  
+        started = false
+        start = time.time()
+        elapsedTime = 0
+        cmd = cliRef + echo + ['ping']
```
dgassaway: Weird mix of list vars and local definitions (echo vs ['ping']), suggest normalizing these and making var names more descriptive (command_args, echo_args, etc)

daemon/protocol/python/tests/daemon_cli_test.py
```diff
@@ -0,0 +1,120 @@
...
+        while not started and elapsedTime < 4.0:
+            try:    
+                r = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+                if self.verifyOut(r, 'ping', False):
+                    started = True
+            except Exception as e:
```
dgassaway: Guessing for all of these we are looking for exception subprocess.CalledProcessError

daemon/protocol/python/scoot/daemon_cli.py
```diff
@@ -0,0 +1,156 @@
...
+        except Exception as e1:
+            raise docopt.DocoptExit(e1)
+        
+    elif cmd['poll']:
+        wait = cmd["--wait"]
+        if not wait:
+            wait = 0
+        runIds = cmd["<runIds>"]
+        if not cmd["<runIds>"]:
+            raise docopt.DocoptExit("RunId list is empty!")
```
dgassaway: We have an awful lot of code here devoted to manual parsing and checking input options.  If there is not a way to handle the subcommand and required parameters automatically w/ docopt I am going to suggest using the standard argparse library.

daemon/protocol/python/scoot/scoot.py
```diff
@@ -0,0 +1,199 @@
...
+    raise docopt.DocoptExit("error parsing command: '{0}'.".format(str(e)))
+  
+  # make the client connection
+  try:
+    proto.start()
+  #handle errors making the client connection
+  except proto.ScootException as e:
+    if re.search("UNAVAILABLE", e) > 0:
+      raise docopt.DocoptExit("Scoot Daemon is not running!\n")
+    raise docopt.DocoptExit("connecting to daemon error: '{0}' (make sure you have started the daemon).".format(str(e)))
```
dgassaway: Again not sure why we would re-raise as a docopt error. I would think DocoptExit would be for command parsing errors, this is a scoot error

daemon/protocol/python/scoot/scoot.py
```diff
@@ -0,0 +1,199 @@
...
+    if re.search("Not started", str(e)) > 0 or re.search("UNAVAILABLE", str(e)) > 0:
+      raise docopt.DocoptExit("Scoot Daemon is not running!\n")
+    raise docopt.DocoptExit("echo request error:'{0}'".format(str(e)))
+
+
+if __name__ == '__main__':
+  # parse the command line
+  try:
+    cmd = docopt.docopt(__doc__)
+  except Exception as e:
```
dgassaway: This try/except doesn't really do anything. Catching all Exceptions is usually a bad sign. If we aren't catching anything specific and are just re-raising, we should just let this explode

daemon/protocol/python/scoot/scoot.py
```diff
@@ -0,0 +1,199 @@
...
+  wait = cmd["--wait"]
+  if not wait:
+    wait = 0
+  try:
+    wait = int(wait)
+  except Exception as e:
+    raise docopt.DocoptExit("Wait must be an integer.")
+  runIds = cmd["<runId>"]
+  if not cmd["<runId>"]:
+    raise docopt.DocoptExit("RunId list is empty!")
```
dgassaway: We have an awful lot of code here devoted to manual parsing and checking input options, types, etc. If there is not a way to handle the subcommand and required parameters automatically w/ docopt I am going to suggest using the standard argparse library.

daemon/protocol/python/tests/daemon_cli_test.py
```diff
@@ -0,0 +1,157 @@
...
+  daemonProcess = None
+  pypath = ''
+     
+  def setUp(self):
+    self.tmpdir = tempfile.mkdtemp()
+
+    # Note: the following does not work - the process from the pool does not have GOPATH defined so it can't find the binary
+    try:
+      self.daemonProcess = subprocess.Popen(['{0}/bin/daemon'.format(gopath), '-execer_type', 'os'])
+    except Exception as e:
```
dgassaway: What specific exception are we trying to catch here? Catching all Exceptions is unpythonic, if we should make it specific like OSError, or ditch the try block altogether.

daemon/protocol/python/tests/daemon_cli_test.py
```diff
@@ -0,0 +1,157 @@
...
+      self.fail('Fail:{0}'.format(e))  
+    started = False
+    start = time.time()
+    elapsedTime = 0
+    cmd = cliRef + echo + ['ping']
```
dgassaway: Weird mix of list vars and local definitions (echo vs ['ping']), suggest normalizing these and making var names more descriptive (command_args, echo_args, etc)

daemon/protocol/python/tests/daemon_cli_test.py
```diff
@@ -0,0 +1,157 @@
...
+    while not started and elapsedTime < 4.0:
+      try:    
+        r = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+        if self.verifyOut(r, 'ping', False):
+          started = True
+      except Exception as e:
```
dgassaway: Guessing for all of these we are looking for exception subprocess.CalledProcessError

daemon/protocol/python/tests/daemon_cli_test.py
```diff
@@ -0,0 +1,157 @@
...
+import os
+import re
+import shutil
+import subprocess
+import tempfile
+import time
+import unittest
+
+gopath = os.environ['GOPATH']
+pypath = gopath + '/src/github.com/scootdev/scoot/daemon/protocol/python/scoot'
```
jasonbs10: We need to handle the case where multiple paths are given. I think the default behavior is to take the first path (where paths are delimited by ':') and drop everything else.

daemon/protocol/python/tests/daemon_cli_test.py
```diff
@@ -0,0 +1,157 @@
...
+cliRef = ['python', pypath + '/scoot.py']
```
jasonbs10: cliScript might be easier to read.

daemon/protocol/python/tests/daemon_cli_test.py
```diff
@@ -0,0 +1,157 @@
...
+      cmd = cliRef + createS + [self.tmpdir]
+      r = subprocess.check_output(cmd)
+      self.verifyOut(r, 'snapshot id = [0-9]+', True)
+      m = re.findall(r'[0-9]+', r)
+      sId = m[0]
+      
+      # issue createSnapshot without a path validate get error message and usage prompt 
+      cmd = cliRef + checkoutS + [sId, self.tmpdir]
+      r = subprocess.check_output(cmd).rstrip()
+      self.assertTrue(r == '', "Error: expected '' from checkout, got '{0}'".format(r))
```
jasonbs10: Is this msg correct? It doesn't seem to match with the comment above.

daemon/protocol/python/tests/daemon_cli_test.py
```diff
@@ -0,0 +1,157 @@
...
+       
+  def assert_invalid_command(self, cmd):
+    try:
+      r = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+    except Exception as e:
+      return
+    self.fail('{0} did not return an error.'.format(*cmd))
+
+        
+  def test_cmdErrors(self):
```
jasonbs10: I think these are less important if using argparse, but they don't hurt.

daemon/protocol/python/tests/daemon_cli_test.py
JeanetteBruno: changed to CalledProcessError

daemon/protocol/python/tests/daemon_cli_test.py
JeanetteBruno: did some renaming and moved the variables inside the class.

daemon/protocol/python/tests/daemon_cli_test.py
JeanetteBruno: yes - changed

daemon/protocol/python/scoot/scoot.py
```diff
@@ -0,0 +1,199 @@
...
+import sys
+import re
+
+sys.path.append(os.path.join(os.path.dirname(__file__), '../../protocol'))
+import client_lib as proto
+
+
+def display_statuses(statuses):
+  for status in statuses:
+    print("\nRunId:{0},\n\tState:{1},\n\tExitCode:{2},\n\tError:{3},\n\tSnapshot:{4}\n".format(status.run_id, proto.display_state(status.state), status.exit_code, status.error, status.snapshot_id))
```
JeanetteBruno: changed to {}.

daemon/protocol/python/tests/daemon_cli_test.py
JeanetteBruno: done

daemon/protocol/python/tests/daemon_cli_test.py
JeanetteBruno: changed to cliPath

daemon/protocol/python/tests/daemon_cli_test.py
JeanetteBruno: yes - checkout doesn't display any response to the user (they should be able to see values in the destDir).  We can change this to display a message such as 'Snapshot copied to <destDir>'.  Do we want to do this?

daemon/protocol/python/tests/daemon_cli_test.py
JeanetteBruno: I'm going to leave them in.

daemon/protocol/python/scoot/scoot.py
JeanetteBruno: turns out none of this code was necessary - docopt parsing already had caught these errors.
Removed the code.

daemon/protocol/python/scoot/scoot.py
```diff
@@ -0,0 +1,181 @@
...
+                          0:  return immediately with the status(s) of the runs.  
+                          >0: wait up to <wait>(seconds) for at least one of the runs to finish.  
+                          Default: 0.
+  --all                   Return the status of all the runs not just finished runs.
+
+"""
+import docopt
+import os
+import sys
+import re
```
dgassaway: Just noticing this now, but we don't need the "re" module at all, we should remove it and replace all of the re.search statements with:
`if '<substring>' in str(e) or '<other_substring>' in str(e):`

dgassaway:
Since we got rid of most of the manual parsing I am fine with the way docopt is working, but if we're still going to keep it we should remove the real docopt.py module and include the dependency in the client setup.py (assuming the cli and lib will be in the same distribution).

daemon/protocol/python/scoot/scoot.py
JeanetteBruno: done

PR #172, opened 2016-11-18
[scheduler] Caitie/sagalogconfig

CaitieM20:
we want to be able to configure the SagaLog to be in memory or potentially a durable option.  Adding config options to make this easy to specify.

I also refactored scootapi/setup.go to explicitly specify all inputs and outputs into MagicBag like we did with worker server, also added all dependencies.  So no additions are necessary but all options can be overridden.


config/scootconfig/modules.go
```diff
@@ -15,12 +15,26 @@ import (
...
 	"github.com/scootdev/scoot/ice"
+	"github.com/scootdev/scoot/saga"
+	"github.com/scootdev/scoot/saga/sagalogs"
 	"github.com/scootdev/scoot/sched/scheduler"
 	"github.com/scootdev/scoot/sched/worker"
 	"github.com/scootdev/scoot/sched/worker/workers"
 	"github.com/scootdev/scoot/workerapi/client"
 )
 
+type InMemorySagaLogConfig struct {
```
dgassaway: Godoc comment!

PR #174, opened 2016-11-22
snapshot/git/gitfiler: Fetch repo when checkout fails

ryancouto:
Attempt to fetch repo when first pass clean/checkout fails


snapshot/git/gitfiler/checkouter.go
```diff
@@ -49,12 +47,25 @@ func (c *Checkouter) Checkout(id string) (co snapshot.Checkout, err error) {
 		{"checkout", id},
 	}
 
+	if err := c.runGitCmds(cmds, repo); err != nil {
+		// try fetching for new commits before returning error
+		// takes a long time (~5 min)
```
dbentley: Eep! 5 minutes is long... Do you have a breakdown? Does it take that long if you ssh in?

ryancouto: Yeah, same story ssh'd in

PR #176, opened 2016-12-05
Provides a way to access runner output from an http endpoint.

jasonbs10:
Provides a way to access runner output from an http endpoint without having to wait for a run to finish. This change reduces the need for custom OutputCreators and meshes with the eventual addition of snapshot-persisted output. There are a couple of TODOs but this is largely complete and shouldn't break anything.

The motivation here is that scoot clients are not provided with task stdout/stderr URIs until the task has finished, and we have no way of automating the retrieval of task logs prior to that time. An additional and separate desire is to facilitate viewing/tailing logs and viewing metrics in a browser. Persistence would be nice but we'll forgo that temporarily until blobstore bundles with stdout/stderr are implemented.

The first attempt to address persistence, in source, was to create an hdfs OutputCreator. That works fine but will not scale like we want in production. Further, it doesn't address the missing URIs before a task is finished, or viewing in a browser.

This PR does the minimal amount of plumbing required to send http URIs back to the client so it can start streaming logs immediately. The general flow is:

- Instead of scheduler doing log.StartTask & worker.RunAndWait, do log.StartTask & worker.Start & log.StartTask(status) & worker.Wait
- Runners must initialize and return StdoutRef/StderrRef so worker.Start sees those immediately.
- This allows scheduler to sagalog a trackable run. The client can poll based on sagalog contents.

- Worker: pass an endpoints.ResourceHandler to OutputCreator so callers can call Output.Register
- Worker: runner calls Output.Register with globally defined Stdout/Stderr http paths, and namespaced under RunId.
-Worker: Output.Register will create a new http endpoint which will query the log file directly.

Corresponds with go/phab/D14557


sched/scheduler/task_runner.go
```diff
@@ -18,44 +19,85 @@ const DeadLetterExitCode = -200
...
+
+	// Handle errors the same way for Start()/Wait().
+	workerErrStatus := func(err error) error {
+		if err != nil {
+			stat.Counter("failedTaskRunCounter").Inc(1)
+
+			// if we should mark tasks complete even if an error occurs set processStatus
+			// and don't return.  Otherwise return the error and the task will
+			// get rescheduled.
+			if markCompleteOnFailure {
```
JeanetteBruno: how does markCompleteOnFailure get set?

sched/scheduler/task_runner.go
```diff
@@ -18,44 +19,85 @@ const DeadLetterExitCode = -200
...
+	// Start task on the worker
+	// If err, check markCompletedOnFailure. If true, log and continue. If false, return err immediately.
+	var startErr error
+	processStatus, startErr = worker.Start(task)
+	if workerErrStatus(startErr) != nil {
+		return startErr
+	}
+
+	// Log the preliminary ProcessStatus.
+	if statusUrl != "" {
```
JeanetteBruno: how/where is statusUrl set?

sched/scheduler/stateful_scheduler_test.go
```diff
@@ -170,7 +175,7 @@ func Test_StatefulScheduler_TaskGetsMarkedCompletedAfterMaxRetries(t *testing.T)
 		t.Fatalf("Expected Tries: %v times, Actual Tries: %v", deps.config.MaxRetriesPerTask+1, s.inProgressJobs[jobId].Tasks[taskId].NumTimesTried)
 	}
 
-	// advance scheduler until job gets marked completed
+	// // advance scheduler until job gets marked completed
```
ryancouto: extra //

sched/worker/workers/polling.go
```diff
@@ -41,29 +41,38 @@ type PollingWorker struct {
...
+	timeSpent := 0 * time.Second
 	for !r.enforceTimeout || (r.enforceTimeout && timeSpent < r.timeout) {
 		time.Sleep(r.pollingPeriod)
 		timeSpent += r.pollingPeriod
 
-		status, err = r.runner.Status(id)
+		status, err = r.runner.Status(runId)
 		if err != nil || status.State.IsDone() {
-			return status, err
+			return
```
ryancouto: i like the named return values here, but it's a little weird with Start, Status, and RunAndWait using explicit returns.

dbentley:
I like the approach, overall.

Comments, from highest-level to lowest-level:

This proposal seems to be about streaming/debugging worker results, not about persisting them. Is that right? It seems like a good and worthy thing, but slightly different from persisting worker results (as this mentions). (I want to be explicit about the change, or see if I'm mistaken)

Scheduler: I believe the Saga Log pattern means we have to log our intention to do something *before* we do it. Otherwise, we could die after starting the task but before logging it. If the task needed a compensating action, we wouldn't know to run it. This might require extending our SagaLog to allow you to log a message with an info message.

Scheduler/Worker protocol: AIUI, the Scheduler can now contact the worker to ask for its base URI. It uses this to populate a link to the log viewer URI into the status. Could we just have the Worker populate the URI and have runTaskAndLog wait until the URIs are populated? The Worker server knows its addr, so it can populate the ProcessStatus correctly, no? This would remove the need to extend the Worker API.

(This comment might be unnecessary if we don't extend the Worker API.) The extension to the Worker API is done in this weird way by creating a certain kind of bad request. I'd rather be explicit, and add a method to the thrift definition if we need one.

Worker server: Could we just make a server that serves a specific directory? Then we use a new TempDir to store these output files, and creating the URI is as easy as http://<addr>/output/<id>

common/endpoints/endpoints.go
```diff
@@ -64,3 +77,97 @@ func MakeStatsReceiver(scope StatScope) stats.StatsReceiver {
...
+const StderrName = "stderr"
+
+func NewResourceHandler(uri string, ajaxInterval time.Duration) *ResourceHandler {
+	hostname, _ := os.Hostname()
+	return &ResourceHandler{resources: make(map[string]map[string]string), hostname: hostname, uri: uri, ajaxInterval: ajaxInterval}
+}
+
+type ResourceHandler struct {
+	// Defines: map[Namespace]map[ResourceName]ResourcePath
+	resources    map[string]map[string]string
```
dbentley: Some comments on the resource handler would be good here... Is it just to serve the JS?

runner/local/output.go
```diff
@@ -60,6 +65,13 @@ func (o *localOutput) Close() error {
 	return o.f.Close()
 }
 
+func (o *localOutput) Register(namespace, name string) (uri string) {
+	if o.handler != nil {
+		o.uri = o.handler.AddResource(namespace, name, o.AsFile())
```
dbentley: Why do we need to add the resource? It seems like it would just end up in the temp dir and so the link would work...

runner/output.go
```diff
@@ -26,4 +26,9 @@ type Output interface {
 
 	// Copies (if necessary) the URI target to local file and returns that absolute file path.
 	AsFile() string
+
+	// Add this Output to a list, probably an endpoints.ResourceHandler.
+	// Multiple names may be registered under a single namespace.
+	// Returns the uri resulting from this registration.
+	Register(namespace, name string) (uri string)
```
dbentley: What's the goal of putting it in a namespace? As opposed to just having it get a unique, opaque URI?

sched/scheduler/task_runner.go
```diff
@@ -29,33 +31,65 @@ func runTaskAndLog(
 	task sched.TaskDefinition,
 	markCompleteOnFailure bool,
 	stat stats.StatsReceiver) error {
-	// Log StartTask Message to SagaLog
-	err := saga.StartTask(taskId, nil)
+
+	// Start Task and then log StartTask message a status payload containing stdout/stderr URIs.
+	// Note: this ordering will result in an orphaned worker if logging fails.
+	//       also, URIs will likely change upon task completion, i.e. from local file to snapshot URI.
+	// TODO: !!this no longer works if we ever start allowing compensating actions on failure!!
```
dbentley: We should check with Caitie on this. I'd prefer to not compromise the model. It would probably be easy to add a new message. WDYT?

jasonbs10: Yeah, we should let Caitie decide. I'm ok either way.

common/endpoints/endpoints.go
jasonbs10: Yep, this needs more comments (pending the outcome of this review).

runner/local/output.go
dbentley: I think it'd be much simpler to say "everything in this tempdir is shared" and that's a good way of specifying intent. If we need more complexity, I'd add it later.

runner/output.go
dbentley: Could you help me understand with examples?

I think it'd be nice to have less coupling between these structures. I'd hope that output.go doesn't have to know about ResourceHandler *at all*.

common/endpoints/endpoints.go
```diff
@@ -8,35 +8,48 @@ import (
...
 }
 
 func (s *TwitterServer) Serve() error {
-	mux := http.NewServeMux()
-	mux.HandleFunc("/", healthHandler)
-	mux.HandleFunc("/admin/metrics.json", s.statsHandler)
-	server := &http.Server{
-		Addr:    s.Addr,
-		Handler: mux,
+	http.HandleFunc("/", helpHandler)
```
dbentley: Err, http.HandleFunc is a global... Could we keep using a specific one?

E.g., for hackweek, I'm adding a second http servemux into a server and it'd be nice to not have to reason about this.

sched/scheduler/task_runner.go
dgassaway: To add my 2c I would want to leave the existing sagalog guarantees in place.

runner/local/output.go
```diff
@@ -35,17 +68,92 @@ func (s *localOutputCreator) Create(id string) (runner.Output, error) {
...
+  setInterval(sendRequest, 2000)
+</script>
+</html>
+`
+	if strings.TrimSuffix(r.URL.Path, "/")+"/" == s.HttpPath() {
+		http.StripPrefix(s.HttpPath(), http.FileServer(http.Dir(s.tmp.Dir))).ServeHTTP(w, r)
+		return
+	}
+	path := strings.TrimPrefix(r.URL.Path, s.HttpPath())
+	filepath, ok := s.pathMap[path]
```
dbentley: I'd just skip the pathMap and just use the path from the URL...

runner/local/output.go
```diff
@@ -35,17 +68,92 @@ func (s *localOutputCreator) Create(id string) (runner.Output, error) {
...
+	if !ok {
+		http.Error(w, "Unrecognized path", http.StatusNotFound)
```
dbentley: https://golang.org/pkg/net/http/#ServeFile

Oh, but you want to do this JS thing to keep reloading, huh?

runner/local/simple.go
```diff
@@ -62,11 +62,28 @@ func (r *simpleRunner) Run(cmd *runner.Command) (runner.ProcessStatus, error) {
 		return runner.ProcessStatus{}, fmt.Errorf(runner.RunnerBusyMsg)
 	}
 
+	// Redirect command output to the stdout/stderr Output vars. Outputs shall be closed in run().
```
dbentley: Moving this is a bit of a layering violation... Is there a reason to do it here? Just to make it simpler to not have to wait?

runner/status.go
```diff
@@ -69,6 +69,7 @@ type ProcessStatus struct {
 
 	State ProcessState
 	// References to stdout and stderr, not their text
+	// Runner impls shall provide valid refs for all States (but optionally may not for UNKNOWN/BADREQUEST).
```
dbentley: What's the point of a stdoutref if we haven't even started running it?

runner/local/output.go
jasonbs10: I'd like to leave this in for now since it's useful as a dev to have a stable path for job output across scheduler/worker restarts.

runner/local/simple.go
jasonbs10: It's easy to do it here and falls under the category of job initialization shared across all jobs of all types. See status.go comment.

runner/status.go
jasonbs10: I view it in the same vein as RunId, I want to provide all the ways to address a run at inception.

runner/local/output.go
jasonbs10: Might as well, it seemed simple enough. I hope other people also think http log tailing is useful...

CaitieM20:
@dbentley is correct we have to log the StartTask message first before performing an action.

The easiest way to extend this right now would be to just log another StartTask message with the links info once we get it, since we dont store anything with StartTask now.  This is totally fine, and in our current durable implementation will just overwrite the additional message!   This way the Status API can get access to the in progress logs as the Task is running

sched/scheduler/task_runner.go
CaitieM20: agree I would recommend leave 
err := saga.StartTask(taskId, nil)

make the call to the work
then make the start call to the worker, and then do 
saga.StartTask(taskId, statusAsBytes) 

Also do we want to log the whole ProcessStatus?   Is all the info useful?  We should be careful what we expose in StartTask since clients may take dependencies on this.  I like limiting what we expose to the bare minimum we need

jasonbs10: Done. I'm going to keep ProcessStatus for now since that mirrors the EndTask payload. We should revisit the pair later in another PR if necessary.

sched/scheduler/task_runner.go
```diff
@@ -29,37 +31,73 @@ func runTaskAndLog(
...
+					`Error Running Task %v: dead lettering task after max retries.
 				TaskDef: %+v, Saga Id: %v, Error: %v`,
-				taskId, task, saga.GetState().SagaId(), err)
-		} else {
-			//TODO: Check error to see if its retryable.
-			return err
+					taskId, task, saga.GetState().SagaId(), err)
+			} else {
+				//TODO: Check error to see if its retryable.
+				return err
```
JeanetteBruno: This value (err) is 'swallowed' by all callers.  Can we remove the else clause or change the callers to use this value?

sched/scheduler/task_runner.go
```diff
@@ -29,37 +31,73 @@ func runTaskAndLog(
...
+	workerErrStatus := func(err error) error {
+		if err != nil {
+			stat.Counter("failedTaskRunCounter").Inc(1)
+
+			// if we should mark tasks complete even if an error occurs set processStatus
+			// and don't return.  Otherwise return the error and the task will
+			// get rescheduled.
+			if markCompleteOnFailure {
+				processStatus.Error = err.Error()
+				processStatus.ExitCode = DeadLetterExitCode
```
JeanetteBruno: why is this called 'DeadLetterExitCode'?  Suggest 'FailureExitCode' or 'CompleteWithFailure'

sched/scheduler/task_runner.go
jasonbs10: This is Caitie's TODO so I'd rather leave this to her.

sched/scheduler/task_runner.go
jasonbs10: Ditto. And imo 'DeadLetter' is a thing and is more descriptive than 'Failure'.

sched/scheduler/task_runner.go
CaitieM20: yea leave that for me, i'll clean it up.  Its ok for now, basically we may retry an error (bad input) that will never succeed multiple times.  It doesn't break anything its just an optimization

PR #178, opened 2016-12-08
Decompose runner.Runner

dbentley:
runner.Runner was unwieldy and hard to implement/understand/use. Break it down.

This implements the API from https://github.com/scootdev/scoot/pull/167 (and fixes all references to new names)

Names that could be better:
  Statuses
  ServiceFacade


config/scootconfig/modules.go
```diff
@@ -129,7 +129,7 @@ func (c *WorkersThriftConfig) Create(
 		di := dialer.NewSimpleDialer(tf, pf)
 		cl, _ := client.NewSimpleClient(di, string(node.Id()))
 		q := runners.NewPollingService(cl, cl, cl, pollingPeriod)
-		return workers.NewServiceWorker(q, timeout, defaultOverhead)
+		return workers.NewServiceWorker(q, taskTimeout, defaultOverhead)
```
jasonbs10: Not necessarily the PR for it, but: what is the right place to clarify that a task has to reach a PREPARING/RUNNING state before taskTimeout is started? When we use timeout in smoketest, iirc, we expect a task to be canceled regardless of queued status. Should we add a way to specify when timeout starts ticking?

runner/runners/statuses.go
```diff
@@ -0,0 +1,169 @@
...
+
+// NewRun creates a new RunID in state Preparing
+func (s *Statuses) NewRun() (runner.RunStatus, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	id := runner.RunID(strconv.FormatInt(s.nextRunID, 10))
+	s.nextRunID++
+
+	st := runner.RunStatus{
```
jasonbs10: Reminder, if PR-176 is accepted in its current state, we will need to initialize StdoutRef/StderrRef here.

runner/runners/facade.go
```diff
@@ -0,0 +1,13 @@
+package runners
+
+import (
+	"github.com/scootdev/scoot/runner"
+)
+
+// ServiceFacade makes a Service from component parts.
```
dgassaway: Why not just RunnerService? Facade implies testing usage to me (as does file name)

runner/runners/facade.go
```diff
@@ -0,0 +1,13 @@
+package runners
```
dgassaway: Can we put a brief comment header on all of the runner/runners/*.go implementation and other files

runner/README.md
```diff
@@ -1,11 +1,57 @@
...
+## Statuses interfaces (runner/statuses.go) ##
+There are several interfaces having to do with reading/writing RunStatus'es.
+
+StatusWriter for writing Statuses.
+
+There are more reader interfaces:
+* StatusReader includes all the ones below
+* LegacyStatusReader offers legacy APIs
+* StatusQuerier offers Query, which allows waiting for some period of time
+* StatusQuerierNow offers QueryNow which is the same as Query, but does not allow waiting
```
dgassaway: Update readme interface names (e.g. StatusQuerierNow vs StatusQueryNower)

runner/README.md
```diff
@@ -1,11 +1,57 @@
...
+* tenancy (run one at a time? Or multiple? Fixed, or based on system utilization?)
+* queueing (should we accept work we can't start right now)
 
-There will be an Execer that calls os/exec to exec.
\ No newline at end of file
+When a Controller wants to start a Run of a Command, it will use an Invoker.
+
+For now we have one implementation, QueueController
+
+## Statuses interfaces (runner/statuses.go) ##
```
dgassaway: Suggestions - StatusManager, StatusReaderWriter

runner/README.md
```diff
@@ -1,11 +1,57 @@
...
+There are several interfaces having to do with reading/writing RunStatus'es.
+
+StatusWriter for writing Statuses.
+
+There are more reader interfaces:
+* StatusReader includes all the ones below
+* LegacyStatusReader offers legacy APIs
```
dgassaway: Somewhat off topic, but I think we should be very aggressive in deleting all vestigial traces of legacy APIs quickly. Cleaning this out will never be easier than it is now.  I don't remember exactly what this is supporting - closed source binaries?

config/scootconfig/modules.go
dbentley: Ugh, yeah, we should figure that out. Added a TODO in runner.go

runner/runners/statuses.go
dbentley: Commented in task_runner.go that we need to reenable this.

(As discussed with Jason off-line, turning this off for now to make this PR easier to land)

runner/runners/facade.go
dbentley: Added a bunch of comments at the top of files whose purpose is non-obvious; wdyt?

runner/runners/facade.go
dbentley: Renamed to Service.

runner/README.md
dbentley: Renamed Statuses to StatusManager

runner/README.md
dbentley: Our Thrift API is still in terms of all Statuses and Erase'ing. Added a comment to worker.thrift.

runner/README.md
dbentley: Done.

runner/runners/facade.go
dgassaway: Lgtm.  Last thing mv facade.go service.go?

dbentley: Had already done, but had forgotten to push. Now pushed.

PR #179, opened 2016-12-19
[Scheduler] saga recovery

CaitieM20:
Adding Saga Recovery to the Scheduler, if recovery is configured then active/in-progress jobs  will be recovered from the SagaLog and rescheduled.  Currently the scheduler blocks until all jobs can be recovered.  

Currently recovery is turned off for the deployable binary which uses NewStatefulSchedulerFromCluster(...) to create the scheduler.  A follow up code review will enable recovery and make this a configurable parameter.  

This change is a bit large so I added a detailed breakdown of what was added in each file.  The bulk of the code is in recover_jobs.go.

Metrics will also be added in the next pass.

sagacoordinator.go 
* fix sagaCoord bug, if the Saga doesn't exist return nil.  Added test case

definitions.go 
* Added Serialize and DeserializeJob methods to sched definitions.go for convenience
* added property based tests to validate

job_state.go 
  * converted scheduler.jobState to have a pointer to Job, less copying data and more consistent use of object
* newJobState no longer assumes that the job is brand new.  Now has logic to recover state of completed task state from the associated saga.
* additional test cases to verify

recover_jobs.go 
* contains the recovery logic for sagas at startup.  RecoverJobs queries for active sagas, and then recovers them in parallel.  
* If failures occur and are retry-able exponential backoff is used between requests to avoid slamming the SagaLog with requests.  
* lots of test cases

stateful_scheduler.go 
* Make recovering jobs configurable via constructor parameter.
* On startup initiate recovery and block until all jobs are recovered. 
* Added serialized job to StartSaga Message at Job  creation time


sched/scheduler/recover_jobs.go
```diff
@@ -0,0 +1,131 @@
...
+	// TODO: limit max parallel requests to RecoverSagaState?
+	for _, sagaId := range activeSagas {
+
+		go func(sagaId string) {
+			defer wg.Done()
+			activeSaga := recoverSaga(sc, sagaId)
+			if activeSaga != nil {
+				job, err := sched.DeserializeJob(activeSaga.GetState().Job())
+				if err != nil {
+					// TODO: Increment counter? A breaking change was made
```
jasonbs10: This might be a panic for now, or an expanded TODO. We can't drop this job until we have a mechanism to inform clients that their active job will never be restarted.

sched/scheduler/stateful_scheduler.go
```diff
@@ -71,6 +75,7 @@ func NewStatefulScheduler(
 	config SchedulerConfig,
 	stat stats.StatsReceiver,
 	debugMode bool,
+	recoverJobs bool,
```
jasonbs10: nit: I'd prefer not to shadow function names with variable names.

sched/scheduler/recover_jobs.go
CaitieM20: I dont want it to panic because it'll crash the server, and negatively impact everyone if only one job gets messed up.

It also won't get removed from the log, so it'll get retried at next server start up.  Ideally this should fire some kind of critical alert, because it'll require manual intervention.

sched/scheduler/stateful_scheduler.go
CaitieM20: changed to recoverJobsOnStartup

PR #180, opened 2016-12-20
[Scheduler] Specify Required fields in  job_def.thrift

CaitieM20:
Currently the job_def.thrift structs did not specify required or optional, so they all default to optional.  However, we do have required fields, and our code was making assumptions about some fields being required.  

This caused problems and panics to be thrown instead of errors returned on deserialization of improperly formatted thrift structs.  

This pull request explicitly defines if fields are required or not, and has additional test cases.  

It also contains refactors, notably
* makeDomainJobFromThriftJob & makeThriftJobFromDomainJob moved to definitions.go instead of generators.go.  This is core logic and not just code used for property based testing. 
* ValidateSerialization returns a bool, instead of invoking t.Errorf, this caused property tests to "pass" and throw a test error which is not the normal execution path.  You miss out on some debugging info by having the property always return true.  
* moved all generator code into generators.go, and all jobs are now more realistic and randomly generated.


sched/generators.go
```diff
@@ -2,15 +2,28 @@ package sched
...
 	"github.com/leanovate/gopter"
 	"github.com/scootdev/scoot/runner"
-	"github.com/scootdev/scoot/sched/gen-go/schedthrift"
 	"github.com/scootdev/scoot/tests/testhelpers"
-	"time"
 )
 
 func GenJob(id string, numTasks int) Job {
-	jobDef := GenJobDef(numTasks)
+	rand := testhelpers.NewRand()
```
jasonbs10: The naming of GenJob and GenRandomJob is a little confusing at first. What's the difference between NewRand() and genParams.Rng and can these funcs be combined?

sched/job_def.thrift
```diff
@@ -9,26 +9,26 @@
...
 
 struct TaskDefinition {
-  1: Command command,
+  1: required Command command,
 }
 
 struct JobDefinition {
-  1: string jobType,
-  2: map<string, TaskDefinition> tasks,
+  1: optional string jobType,
```
jasonbs10: Shouldn't most structs have at least one 'required' field? (tasks in this case)

sched/job_def_serializer_property_test.go
```diff
@@ -15,14 +16,32 @@ func Test_RandomSerializerDeserializer(t *testing.T) {
 
 	properties.Property("Serialize JobDef", prop.ForAll(
 		func(job *Job) bool {
-			ValidateSerialization(job, false, t)
-			ValidateSerialization(job, true, t)
+			serializeBinary := ValidateSerialization(job, false)
```
jasonbs10: nit: there are a few places in this PR where you can combine 'if' and the line doing assignment.

sched/job_def.thrift
CaitieM20: They don't have to.  In theory you could send us a job with no tasks, it'll work fine, just nothing will be ran,  Also we don't currently use jobType, but may in the future?

sched/job_def_serializer_property_test.go
CaitieM20: cool fixed

PR #182, opened 2016-12-21
scootapi/setup: improve waiting and killing

dbentley:
Problem: if you start a command with Cmds, and it then starts another command, Cmds.Kill() will not kill it. This is especially problematic if the command it starts is a server that binds to a port.

Solution: Cmds.Kill now sends the signal to the entire process group. It also starts processes in a new process group.

(We weren't hitting this in open source, but were internally, where our saga log implementation is a server on a port)


dbentley:
I noticed that some (internal) commands weren't nicely dying, because they were spawning processes. This switches us to sending signals to process groups instead of processes.

dbentley:
Ugh, nope, this won't work with this approach. NM.

jasonbs10:
Can you remind me how user commands like 'smoketest' exit cleanly? I can't seem to find any killing of the 'setup' stuff after a clean cmd exit...

What was broken about this?

dbentley:
Updated the PR description; how's that?

scootapi/setup/cmds.go
```diff
@@ -44,26 +48,70 @@ func NewSignalHandlingCmds(tmp *temp.TempDir) *Cmds {
...
+	}
+	c.mu.Lock()
+
+	if len(c.watching) == 0 {
+		return
+	}
+
+	time.Sleep(10 * time.Second)
+
+	// First, send SIGINT to each
```
dgassaway: comment - SIGKILL

scootapi/setup/cmds.go
```diff
@@ -44,26 +48,70 @@ func NewSignalHandlingCmds(tmp *temp.TempDir) *Cmds {
...
+	case <-time.After(5 * time.Second):
+		log.Printf("Still waiting; killing all")
+	}
+	c.mu.Lock()
+
+	if len(c.watching) == 0 {
+		return
+	}
+
+	time.Sleep(10 * time.Second)
```
dgassaway: Shouldn't need to sleep here again? Or change time.After() in select above to reflect actual patience?

dbentley: Thx; fixed.

PR #183, opened 2016-12-27
Caitie/recovery config

CaitieM20:
Make RecoverJobsOnStartup and Debug Mode configurable.


binaries/scheduler/config/local.local
```diff
@@ -10,7 +10,9 @@
   },
   "SchedulerConfig": {
     "Type": "stateful",
-    "MaxRetriesPerTask" : 2
+    "MaxRetriesPerTask" : 2,
+    "DebugMode" : false,
+    "RecoverJobsOnStartup" : false
```
JeanetteBruno: can you put a comma at the end of this line? - then when we add a new configuration, we won't have to add a comma to the existing one.

binaries/scheduler/config/local.memory
```diff
@@ -5,7 +5,9 @@
   },
   "SchedulerConfig": {
     "Type": "stateful",
-    "MaxRetriesPerTask" : 2
+    "MaxRetriesPerTask" : 2,
+    "DebugMode" : false,
+    "RecoverJobsOnStartup" : false
```
JeanetteBruno: comma at end of line (see prior comment)

sched/scheduler/stateful_scheduler.go
```diff
@@ -92,11 +105,11 @@ func NewStatefulScheduler(
 
 	// TODO: we need to allow the scheduler to accept new jobs
 	// while recovering old ones.
-	if recoverJobsOnStartup {
+	if config.RecoverJobsOnStartup {
 		sched.startUp()
 	}
 
-	if !debugMode {
+	if !config.DebugMode {
```
JeanetteBruno: exposing a debug mode flag in the config for automated (non-user interactive) testing seems a bit off to me....  if we need automated testing to have control over the progress of the scheduler loop couldn't we have injected instrumented workers where the test control's when a worker 'finishes'.  That way we don't have to put the sched.loop() in an 'if !debug mode {}' clause (change the code behavior for debug vs non-debug).... if we plan to expose the debug feature to the user then step() needs to be Step(), Step() should be in the API and the ScheduleJob() request would have the debug flag passed in on the request.

CaitieM20: So the idea of having it in config, is we can debug real scheduler instances in prod.  We just need to implement an api, that advances the loop.  This way if we run into some nasty bug in production, we can start it up in Debug mode and Advance one step at a time to figure out what's happening.  [DX-1645](https://jira.twitter.biz/browse/DX-1645) tracks this work

CaitieM20: Debug mode needs to be implemented for the whole Scheduler and not just per job, because it controls the execution of the Scheduler's inner loop.  The only difference is if this advances automatically or if we do it manually for debugging

binaries/scheduler/config/local.memory
CaitieM20: The JSON parser actually throws an error if there is a comma on the last element

"Error Parsing Config File local.local, with Error Couldn't parse top-level config: invalid character '}' looking for beginning of object key string"

binaries/scheduler/config/local.local
CaitieM20: causes the JSON parser to throw an error

sched/scheduler/stateful_scheduler.go
jasonbs10: This seems like it will be hard to use efficiently in practice. I'm not sure this is useful in production without a mechanism to specify which jobs are to be manually advanced, or a way to block jobs that are spam wrt debugging. So I would say this is fine but incomplete.

dgassaway: I can see how this review works on its own, but not sure how it fits into the context of DX-1645 - if API can/will toggle scheduler debug mode, why have it in config?  Just more flexibility to bring it up in debug mode from the start?

CaitieM20: in case you want to bring scheduler up in Debug mode, otherwise you'd have this weird race.  of bringing it up and then having to send an API command.

CaitieM20: It really doesn't make any progress on DX-1645 other than letting us specify via config, instead of changing code.  

The goal for this review was to make adding config a non breaking and easier change you just add more variables to the config object.

PR #184, opened 2016-12-28
workerapi: offer QueryNow in worker thrift

dbentley:
Expands the Worker Thrift API to also offer the QueryNow functionality from runner.Service.

### DIFF
--- a/workerapi/api_test.go
+++ b/workerapi/api_test.go
@@ -15,13 +15,25 @@ var zero = int32(0)
 var nonzero = int32(12345)
 var emptystr = ""
 var nonemptystr = "abcdef"
+var fortyEight = int64(48)
+var trueVar = true
 
 var cmdFromThrift = func(x interface{}) interface{} { return ThriftRunCommandToDomain(x.(*worker.RunCommand)) }
 var cmdToThrift = func(x interface{}) interface{} { return DomainRunCommandToThrift(x.(*runner.Command)) }
 var rsFromThrift = func(x interface{}) interface{} { return ThriftRunStatusToDomain(x.(*worker.RunStatus)) }
 var rsToThrift = func(x interface{}) interface{} { return DomainRunStatusToThrift(x.(runner.RunStatus)) }
+var rssFromThrift = func(x interface{}) interface{} {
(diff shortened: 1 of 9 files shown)


JeanetteBruno:
What is the design for the worker maintaining a history of task requests and their states?  Don't we have the saga implementation for this information.... I thought the worker api is synchronous, so it only has one active request at a time.  (I think I'm missing something.)

dbentley:
1) The worker will maintain history only as long as it's alive, with the in-memory statuses impl before.
2) Yes, this information wants to be in the saga log. The worker thrift API is just to get the data from the worker to the scheduler.
3) The API is not synchronous. Thrift doesn't like RPC's that last the 40 minutes running on the worker can take.
4) Currently, yes, we only run one at a time, but there's no fundamental reason we couldn't be running multiple runs on one worker at a time.

workerapi/api_test.go
```diff
@@ -137,6 +149,47 @@ var tests = []struct {
...
+		},
+		[]runner.RunStatus{
+			runner.RunStatus{RunID: "a", State: runner.COMPLETE,
+				StdoutRef: "", StderrRef: "", ExitCode: 0, Error: ""},
+			runner.RunStatus{RunID: "b", State: runner.COMPLETE,
+				StdoutRef: "", StderrRef: "", ExitCode: 0, Error: ""},
+		},
+	},
+
+	// RunQuery
```
dgassaway: Looks like thrift struct RunsQuery fields are all optional - worth adding a test case with empty struct to make sure that works as intended (does nothing?)

workerapi/client/client.go
```diff
@@ -134,12 +133,19 @@ func (c *simpleClient) StatusAll() ([]runner.RunStatus, error) {
 	return st.Runs, nil
 }
 
+// Implements Scoot Worker API
 func (c *simpleClient) QueryNow(q runner.Query) ([]runner.RunStatus, error) {
```
JeanetteBruno: While we are in here, can we change the name to be consistent with our other run status retrieval apis (right now we have Status, StatusAll and QueryNow).... can we change QueryNow to FilteredStatus (or something with the word Status in it)

dgassaway:
@dbentley I think this was superseded a while back?

PR #185, opened 2016-12-28
[Sagas] Caitie/file saga log

CaitieM20:
Create a file based saga log.  This is to enable testing of SagaRecovery, currently InMemorySaga does not survive beyond process restart.  File Based SagaLog will be durable beyond process restart , but not beyond machine failure. 

Added CorruptedSagaLog as a Fatal Error.  This should not happen in production unless we did something really wrong (made a breaking change), but allows us to easily detect via test cases if our parsing and reading of the SagaLog files works correctly.

### DIFF
--- a/saga/sagalog.go
+++ b/saga/sagalog.go
@@ -42,16 +42,18 @@ type SagaLog interface {
 // that the data stored in the sagalog for a specified saga
 // is corrupted and unrecoverable.
 type CorruptedSagaLogError struct {
-	s string
+	msg string
+	id  string
 }
 
 func (e CorruptedSagaLogError) Error() string {
-	return e.s
+	return fmt.Sprintf("SagaLog for %v has been corrupted, Error: %v", e.id, e.msg)
 }
 
 func NewCorruptedSagaLogError(sagaId string, msg string) error {
 	return CorruptedSagaLogError{
-		s: fmt.Sprintf("SagaLog for %v has been corrupted, Error: %v", sagaId, msg),
+		msg: msg,
+		id:  sagaId,
 	}
 }
 
--- a/saga/sagalog.go
+++ b/saga/sagalog.go
@@ -2,6 +2,10 @@ package saga
 
 //go:generate mockgen -source=sagalog.go -package=saga -destination=sagalog_mock.go
 
+import (
+	"fmt"
+)
+
 /*
  *  SagaLog Interface, Implemented
  */
@@ -32,6 +36,23 @@ type SagaLog interface {
 	 * Returns an error if it fails.
 	 */
 	GetActiveSagas() ([]string, error)
+}
+
+// CorruptedSagaLogError this is a critical error specifies
+// that the data stored in the sagalog for a specified saga
+// is corrupted and unrecoverable.
+type CorruptedSagaLogError struct {
+	s string
+}
+
+func (e CorruptedSagaLogError) Error() string {
+	return e.s
+}
+
+func NewCorruptedSagaLogError(sagaId string, msg string) error {
+	return CorruptedSagaLogError{
+		s: fmt.Sprintf("SagaLog for %v has been corrupted, Error: %v", sagaId, msg),
+	}
 }
 
 // InvalidRequestError should be returned by the SagaLog
--- a/saga/sagalogs/file_saga_log.go
+++ b/saga/sagalogs/file_saga_log.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path"
 	"time"
 
 	"github.com/scootdev/scoot/saga"
@@ -51,12 +52,8 @@ type fileSagaLog struct {
 // If the directory does not exist it will create it.
 func MakeFileSagaLog(dirName string) (*fileSagaLog, error) {
 
-	if _, err := os.Stat(dirName); err != nil {
-		if os.IsNotExist(err) {
-			os.Mkdir(dirName, os.ModePerm)
-		} else {
-			return nil, err
-		}
+	if err := os.MkdirAll(dirName, os.ModePerm); err != nil {
+		return nil, err
 	}
 
 	return &fileSagaLog{
@@ -67,12 +64,12 @@ func MakeFileSagaLog(dirName string) (*fileSagaLog, error) {
 // all files for a saga log are stored in a directory named
 // by the specified sagaId.
 func (log *fileSagaLog) getSagaDirectory(sagaId string) string {
-	return fmt.Sprintf("%v/%v", log.dirName, sagaId)
+	return path.Join(log.dirName, sagaId)
 }
 
 // Returns the name of the sagalog for the specified file.
 func (log *fileSagaLog) getSagaLogFileName(sagaId string) string {
-	return fmt.Sprintf("%v/%v", log.getSagaDirectory(sagaId), "log")
+	return path.Join(log.getSagaDirectory(sagaId), "log")
 }
 
 // Returns the name of the file to store task data in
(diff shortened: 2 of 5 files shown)


saga/sagalog.go
```diff
@@ -34,6 +38,23 @@ type SagaLog interface {
...
+type CorruptedSagaLogError struct {
+	s string
+}
+
+func (e CorruptedSagaLogError) Error() string {
+	return e.s
+}
+
+func NewCorruptedSagaLogError(sagaId string, msg string) error {
+	return CorruptedSagaLogError{
```
dbentley: Might just want to store sagaID and msg as fields in CorruptedSagaLogError, and then do the Sprintf in the call to Error(). This could make it easier to handle the data programatically?

saga/sagalogs/file_saga_log.go
```diff
@@ -0,0 +1,387 @@
...
+	dirName string
+}
+
+// Creates a FileSagaLog with files stored at the specified directory
+// If the directory does not exist it will create it.
+func MakeFileSagaLog(dirName string) (*fileSagaLog, error) {
+
+	if _, err := os.Stat(dirName); err != nil {
+		if os.IsNotExist(err) {
+			os.Mkdir(dirName, os.ModePerm)
```
dbentley: 1) os.MkdirAll lets you call it unconditionally
2) You should check if Mkdir returns an error

saga/sagalogs/file_saga_log.go
```diff
@@ -0,0 +1,387 @@
...
+
+	return &fileSagaLog{
+		dirName: dirName,
+	}, nil
+}
+
+// all files for a saga log are stored in a directory named
+// by the specified sagaId.
+func (log *fileSagaLog) getSagaDirectory(sagaId string) string {
+	return fmt.Sprintf("%v/%v", log.dirName, sagaId)
```
dbentley: import path, and then you get:

path.Join

saga/sagalogs/file_saga_log.go
```diff
@@ -0,0 +1,387 @@
...
+// This is a very dumb implementation currently just returns
+// a list of all sagas.  This can be improved with an index
+//
+func (log *fileSagaLog) GetActiveSagas() ([]string, error) {
+	files, err := ioutil.ReadDir(log.dirName)
+	if err != nil {
+		return nil, err
+	}
+
+	sagaIds := make([]string, len(files), len(files))
```
dbentley: You can use the two-arg form of make (the last arg is capacity, which is unneeded if you want size == capacity)

saga/sagalogs/file_saga_log_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"fmt"
+	"os"
+	"reflect"
+	"testing"
+
+	"github.com/scootdev/scoot/saga"
+)
+
+func getDirName() string {
+	path, err := os.Getwd()
```
dbentley: github.com/scootdev/scoot/os/temp lets this go in a temp dir, which is nice to avoid spurious writes in our dev tree.

saga/sagalogs/file_saga_log_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	if !bytes.Equal(data, msgs[0].Data) {
+		t.Errorf("Expected Start Saga Message Data to be: %v, Actual: %v", data, msgs[0].Data)
+	}
+
+	if !isSagaInActiveList(sagaId, slog) {
+		t.Errorf("Expected Saga to be in active list")
+	}
+
+	testCleanup(t)
```
dbentley: Put these in a defer at the top to handle panics?

saga/sagalogs/file_saga_log.go
```diff
@@ -0,0 +1,387 @@
...
+	"bufio"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"time"
+
+	"github.com/scootdev/scoot/saga"
+)
+
+// Writes Saga Log to file system.  Not durable beyond machine failure
```
JeanetteBruno: why not durable beyond machine failure?

saga/sagalogs/file_saga_log.go
```diff
@@ -0,0 +1,387 @@
...
+func (log *fileSagaLog) getSagaDirectory(sagaId string) string {
+	return fmt.Sprintf("%v/%v", log.dirName, sagaId)
+}
+
+// Returns the name of the sagalog for the specified file.
+func (log *fileSagaLog) getSagaLogFileName(sagaId string) string {
+	return fmt.Sprintf("%v/%v", log.getSagaDirectory(sagaId), "log")
+}
+
+// Returns the name of the file to store task data in
```
JeanetteBruno: Storing the SagaMessage.Data? (data for saga start, end, abort as well as task messages)  if so then comment should read 'store saga message data' .

saga/sagalogs/file_saga_log.go
```diff
@@ -0,0 +1,387 @@
...
+	return fmt.Sprintf(
+		"%v/StartSagaData_%v",
+		log.getSagaDirectory(sagaId),
+		time.Now().Format(time.StampNano),
+	)
+}
+
+// Log a Start Saga Message message to the log.
+// Returns an error if it fails.
+func (log *fileSagaLog) StartSaga(sagaId string, job []byte) error {
```
JeanetteBruno: can we call this LogSagaStart()?

saga/sagalogs/file_saga_log.go
```diff
@@ -0,0 +1,387 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"time"
+
+	"github.com/scootdev/scoot/saga"
+)
+
+// Writes Saga Log to file system.  Not durable beyond machine failure
+// Sagas are stored in a directory.  Each saga has a corresponding directory
```
JeanetteBruno: would implementation be simpler if we  used json format for the log contents?  - avoid the formatting and parsing in the code below?

saga/sagalog.go
CaitieM20: yup your right probably more idiomatic, fixed

saga/sagalogs/file_saga_log.go
CaitieM20: awesome thanks didn't know about that trick!  yup missed checking the error fixed!

saga/sagalogs/file_saga_log.go
CaitieM20: done!

saga/sagalogs/file_saga_log_test.go
CaitieM20: yup, fixed!

saga/sagalogs/file_saga_log.go
CaitieM20: yup, fixed!

saga/sagalogs/file_saga_log_test.go
CaitieM20: done

saga/sagalogs/file_saga_log.go
CaitieM20: because if the hard drive fails, the file logs aren't replicated to another machine.  So they are lost forever.  

In our final solution using Manhattan, the log is replicated 3 times so it'll survive 2 machine failures

saga/sagalogs/file_saga_log.go
CaitieM20: Sagas are more generic than our application.  Hence why they have our own library.  This in theory could be reused, its a pattern / protocol for Distributed Task Coordination.  

The JobDefinition / JobState is our application specific semantics.  Currently we only use ForwardRecovery Sagas.  Its analogous to a DB Schema and Application specific code.

PR #186, opened 2016-12-28
snapshot: DB interface

dbentley:
DB supersedes Filer. It will allow clients to:
*) ingest data into Scoot
*) manipulate data in Scoot
*) distribute data in Scoot
*) check data out of Scoot (and into the local filesystem)

This is an interface-only change to show the direction.


dbentley:
Hey @jasonbs10 this is the interface I was talking about last week; WDYT?

snapshot/db.go
```diff
@@ -0,0 +1,45 @@
...
+	// IngestGitCommit ingests the commit identified by commitish from ingestRepo
+	// commitish may be any string that identifies a commit
+	// The created value is a Revision.
+	IngestGitCommit(ingestRepo *repo.Repository, commitish string) (SnapID, error)
+
+	// Operations
+
+	// SnapshotForRevision unwraps a Revision and returns a Snapshot.
+	// Errors if id does not identify a Revision.
+	SnapshotForRevision(id ID) (ID, error)
```
JeanetteBruno: what is 'Revision'?  (git revision id? - don't we also need the repo?
what does 'unwraps' mean? - ingest into scoot snapshot db
if Revision is git revision id - how is this different from IngestGitCommit(...) above?

comment should say it returns a Snapshot ID.
changing the two id input parameter to a distinct names would be more readable. eg. SnapshotForRevision(revId ID) (ID, error)

snapshot/db.go
```diff
@@ -0,0 +1,45 @@
+package snapshot
+
+// ID identifies a value in DB
```
JeanetteBruno: if 'value' is a snapshot the comment should use the word 'snapshot' otherwise, we need more info about what a 'value' is.

snapshot/db.go
```diff
@@ -0,0 +1,45 @@
...
+type ID string
+
+// DB is the Scoot Database. It holds Values identified by an ID.
```
JeanetteBruno: see prior comment about 'value'

snapshot/db.go
```diff
@@ -0,0 +1,45 @@
...
+
+	// Operations
+
+	// SnapshotForRevision unwraps a Revision and returns a Snapshot.
+	// Errors if id does not identify a Revision.
+	SnapshotForRevision(id ID) (ID, error)
+
+	// Distribute
+
+	// Upload makes sure the value id is uploaded, returning an ID that can be used
```
JeanetteBruno: I don't understand upload/download:
upload/download from where to where?  (local db to remote db?)

snapshot/db.go
```diff
@@ -0,0 +1,45 @@
...
+	Upload(id ID) (ID, error)
+
+	// Upload makes sure the value id is downloaded, returning an ID that can be used
+	// on this computer or an error
+	Download(id SnapID) (SnapID, error)
+
+	// Export
+
+	// Checkout puts the value identified by id in the local filesystem, returning
+	// the path where it lives or an error.
```
JeanetteBruno: -if 'value' is always snapshot - we should say 'snapshot' instead of 'value'
-will the user want to control where the checkout puts the snapshot?

snapshot/db.go
```diff
@@ -0,0 +1,45 @@
...
+	Checkout(id ID) (path string, err error)
+
+	// ReleaseCheckout releases a path from a previous Checkout. This allows Scoot to reuse
+	// the path. Scoot will not touch path after Checkout until ReleaseCheckout.
+	ReleaseCheckout(path string) error
```
JeanetteBruno: - I need a better understanding of the overall checkout/release/other scoot activities use case/sequence flow.  Can we automate releasing the checkout?
- Does ReleaseCheckout also mean delete path contents?  if not why not?  What is the use case for releasing the path and leaving contents around for later use?

jasonbs10:
I would suggest a few things:

- Rename ID to SnapshotID. This will essentially be an opaque URI with a protocol prefix that can be passed to Checkout(). The protocol types could include snapshot:ID, snapshot-local:ID, etc.

- The idea of values is confusing. Why doesn't the DB just map SnapshotID to an unrealized Checkout (in whatever form that may take going forward).

- SnapshotForRevision is confusing. It should take an arbitrary string and convert it to an ID. We will parse the string and do our best to resolve it to a SnapshotID if it's been uploaded globally (preferred resolution?) or ingested locally. The string can be a git revision or anything else, possibly resulting in a no-op resolution.

- Replace SnapshotForRevision with: ResolveRevision(sid SnapshotID), ResolveSnapshot(id string).

- We should turn Repository into a snapshot.Repo interface imo and remove references to git.

- IngestGitCommit does/says too much. Change this to IngestRepository(snapshot.Repo, rev string). The caller can deal with converting commitish into a revision.

- ReleaseCheckout() and cleanup seems tricky. One approach could be adding a func that lists local SnapshotIDs in case the caller needs to manually release unrecognized/unreleased checkouts from crashes etc.

- Elaborate on Upload and Download. These work with persisted snapshots and the exact details of that persistence are not important here.

snapshot/db.go
dbentley: I added a more holistic description of the structure in doc.go. (Also, renamed Revision to SnapshotWithHistory).

Unwraps means to create a Snapshot without History.

You mean saying it returns a Snapshot ID as opposed to saying "Snapshot"?

There's only one ID input parameter.

snapshot/db.go
dbentley: Value described in doc.go

snapshot/db.go
dbentley: It persists them to a global central store; the details are hidden behind the interface.

I think for V1, we can have two places:
*) local
*) global

This puts a high bar on the service (we need global consistency), but allows us to work on your macbook while offline. WDYT?

snapshot/db.go
dbentley: Value: as above.

I can't think of cases where the user wants to control the checkout. If the user does, they can always copy it, or if that's too expensive, then we can add it then.

snapshot/db.go
dbentley: Release means the user is done with the Checkout.

Releasing it instead of just deleting us allows ScootDB to reuse it. This is useful because doing a git checkout of our work tree takes ~3 minutes, so being able to reuse a checkout once someone is done is a big savings.

If the client doesn't return it to us, it'll be fine, just use extra HD space (unless it's reaped by a tmpreaper)

Added a comment about how we could have more of a way to garbage collect checkouts.

snapshot/db.go
```diff
@@ -0,0 +1,42 @@
+package snapshot
+
```
dbentley: D'oh. Thanks.

snapshot/db.go
```diff
@@ -0,0 +1,45 @@
...
+	// Ingest
+
+	// IngestDir ingests a directory directly.
+	// The created value is a Snapshot.
+	IngestDir(dir string) (ID, error)
+
+	// IngestGitCommit ingests the commit identified by commitish from ingestRepo
+	// commitish may be any string that identifies a commit
+	// The created value is a Revision.
+	IngestGitCommit(ingestRepo *repo.Repository, commitish string) (SnapID, error)
```
dbentley: An error. Meant to be ID.

snapshot/db.go
```diff
@@ -0,0 +1,45 @@
...
+	// Errors if id does not identify a Revision.
+	SnapshotForRevision(id ID) (ID, error)
+
+	// Distribute
+
+	// Upload makes sure the value id is uploaded, returning an ID that can be used
+	// anywhere or an error
+	Upload(id ID) (ID, error)
+
+	// Upload makes sure the value id is downloaded, returning an ID that can be used
```
dbentley: Thx.

dbentley:
@jasonbs10 : think I addressed your comments, as we discussed.

snapshot/db.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// Distribute
+
+	// Upload makes sure the value id is uploaded, returning an ID that can be used
+	// anywhere or an error
+	Upload(id ID) (ID, error)
+
+	// Download makes sure the value id is downloaded, returning an ID that can be used
+	// on this computer or an error
+	Download(id SnapID) (SnapID, error)
```
jasonbs10: SnapID -> ID

snapshot/db.go
```diff
@@ -0,0 +1,50 @@
...
+	// IngestGitCommit ingests the commit identified by commitish from ingestRepo
+	// commitish may be any string that identifies a commit
+	// The created value is a SnapshotWithHistory.
+	IngestGitCommit(ingestRepo *repo.Repository, commitish string) (ID, error)
+
+	// Operations
+
+	// UnwrapSnapshotHistory unwraps a SnapshotWithHistory and returns a Snapshot ID.
+	// Errors if id does not identify a SnapshotWithHistory.
+	SnapshotForRevision(id ID) (ID, error)
```
jasonbs10: Needs to be renamed to match comment.

snapshot/db.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// IngestGitCommit ingests the commit identified by commitish from ingestRepo
+	// commitish may be any string that identifies a commit
+	// The created value is a SnapshotWithHistory.
+	IngestGitCommit(ingestRepo *repo.Repository, commitish string) (ID, error)
+
+	// Operations
+
+	// UnwrapSnapshotHistory unwraps a SnapshotWithHistory and returns a Snapshot ID.
+	// Errors if id does not identify a SnapshotWithHistory.
```
jasonbs10: nit - might want to make this a no-op not an error.

snapshot/db.go
```diff
@@ -0,0 +1,50 @@
+package snapshot
+
+import (
+	"github.com/scootdev/scoot/snapshot/git/repo"
+)
+
+// ID identifies a value in DB. Opaque to the client.
```
jasonbs10: nit - capitalize Value (here and below). And does it make sense to add a breadcrumb to doc.go wrt Value?

snapshot/db.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// Download makes sure the value id is downloaded, returning an ID that can be used
+	// on this computer or an error
+	Download(id SnapID) (SnapID, error)
+
+	// Export
+
+	// Checkout puts the value identified by id in the local filesystem, returning
+	// the path where it lives or an error.
+	Checkout(id ID) (path string, err error)
```
jasonbs10: Will this auto-Download if necessary?

dgassaway:
Can we update scoot/snapshot/README.md in this, as well?

snapshot/db.go
dbentley: Good ideas; all done.

snapshot/db.go
dbentley: I figured it's easier to be strict from the beginning and loosen than vice versa.

snapshot/db.go
dbentley: Done.

snapshot/db.go
dbentley: Ugh; done. Sorry.

dbentley:
Updated README.md, Drew.

snapshot/db.go
jasonbs10: I would consider the API finicky if it didn't...

dbentley: The flip side is that it's nice if the network operations are explicit, so that if something is slow, it's because you chose to Download it.

PR #187, opened 2017-01-03
refactor modules & add FileSagaLogConfig

CaitieM20:
Refactor config/modules.go into files per config section.  This should make understanding config easier, and various versions of things will be next to each other in one place.

Also added FileSagaLog Config Object, so using this version is configurable.


config/scootconfig/scheduler_config_modules.go
```diff
@@ -0,0 +1,33 @@
+package scootconfig
```
ryancouto: for consistency, can this file be named scheduler_modules.go? Looks like the other filenames don't mention config

CaitieM20: so the config section is actually called "SchedulerConfig" that's why its named that

PR #188, opened 2017-01-04
retry smoke test & dial update

CaitieM20:
Currently SmokeTest fails as soon as the scheduler dies.  Any call to ScootAPI that fails results in the test failing.  However, we want to test scheduler restarts, which means some calls may fail.

This update allows failures to the scheduler to be retried.  Currently by reusing the ScootApiClient associated with the Dialer, a broken pipe will be broken forever so Dial() needs to create a new scootClient connection on every call

### DIFF
--- a/scootapi/client/cli.go
+++ b/scootapi/client/cli.go
@@ -49,8 +49,9 @@ func NewSimpleCLIClient(d dialer.Dialer) (CLIClient, error) {
 }
 
 func (c *simpleCLIClient) Dial() error {
-	// Always recreate a connection.  This way if something went wrong with
-	// the previous connection we don't have a clean one.
+	// Always create a new connection, so that we have a clean one
+	// If we reuse connections and something goes wrong (pipe breaks etc...)
+	// there is no way to recreate it.
 	return c.createScootClient()
 }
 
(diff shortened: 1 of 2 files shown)


scootapi/client/smoke_test_cmd.go
```diff
@@ -60,6 +60,11 @@ func (r *smokeTestRunner) run(numJobs int, timeout time.Duration) error {
 		if err != nil {
```
jasonbs10: Don't you need to delete this block?

scootapi/client/cli.go
```diff
@@ -49,28 +49,31 @@ func NewSimpleCLIClient(d dialer.Dialer) (CLIClient, error) {
 }
 
 func (c *simpleCLIClient) Dial() error {
-	_, err := c.dial()
-	return err
+	// Always recreate a connection.  This way if something went wrong with
+	// the previous connection we don't have a clean one.
```
CaitieM20: Yes.  The goal is to always have a clean connection.  Updated the comment to be a little less rambly :)

scootapi/client/smoke_test_cmd.go
CaitieM20: yes! done, sorry missed re-removing it after testing something out

PR #189, opened 2017-01-04
scootapi/client - change CLI to use normal json instead of thrifty-json

dgassaway:
Thrift's JSON files are strange and difficult to read and generate correctly as a person. Since test targets was refactored to use the thrift interface directly, the ingestion of thrift-generated JSON files was no longer used.  This update lets the scootapi CLI ingest jobdefs from JSON files that look like the JobDefinition structure.

This is planned to be used by the scoot/iron compatibility binary.


scootapi/client/run_job_cmd.go
```diff
@@ -22,10 +23,19 @@ func (c *runJobCmd) registerFlags() *cobra.Command {
 		Short: "run a job",
 	}
 	r.Flags().StringVar(&c.snapshotId, "snapshot_id", scoot.TaskDefinition_SnapshotId_DEFAULT, "snapshot ID to run job against")
-	r.Flags().StringVar(&c.jobFilePath, "job_file_path", "", "file to read targets from")
+	r.Flags().StringVar(&c.jobFilePath, "job_file_path", "", "regular (non-thrift) JSON file to read jobs from")
```
dbentley: I might rename to --job_def while we're here.

scootapi/client/run_job_cmd.go
```diff
@@ -22,10 +23,19 @@ func (c *runJobCmd) registerFlags() *cobra.Command {
...
 	return r
 }
 
+// Types to handle JobDefinitions from regular (non-thrift) JSON files
```
dbentley: will "(non-thrift)" make sense in 12 months?

PR #191, opened 2017-01-04
snapshot/git/gitdb: begin implementing snapshot.DB

dbentley:
gitdb implements snapshot.DB using a git repo as its data storage.

For now, only implements IngestDir and Checkout.


snapshot/git/gitdb/db.go
```diff
@@ -0,0 +1,155 @@
...
+	// TODO(dbentley): should we use update-index instead of add? Maybe add looks at repo state
+	// (e.g., HEAD) and we should just use the lower-level plumbing command?
+	cmd := db.dataRepo.Command("add", ".")
+	cmd.Env = append(cmd.Env, extraEnv...)
+	_, err = db.dataRepo.RunCmd(cmd)
+	if err != nil {
+		return "", err
+	}
+
+	cmd = db.dataRepo.Command("write-tree")
```
jasonbs10: Do we want to redefine GIT_WORK_TREE here? I feel like we're overloading the idea of TempDir, but we also need a permanent directory to store our data (I don't think we want to touch 'dir' here), maybe that's another TmpDir member?

snapshot/git/gitdb/db.go
```diff
@@ -0,0 +1,155 @@
...
+	sha, err = repo.ValidateSha(sha)
+	if err != nil {
+		return "", err
+	}
+
+	v := &localValue{sha: sha, kind: kindSnapshot}
+	return v.ID(), nil
+}
+
+// IngestGitCommit ingests the commit identified by commitish from ingestRepo
```
jasonbs10: I haven't been duplicating interface comments in my impls. Do we want to duplicate or breadcrumb the interface comments?

snapshot/git/gitdb/local_data.go
```diff
@@ -0,0 +1,71 @@
...
+		return nil, fmt.Errorf("empty snapshot ID")
+	}
+	if !strings.HasPrefix(string(id), localIDText+"-") {
+		return nil, fmt.Errorf("unrecognized snapshot ID scheme in %s", id)
+	}
+	parts := strings.Split(string(id), "-")
+	if len(parts) != 3 {
+		return nil, fmt.Errorf("did not scan 3 items from %s", id)
+	}
+	kindText, sha := parts[1], parts[2] // skip parts[0] because we know it's local
```
jasonbs10: How do we know it's local? This is in local_data.go but isn't tied to localValue.

snapshot/git/gitdb/db.go
```diff
@@ -0,0 +1,155 @@
...
+	"path/filepath"
+
+	"github.com/scootdev/scoot/os/temp"
+	"github.com/scootdev/scoot/snapshot"
+	"github.com/scootdev/scoot/snapshot/git/repo"
+)
+
+// valueKind describes the kind of a Value: is it a Snapshot or a SnapshotWithHistory?
+// kind instead of type because type is a keyword
+type valueKind int
```
jasonbs10: Can we make this a string and avoid the translation logic in local_data.go?

snapshot/git/gitdb/db.go
```diff
@@ -0,0 +1,155 @@
...
+		return "", err
+	}
+
+	indexFilename := filepath.Join(indexDir.Dir, "index")
+	defer os.RemoveAll(indexDir.Dir)
+
+	extraEnv := []string{"GIT_INDEX_FILE=" + indexFilename, "GIT_WORK_TREE=" + dir}
+
+	// TODO(dbentley): should we use update-index instead of add? Maybe add looks at repo state
+	// (e.g., HEAD) and we should just use the lower-level plumbing command?
```
jasonbs10: If update-index works without any finessing I'd go with that since it's more explicit and it mirrors checkout-index.

snapshot/git/gitdb/db.go
```diff
@@ -0,0 +1,155 @@
...
+
+	indexDir, err := db.tmp.TempDir("git-index")
+	if err != nil {
+		return "", err
+	}
+
+	indexFilename := filepath.Join(indexDir.Dir, "index")
+	defer os.RemoveAll(indexDir.Dir)
+
+	coDir, err := db.tmp.TempDir("checkout")
```
jasonbs10: It's a little weird that we don't store coDir.Dir and rely on ReleaseCheckout but you do have a TODO below. In that vein though, I'd want to tie these checkouts to a particular repo so you can't recycle checkouts from an entirely different repository.

snapshot/git/gitdb/db.go
jasonbs10: Nevermind, dataRepo clearly does this.

snapshot/git/gitdb/db.go
```diff
@@ -0,0 +1,155 @@
...
+	cmd := db.dataRepo.Command("add", ".")
+	cmd.Env = append(cmd.Env, extraEnv...)
+	_, err = db.dataRepo.RunCmd(cmd)
+	if err != nil {
+		return "", err
+	}
+
+	cmd = db.dataRepo.Command("write-tree")
+	cmd.Env = append(cmd.Env, extraEnv...)
+	sha, err := db.dataRepo.RunCmd(cmd)
```
jasonbs10: You could call RunSha() here to avoid some extra lines.

snapshot/git/gitdb/db.go
dbentley: We could, but then everywhere we have to do string comparisons, which are harder. So I'd rather do the translation once (on the way in) then having to check each time.

snapshot/git/gitdb/db.go
dbentley: Doing this will require another step, where we'll have to list every file... Let me think about this a bit.

snapshot/git/gitdb/db.go
dbentley: We definitely want some comments on all Public methods, but yes, they can probably be shorter.

snapshot/git/gitdb/db.go
dbentley: Yup, I think the next step is to store the checkouts. Any kind of reclamation or recycling is TODO.

FWIW, these checkouts, once made, are unconnected to the repo.

snapshot/git/gitdb/db.go
dbentley: We need to extend the environment, and RunSha takes args, not cmd. I could add RunCmdSha, but then we're just leading repo down a combinatorial explosion. WDYT?

snapshot/git/gitdb/local_data.go
jasonbs10: I'd prefer to do that check as part of all the post-split checks but no big deal.

snapshot/git/gitdb/db.go
jasonbs10: It's only one more function right? If we're going to have something like RunSha() at all let's make it easy to use everywhere.

snapshot/git/gitdb/db.go
```diff
@@ -0,0 +1,155 @@
...
+// on this computer or an error
+func (db *DB) Download(id snapshot.ID) (snapshot.ID, error) {
+	return "", fmt.Errorf("not yet implemented")
+}
+
+// Export
+
+// Checkout puts the value identified by id in the local filesystem, returning
+// the path where it lives or an error.
+func (db *DB) Checkout(id snapshot.ID) (path string, err error) {
```
jasonbs10: On second thought this function may not be all that useful wrt to source repo due to size. I guess the idea is that our workers won't call this, they'll instead use it to download bundles to a master repo and then  call Checkouter.Checkout() on a --reference clone?

snapshot/git/gitdb/db_test.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	path, err := fixture.db.Checkout(id)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	actual, err := ioutil.ReadFile(filename)
```
jasonbs10: Is this checking against the right location? I would've expected join(path, foo.txt). Also I take it that it's a TODO to use the fixture's two commits?

snapshot/git/gitdb/db.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	v := &localValue{sha: sha, kind: kindSnapshot}
+	return v.ID(), nil
+}
+
+// IngestGitCommit ingests the commit identified by commitish from ingestRepo
+// commitish may be any string that identifies a commit
+// The created value is a SnapshotWithHistory.
+func (db *DB) IngestGitCommit(ingestRepo *repo.Repository, commitish string) (snapshot.ID, error) {
```
jasonbs10: What would this snapshot contain? Is it a delta that will include a configurable amount of history?

snapshot/git/gitdb/local_data.go
dbentley: Good idea; done.

snapshot/git/gitdb/db.go
dbentley: Done.

snapshot/git/gitdb/db.go
dbentley: There will be no more Checkouter. So this will still be useful, but will need a better impl.

That impl will probably, yes, have to maintain clones/repos in the background.

As is, this will be useful for, e.g., stderr and stdout (where the checkout is small)

snapshot/git/gitdb/db_test.go
dbentley: Embarrassing. Done.

(Yes, I copied it from checkouter_test.go)

snapshot/git/gitdb/db.go
dbentley: For now, all history, because trying to define how much history to bring in is very hard, and I'd suggest we don't block on that rn.

PR #193, opened 2017-01-05
snapshot/db/gitdb: ingest git commit from a local git repo

dbentley:


dbentley:
OK, @jasonbs10 the different strategies between checkoutSnapshot and checkoutSnapshotWithHistory make me think maybe they should be different methods in the API. I think I've come around to your way of thinking.

snapshot/git/gitdb/local_data.go
```diff
@@ -108,13 +109,76 @@ func (db *DB) ingestDir(dir string) (snapshot.ID, error) {
...
+		}
+	}()
+
 	v, err := parseID(id)
 	if err != nil {
 		return "", err
 	}
 
+	// We use different strategies depending on the kind of snapshot.
+	switch v.kind {
```
jasonbs10: I'd unpack v before calling checkout and just pass the sha.

snapshot/git/gitdb/db_test.go
```diff
@@ -54,6 +53,58 @@ func TestIngestDir(t *testing.T) {
 	}
 }
 
+func TestIngestCommit(t *testing.T) {
+	id1, err := fixture.db.IngestGitCommit(fixture.external, fixture.commit1ID)
```
jasonbs10: Doesn't ingesting commit2 take care of this? Depending on what you want to test this detail could be fine.

snapshot/git/gitdb/db_test.go
```diff
@@ -54,6 +53,58 @@ func TestIngestDir(t *testing.T) {
...
+	}
+
+	// test contents
+	data, err := ioutil.ReadFile(filepath.Join(co, "file.txt"))
+	if err != nil || string(data) != "first" {
+		t.Fatalf("error reading file.txt: %q %v (expected \"first\" <nil>)", data, err)
+	}
+
+	// Write temporary data into checkouts to make sure it's cleaned
+	if err = ioutil.WriteFile(filepath.Join(co, "scratch.txt"), []byte("1"), 0777); err != nil {
```
jasonbs10: This makes me wonder if we should have a TODO to relocate cached checkouts when released...

snapshot/git/gitdb/local_data.go
```diff
@@ -108,13 +109,76 @@ func (db *DB) ingestDir(dir string) (snapshot.ID, error) {
...
+
+	// Strategy:
+	// trying to move a commit from ingest to data
+	// delete the ref in the data.
+	// set the ref in the ingest.
+	// push from ingest to data.
+	// delete in both repos.
+	// TODO(dbentley): we could check if sha exists in our repo before ingesting
+
+	if _, err := db.dataRepo.Run("update-ref", "-d", tempRef); err != nil {
```
dbentley: I think if you try to push a ref that also exists in the other place, it could error... But I'm not sure. I guess this was just being careful...

snapshot/git/gitdb/local_data.go
```diff
@@ -108,13 +109,76 @@ func (db *DB) ingestDir(dir string) (snapshot.ID, error) {
...
+	l := &localValue{sha: sha, kind: kindSnapshotWithHistory}
+	return l.ID(), nil
+}
+
+// checkout creates a checkout of id.
 func (db *DB) checkout(id snapshot.ID) (path string, err error) {
-	// Checkout creates a new dir with a new index and checks out exactly that tree.
+
+	defer func() {
+		if path != db.dataRepo.Dir() {
```
dbentley: Commented:

		// If we're our repo dir, we need to keep the work tree locked.
		// Otherwise, we can unlock it.

snapshot/git/gitdb/local_data.go
dbentley: D'oh. I see what you're saying. OK, trying this (though I generally think it's nice to have enums be ints... if we ever see string comparison taking too long, we can revert)

dbentley: Done.

snapshot/git/gitdb/db.go
```diff
@@ -34,10 +35,13 @@ type DB struct {
 	// DB uses a goroutine to serve requests, with requests of type req
 	reqCh chan req
 
+	// must hold workTreeLock before sending a checkoutReq to reqCh
+	workTreeLock sync.Mutex
+
 	// All data below here should be accessed only by the loop() goroutine
 	dataRepo  *repo.Repository
```
dbentley: Done.

snapshot/git/gitdb/db.go
```diff
@@ -108,6 +131,7 @@ type stringAndError struct {
 // Checkout puts the value identified by id in the local filesystem, returning
 // the path where it lives or an error.
 func (db *DB) Checkout(id snapshot.ID) (path string, err error) {
+	db.workTreeLock.Lock()
```
dbentley: Yup, eventually we want to do something smarter. We might also be able to use multiple work-trees if that feature shows up.

snapshot/git/gitdb/db_test.go
dbentley: No, because it doesn't give you the ID for 1.

snapshot/git/gitdb/db_test.go
dbentley: Yeah, there's a lot to improve in how we checkout to balance:
*) correctness
*) cleanliness
*) performance

snapshot/git/gitdb/local_data.go
jasonbs10: I meant that I don't see where path is initialized.

PR #194, opened 2017-01-05
[ScootAPI] Add CloudScootClient Wrapper that Manages Connection & Allow Retries

CaitieM20:
Take2 on making connections retryable.  The [original version](https://github.com/scootdev/scoot/pull/188) was reverted because it didn't close the connections correctly and to do so would have resulted in spreading managing connection logic throughout the code base.  

This version creates a CloudScootClient to manage our connection.  Similar to Finagle Clients but not generic nor fully featured.  This wrapper around the thrift CloudScootClient handles establishing the connection initially and repairing it once it breaks.

This is necessary to allow retries should a connection break.  This is a generally useful library so I added it to scootapi package and modified the CLICLlient in scootapi/client to use this instead of dialing its own.  

Tested by running
go run ./binaries/scootapi/main.go run_smoke_test 100


scootapi/client.go
```diff
@@ -0,0 +1,101 @@
...
+	transport, protocolFactory, err := dialer.Dial(addr)
+	if err != nil {
+		return nil, fmt.Errorf("Error dialing to set up client connection: %v", err)
+	}
+
+	return scoot.NewCloudScootClientFactory(transport, protocolFactory), nil
+}
+
+// helper method to reset the connection
+func (c *CloudScootClient) resetConnection() {
```
JeanetteBruno: change name to closeConnection() ?

scootapi/client/cli.go
```diff
@@ -40,6 +34,22 @@ func NewSimpleCLIClient(d dialer.Dialer) (CLIClient, error) {
...
 
+	// c.addr is populated by flag
+	if c.addr == "" {
+		c.addr = scootapi.GetScootapiAddr()
+		if c.addr == "" {
+			return c, fmt.Errorf("scootapi cli addr unset and no valued in %s", scootapi.GetScootapiAddrPath())
+		}
+	}
+
+	fmt.Printf("Address: %v \n", c.addr)
```
JeanetteBruno: use log

scootapi/client.go
```diff
@@ -0,0 +1,101 @@
+package scootapi
```
JeanetteBruno: move to scootapi/client/?

JeanetteBruno:
I left a couple of comments but they are minor - take them or ignore them.

Two other minor comments .... 

1. stateful_scheduler.go has a startUp() method but it only runs recoverJobs() (line 124).  We can eliminate this method and put the recoverJobs() where startUp() is called (line 109).  

2. Also line 106 has a TODO about allowing scheduler to accept new jobs while recovering old.... should we make this a question? (We could consider recovering jobs as part of spin up - not sure we need to allow accepting jobs while the scheduler is 'spinning up'.

scootapi/client.go
CaitieM20: so i put it in scootapi because our /client package is really a command line tool?  Its more like CLI.  The idea was this is an actual go client for interacting with scootapi.  The thought being if you are looking up the thrift then you'd discover this client and use it hopefully.

scootapi/client.go
CaitieM20: Yes that makes more sense done!

scootapi/client/cli.go
CaitieM20: whoops erroneous debug statement will remove

CaitieM20:
Thanks for the feedback fixed!  I have a PR ready to go once this is merged for the recovery test will fix 1. there.  
> stateful_scheduler.go has a startUp() method but it only runs recoverJobs() (line 124). We can eliminate this method and put the recoverJobs() where startUp() is called (line 109).

> 2. Also line 106 has a TODO about allowing scheduler to accept new jobs while recovering old.... should we make this a question? (We could consider recovering jobs as part of spin up - not sure we need to allow accepting jobs while the scheduler is 'spinning up'.

We want the scheduler to spin up quickly because there is only one of them, so we do want to block on all out recovery before returning the next PR also actually allows this.  The plan is to add back pressure to the scheduler, only allow it to accept so many jobs (there is already a ticket for this) part of that will be making this work with recovery.  My current thoughts are if we have 100 Active Sagas at spin up, and only want to allow 150 jobs at a time, then we should only allow 50 new jobs to be scheduled in the meantime

scootapi/client.go
jasonbs10: I haven't reviewed this but make sure to check out how workerapi/client presents both CLI and actual go client interfaces.

scootapi/client/run_job_cmd.go
```diff
@@ -66,20 +51,7 @@ func (c *runJobCmd) run(cl *simpleCLIClient, cmd *cobra.Command, args []string)
...
-
-		jobDef.Tasks = make(map[string]*scoot.TaskDefinition)
-		for taskName, jsonTask := range jsonJob.Tasks {
-			taskDef := scoot.NewTaskDefinition()
-			taskDef.Command = scoot.NewCommand()
-			taskDef.Command.Argv = jsonTask.Args
-			taskDef.SnapshotId = &jsonTask.SnapshotID
-			jobDef.Tasks[taskName] = taskDef
-		}
+		thrifthelpers.JsonDeserialize(jobDef, asBytes)
```
dbentley: Looks like this is reverting Drew's recent change.

Drew changed it to parse a reasonable JSON, not the thrift JSON (which uses numeric field names instead of, y'know, names)

 Is that intentional?

dgassaway: Was going to mention this, it looks like straight revert of run_job_cmd.go...

CaitieM20: sorry must have been a bad merge, not intentional fixed!

scootapi/client.go
CaitieM20: its similar but doesn't expose Dial() just the CloudScoot Interface, but I suspect after taking a quick look workerapi/client will have the same problem that this client had before.  If the connection dies due to broken pipe or an unrecoverable error, the connection needs to be closed and re-established.  Otherwise every subsequent call will fail.

PR #195, opened 2017-01-06
[Scheduler] Saga Recovery Test

CaitieM20:
Add a Saga Recovery Test to CI, that verifies that the scheduler can be killed and restarted and all running jobs run to completion.

* Modified Scheduler Config for local.local instance to use the File SagaLog, so we have a SagaLog that is durable past scheduler process failure.  Right now creates a directory in the repo which is added to .gitignore.  Open to suggestions on a better place to put this by default.  
* Added a binaries/recoverytest/main.go which will run the Recovery Test.  Spins up a local cluster, starts a bunch of jobs, kills the cluster, recovers the cluster and then ensures all scheduled jobs run to completion.  
* RecoveryTest is ran as part of CI instead of swarmtest, since it does the same thing as swarmtest plus recovery, we don't need to run both.  
* Refactored useful functions from smoke_test into testhelpers package so they can be reused by recovery test.  
* Fixed a bug in FileSagaLog where GetMessages() for a non existent saga will return an error instead of nil. 
* Updated Stateful Scheduler to be created and returned before all jobs are recovered. We want to do this for latency reasons.  Since there is only one stateful scheduler, we need to minimize startup time, to minimize downtime.  Also the way it was written before caused problems when too many jobs were recovered, since we hadn't started the inner scheduler loop before rescheduling jobs.  

tested by running 
* make recoverytest or
* go run ./binaries/recoverytest/main.go


binaries/recoverytest/main.go
```diff
@@ -0,0 +1,92 @@
...
+	go func() {
+		defer wg.Done()
+		err := testhelpers.WaitForJobsToCompleteAndLogStatus(jobIds, scootClient, timeout)
+		if err != nil {
+			log.Fatalf("Error Occurred Waiting For Jobs to Complete.  %v", err)
+		}
+	}()
+
+	// Wait for Jobs to Start Running and make some progress
+	time.Sleep(1 * time.Second)
```
dgassaway: Just curious is there some chance these could complete normally before the first cluster is killed? Or for some reason haven't started after 1 sec?

.gitignore
```diff
@@ -33,4 +33,5 @@ coverage.txt
 # Emacs save files have literal pound signs/octothorpes/hashtags in them
 # So the \# escapes # as comment syntax
 \#*#
-.\#*
\ No newline at end of file
+.\#*
+sagalog/*
```
dgassaway: On file locations, we are starting to generate some litter in various places. Not sure if there is an idiomatic way of doing this but I would prefer we either pick a dir path to place everything like ~/[scootdata], or pick a convention to have all dot files in ~/, like cloudscootaddr

sched/scheduler/stateful_scheduler.go
```diff
@@ -116,15 +113,16 @@ func NewStatefulScheduler(
 		}()
 	}
 
+	// Recover Jobs in a separate go routine to allow the scheduler
```
dgassaway: I wonder if we will want some way to signal that recovery is in progress - if users submit a job and after a restart the recovery takes a long time or freezes or fails, they might think their work disappeared although the scheduler seems healthy.  (this is a long reach)

.gitignore
JeanetteBruno: recommend update the test to remove the directory (on success)... eliminating this ignore.

binaries/recoverytest/main.go
```diff
@@ -0,0 +1,92 @@
...
+                     ((__)        \\||lll|l||///          \_))
+                              (   /(/ (  )  ) )\   )
+                            (    ( ( ( | | ) ) )\   )
+                             (   /(| / ( )) ) ) )) )
+                           (     ( ((((_(|)_)))))     )
+                            (      ||\(|(|)|/||     )
+                          (        |(||(||)||||        )
+                            (     //|/l|||)|\\ \     )
+                          (/ / //  /|//||||\\  \ \  \ _)
+  -------------------------------------------------------------------------------`)
```
JeanetteBruno: awesome - do we have to worry about side effects? (just kidding!)

sched/scheduler/stateful_scheduler.go
CaitieM20: so right now we will accept and run new jobs as we recover.  There is a ticket to add back pressure ( not accept too many jobs ) so users would get a Scoot Too Busy error instead of waiting a long time. 

I agree it would be nice in the future to have some kind of estimated wait time, but that's a bit far off.

dgassaway: SGTM

.gitignore
CaitieM20: Great Suggestion @dgassaway I created a .scootdata folder which can contain misc garbage data files that should not be checked in.  I updated the config to put fileSagaLog files here by default.  I also added deleting the files in this directory to make clean.  This directory is ignored.  

@JeanetteBruno at the end of make travis I invoke clean-data, this way we don't pollute those machines.  I'm hesitant to just have recoverytest blow it away incase people were relying on data that is in there for a previous run.

binaries/recoverytest/main.go
CaitieM20: So it could in theory complete before killing the cluster, but in practice won't with 20 tests that just sleep for 500ms.  

We are guaranteed because I wait till the jobs are started, that they are all in the scheduler queue, if there was some really bad lag they might not start but that would be really weird on a brand new clean cluster.  We also have a unit test already that verifies that the 500ms jobs run in a reasonable amount of time, ensures we are not leaving idle workers with scheduling etc...

We could improve this in the future by doing a check that a percentage of tasks are in progress or completed before killing the cluster in the future as we build out the test harness.

PR #196, opened 2017-01-06
snapshot/db/gitdb: implement streams

dbentley:
A Scoot Stream follows a Git ref that can be fetched from a Git remote.

Create the notion of "Schemes", which abstract how to identify, upload and download Values. There are now two schemes:
*) Local
*) Stream

Refactor out common functions in tests.

Still unimplemented: uploading


snapshot/git/gitdb/schemes.go
```diff
@@ -0,0 +1,65 @@
+package gitdb
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/scootdev/scoot/snapshot"
+)
+
+// GitDB uses several different Schemes to identify, upload and download Values.
```
jasonbs10: I'd get rid of the concept of Value if possible, it adds one too many types. Values are equivalent to different types of snapshots right?

snapshot/git/gitdb/checkout.go
```diff
@@ -0,0 +1,111 @@
...
+func (db *DB) releaseCheckout(path string) error {
+	if path == db.dataRepo.Dir() {
+		db.workTreeLock.Unlock()
+		return nil
+	}
+
+	if exists := db.checkouts[path]; !exists {
+		return nil
+	}
+	err := os.RemoveAll(path)
```
dbentley: I believe not, because of the check above if path == db.dataRepo.Dir() then we return.

snapshot/git/gitdb/local_data.go
```diff
@@ -19,42 +18,31 @@ type localValue struct {
...
-func parseID(id snapshot.ID) (*localValue, error) {
-	if id == "" {
-		return nil, fmt.Errorf("empty snapshot ID")
-	}
-	parts := strings.Split(string(id), "-")
-	if len(parts) != 3 {
-		return nil, fmt.Errorf("did not scan 3 items from %s", id)
+func parseLocalID(id snapshot.ID, kind valueKind, parts []string) (*localValue, error) {
+	if len(parts) != 1 {
+		return nil, fmt.Errorf("cannot parse snapshot ID: expected 3 parts in local id %s", id)
```
dbentley: We expect three parts; really the parts handed in to parseLocalID are the additional parts (because parseID takes care of scheme and kind before calling parseLocalID).

Added a comment.

snapshot/git/gitdb/schemes.go
dbentley: (See also the design doc; should be in your email)

It's annoying in Unix filesystems that a file can be either a directory or... a file? Wait? What?

I'm proposing:
ScootDB stores Values.
A Value can be either a Snapshot or SnapshotWithHistory (for now; eventually we plan to store files, and who knows, maybe we just store a raw AST at some point).

So I think we need k + 1 terms: 1 for each of k types, plus for the baseclass/superclass.

WDYT?

snapshot/git/gitdb/schemes.go
```diff
@@ -0,0 +1,65 @@
...
+// schemeValue abstracts these concerns.
+type schemeValue interface {
```
dbentley: I was thinking this refers to one value in a scheme. So the localValue object refers to one value specifically. I'd expect localScheme to be for the whole scheme.

snapshot/git/gitdb/schemes.go
```diff
@@ -0,0 +1,65 @@
...
+	id() snapshot.ID
+
+	// F suffix because name conflict between Function and member.
```
dbentley: Capitalized.

snapshot/git/gitdb/stream.go
```diff
@@ -0,0 +1,66 @@
+package gitdb
+
+import (
+	"fmt"
+
+	"github.com/scootdev/scoot/snapshot"
+)
+
+// A Stream is a sequence of SnapshotWithHistory's that updates via a
```
dbentley: Let's discuss the naming in the design doc (and punt on it here for now; it's easy to rename).

It's a sequence because it advances outside of ScootDB. If a user passes an ID in the Local Scheme and it's not present in the data repo, ScootDB can't get it. With the scheme stream, it knows to fetch.

snapshot/git/gitdb/stream.go
```diff
@@ -0,0 +1,66 @@
...
+// Git ref that can be fetched from a Git remote
+type StreamConfig struct {
+	// Name (used in IDs, e.g. sm)
```
dbentley: Expanded comment.

snapshot/git/gitdb/stream.go
```diff
@@ -0,0 +1,66 @@
...
+	Name string
+
+	// Remote to fetch from (e.g. https://github.com/scootdev/scoot)
+	Remote string
+
+	// Name of ref to follow in data repo (e.g. refs/remotes/upstream/master)
+	RefSpec string
```
dbentley: Don't need it for this iteration; happy to remove.

We'll use it in uploading to find merge-base for bundle basis selection.

snapshot/git/gitdb/stream.go
```diff
@@ -0,0 +1,66 @@
...
+	return v.id(), nil
+}
+
+func (v *streamValue) download(db *DB) (snapshot.ID, error) {
+	// TODO(dbentley): check if present before downloading
+
+	if db.stream == nil {
+		return "", fmt.Errorf("cannot download snapshot %s: no streams configured", v.id())
+	}
+	if v.streamName != db.stream.Name {
```
dbentley: Because you might want to follow two repo's master branches.

snapshot/git/gitdb/checkout.go
jasonbs10: My mistake, just my immediate reaction when I see a path arg instead of some kind of id.

snapshot/git/gitdb/schemes.go
jasonbs10: Value is just too generic, same as saying it stores Foo imo. If there's some concept that's more specific, we should find a word for that.

snapshot/git/gitdb/stream.go
jasonbs10: We can punt but I've been saying all along that Stream doesn't seem like the most natural fit.

snapshot/git/gitdb/schemes.go
dbentley: OK; let's take this to the design doc (to avoid similar conversations happening in too many places)?

snapshot/git/gitdb/checkout.go
dbentley: Yeah, I think Checkout and Release are overly complicated rn because it's implicit (based on the kind) whether we're going to have a bare checkout (if it's a Snapshot) or it will be our git work-tree (if it's a SnapshotWithHistory).

Also would be nice if we could git's work-tree feature to avoid having to do all this by hand.

snapshot/git/gitdb/stream.go
```diff
@@ -0,0 +1,72 @@
...
+
+	if db.stream == nil {
+		return "", fmt.Errorf("cannot download snapshot %s: no streams configured", v.ID())
+	}
+	if v.streamName != db.stream.Name {
+		return "", fmt.Errorf("cannot download snapshot %s: does not match stream %s", v.ID(), db.stream.Name)
+	}
+
+	// TODO(dbentley): keep stats about fetching (when we do it, last time we did it, etc.)
+	if _, err := db.dataRepo.Run("fetch", db.stream.Remote); err != nil {
```
jasonbs10: TODO: verify that the desired sha is present?

snapshot/git/gitdb/stream.go
```diff
@@ -0,0 +1,72 @@
...
+	Remote string
+
+	// Name of ref to follow in data repo (e.g. refs/remotes/upstream/master)
+	RefSpec string
+}
+
+const streamIDText = "stream"
+const streamIDFmt = "%s-%s-%s-%s"
+
+// parse id as a sream ID, with kind and remaining parts (after scheme and kind were parsed)
```
jasonbs10: sream -> stream

snapshot/git/gitdb/schemes.go
```diff
@@ -0,0 +1,64 @@
...
+	Upload(db *DB) (snapshot.ID, error)
+	Download(db *DB) (snapshot.ID, error)
+}
+
+// parseID parses ID into a schemeValue
+func parseID(id snapshot.ID) (schemeValue, error) {
+	if id == "" {
+		return nil, fmt.Errorf("empty snapshot ID")
+	}
+	parts := strings.Split(string(id), "-")
```
dgassaway: Not clear from the code what a valid snapshot.ID is - I see it's in the doc, I would suggest at some point it's added to a README.md

snapshot/git/gitdb/checkout.go
```diff
@@ -0,0 +1,111 @@
...
+		return "", err
+	}
+
+	db.checkouts[coDir.Dir] = true
+
+	return coDir.Dir, nil
+}
+
+// checkoutSnapshotWithHistory checks out a commit into our work tree.
+// We could use multiple work trees, except our git doens't yet have work-tree support.
```
dgassaway: doens't -> doesn't

snapshot/git/gitdb/stream.go
dbentley: Good call. Done.

snapshot/git/gitdb/stream.go
dbentley: Done.

snapshot/git/gitdb/schemes.go
dbentley: Great point. I added a README.md. (That will be very much in flux).

I also added a description of snapshot/git/gitdb to snapshot/README.md

PR #197, opened 2017-01-09
Added dbentley's Bundlestore (used as a simple Data Access Layer).

jasonbs10:
This is an http server that delegates to a Store type for reading and writing named data.
Currently there is a file store type and an http store type (eg: delegate to san storage).
This does not cache data yet but it will need to.


snapshot/bundlestore/server_test.go
```diff
@@ -0,0 +1,128 @@
...
+	//
+	// Repeat with HttpStore as the client.
+	//
+
+	// Try to write data.
+	httpStore := MakeHTTPStore(rootUri)
+	if err := httpStore.Write("bar", bytes.NewBuffer([]byte("bar_data"))); err != nil {
+		t.Fatalf(err.Error())
+	}
+	time.Sleep(50 * time.Millisecond)
```
dbentley: Why the sleep?

snapshot/bundlestore/store.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"io"
+	"log"
+	"os"
+	"path/filepath"
+
+	"github.com/scootdev/scoot/os/temp"
+)
+
+type StoreRead interface {
```
dbentley: Comment exported Members? (Sorry! I know I'm the slacker who pooped out this uncommented code)

snapshot/bundlestore/server_test.go
```diff
@@ -0,0 +1,128 @@
...
+	if resp, err := client.Get(rootUri + "foo"); err != nil {
+		t.Fatalf(err.Error())
+	} else {
+		resp.Body.Close()
+		if resp.StatusCode != http.StatusNotFound {
+			t.Fatalf("Expected NotFound when getting non-existent data.")
+		}
+	}
+
+	//
```
dgassaway: Would it make sense to break this out as a separate test?

jasonbs10:
The latest push adds the ability to extract bundle resources via cachingBrowseStore.

More documentation to come.

snapshot/bundlestore/browse_store.go
```diff
@@ -0,0 +1,76 @@
...
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/scootdev/scoot/os/temp"
+)
+
+// Wraps another store and, when reading and when bundle name contains a slash, returns interpreted data.
```
dbentley: What does "interpreted" mean?

Why require the slash?

snapshot/bundlestore/browse_store.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (s *cachingBrowseStore) Write(name string, data io.Reader) error {
+	if strings.Contains(name, "/") {
+		return errors.New("'/' not allowed in name unless reading bundle contents.")
+	}
+	return s.underlying.Write(name, data)
+}
+
+func (s *cachingBrowseStore) extract(bundlePath, contentPath string) (io.ReadCloser, error) {
```
dbentley: Oh, I see. Can I suggest waiting on this part and just using a GitDB in the future? I think that will result in fewer requirements and code paths.

snapshot/bundlestore/server_test.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func TestServer(t *testing.T) {
+	// Construct server with a fake store and random port address.
+	store := &FakeStore{files: map[string][]byte{}}
+	listener, _ := net.Listen("tcp", "localhost:0")
+	listener.Close()
+	addr := listener.Addr().String()
+	go MakeServer(store, Addr(addr)).Serve()
+	time.Sleep(50 * time.Millisecond) // needed for travis-ci?
```
dbentley: What is the error if you don't have this on Travis CI?

snapshot/bundlestore/server_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	rootUri := "http://" + addr + "/bundle/"
+	client := &http.Client{Timeout: 1 * time.Second}
+
+	// Try to write data.
+	if resp, err := client.Post(rootUri+"baz-baz-baz", "text/plain", bytes.NewBuffer([]byte("baz_data"))); err != nil {
```
dbentley: I'd put a TODO to test with real Bundle files (but we can wait on that for now)

snapshot/bundlestore/store.go
dbentley: Ping.

snapshot/bundlestore/store.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (c *cache) Write(name string, data []byte) {}
+
+// Wraps another store and does simplistic caching of reads and exists calls.
+func MakeCachingStore(s Store) Store {
+	return &cachingStore{underlying: s}
+}
+
+type cachingStore struct {
```
dbentley: Either expand this or delete.

snapshot/bundlestore/browse_store.go
jasonbs10: I think this is addressed in the next line. The slash is because we can treat that as a reserved character wrt to bundle names and they're also required to represent a path. So it's bundleName+contentPath, which make sense to me. I'll try to clarify a bit.

dbentley: OHHHH, I think I see. I was confusing what this is.

Yeah, I think this will be easier in the future with SnapshotDB, but this is fine for now...

snapshot/bundlestore/browse_store.go
jasonbs10: That's what I was planning to do later while we're adding functionality around bundles like verifying and regenerating w/ new basis (i.e. pulling in a repo + gitdb and modifying those as needed). This is a more immediate step that's fairly small in scope.

snapshot/bundlestore/server_test.go
jasonbs10: Added TODO.

snapshot/bundlestore/store.go
jasonbs10: Yes, required before merging.

PR #198, opened 2017-01-12
scootapi: allow configuring number of tasks per job; output job progress

dbentley:


scootapi/client/smoke_test_cmd.go
```diff
@@ -10,57 +10,62 @@ import (
...
-func (r *smokeTestRunner) run(numJobs int, timeout time.Duration) error {
+func (r *smokeTestRunner) run(numJobs int, numTasks int, timeout time.Duration) error {
 	jobs := make([]string, 0, numJobs)
 
 	for i := 0; i < numJobs; i++ {
-		id, err := testhelpers.GenerateAndStartJob(r.cl.scootClient)
-		// retry starting job until it succeeds
-
-		for err != nil {
+		for {
```
jasonbs10: nit - unnecessary deletion of comment, it would be better if expanded instead.

scootapi/client/watch_job_cmd.go
```diff
@@ -76,8 +76,12 @@ func PrintJobStatus(jobStatus *scoot.JobStatus) {
 		fmt.Printf("\tTask %s {\n", taskId)
 		fmt.Printf("\t\tStatus: %s\n", taskStatus.String())
 		runStatus := jobStatus.TaskData[taskId]
-		fmt.Printf("\t\tStdout: %v\n", *runStatus.OutUri)
-		fmt.Printf("\t\tStderr: %v\n", *runStatus.ErrUri)
+		if runStatus.OutUri != nil {
```
jasonbs10: Is this addressing a symptom or root cause? My expectation, unless I'm forgetting what's been merged, is that these should always be set.

tests/testhelpers/generators.go
```diff
@@ -17,13 +17,15 @@ func NewRand() *rand.Rand {
 // Test Helpers that are useful for Generating random Scoot Api Structs
 // To help with testing the Scoot API and Scheduler
 
-// Generates a scoot.JobDefinition
-func GenJobDefinition(rng *rand.Rand) *scoot.JobDefinition {
-
+// Generates a scoot.JobDefinition with numTasks tasks (or random if numTasks == -1)
```
jasonbs10: The semantics of numTasks should be made visible to the user via help string, and it needs to be bounded.

tests/testhelpers/jobHelpers.go
```diff
@@ -67,7 +66,9 @@ func WaitForJobsToCompleteAndLogStatus(
 		for jobId, oldStatus := range jobs {
 
 			if !IsJobCompleted(oldStatus) {
+				log.Println("getting status")
```
jasonbs10: What were these prints intended to address? It seems like it might unnecessarily spam the output.

scootapi/client/smoke_test_cmd.go
```diff
@@ -10,57 +10,62 @@ import (
...
+	r.Flags().DurationVar(&c.timeout, "timeout", 180*time.Second, "how long to wait for the smoke test")
 	return r
 }
 
 func (c *smokeTestCmd) run(cl *simpleCLIClient, cmd *cobra.Command, args []string) error {
 	fmt.Println("Starting Smoke Test")
 
-	numJobs := 100
-
+	// TODO(dbentley): migrate away from positional args to flags
```
jasonbs10: What needs to be migrated? There can't be that much stuff that relies on smoke_test invocation...

dbentley: I don't know what scripts people have. I've updated the Makefile. Want me to be bold and just delete them now?

scootapi/client/smoke_test_cmd.go
dbentley: Hrm; I don't know why we do this...

Added the comment back in; I'll see if I can track down history.

scootapi/client/watch_job_cmd.go
dbentley: If a task isn't (or even started), it doesn't seem crucial to set it.

tests/testhelpers/generators.go
dbentley: Added to help string.

What bound should there be? Why bound on the client instead of on the server?

I was playing last night with 20k, which seems totally reasonable.

tests/testhelpers/jobHelpers.go
dbentley: Yup; removed.

scootapi/client/smoke_test_cmd.go
jasonbs10: I would remove it since it only concerns dx-team at this time.

scootapi/client/watch_job_cmd.go
jasonbs10: That's not exactly what I'm saying. I'm asserting that my intention in an earlier PR was to guarantee that these were set. If they're not, it probably contradicts some of my code comments.

tests/testhelpers/generators.go
jasonbs10: Less than 1B? Actually it would be good to know how much is too much on a macbook.

dbentley: I don't want to have a whole bunch of tests all throughout the code base that we have to try to keep synced. It seems like the server should implement that (because it's where we want backoff to be done)

scootapi/client/watch_job_cmd.go
dbentley: Yeah, I think that got affected by my runner refactor and we haven't rolled it forward.

PR #199, opened 2017-01-12
saga: copy on get and logMessage, not update. 

dbentley:
Getting the status for a Job should be quick (even a second is longer than ideal). It currently takes several minutes for a job with 8000 tasks. Eep!

Getting the status requires reading the Saga State. Constructing the Saga State requires Recovery.

Saga recovery has to replay each message (which is O(number of messages)). Currently, updateSagaState copies the Saga State (which is O(size of state)).

This means that the time to recover a saga is superlinear (always copying a growing object).

This isn't needed on recovery, because we trust the messages are valid. Move the copy out of recovery and do it only on get (when the state is handed to a client) and in logMessage (where the write to durable storage may fail, and then we'll want to revert the state). This makes updateSagaState mutate, and reduces the cost of saga recovery from minutes (and growing) to seconds (even up to 15,000 tasks).


saga/saga.go
```diff
@@ -182,17 +182,9 @@ func (s *Saga) updateSagaStateLoop() {
...
-
-	if err == nil {
-		s.state = newState
-
-		update.resultCh <- nil
-
-	} else {
-		update.resultCh <- err
-	}
+	var err error
```
JeanetteBruno: while we are changing this area of the code, can we change the name to something like 'lockedUpdate'

dbentley: How's that?

JeanetteBruno: much better -thx

JeanetteBruno:
lgtm

jasonbs10:
I'm not connecting the dots - what is this trying to accomplish and how does this accomplish that? I'd like the comments to be beefed up here.

saga/saga.go
```diff
@@ -201,23 +194,27 @@ type sagaUpdate struct {
 }
 
 //
-// logs the specified message durably to the SagaLog & updates internal state if its a valid state transition
+// checks the message is a valid transition and logs the specified message durably to the SagaLog
+// if msg is an invalid transition, it will neither log nor update internal state
```
jasonbs10: I'd add a comment saying that it returns the original state when there's an err since that's not typical behavior in our codebase.

saga/saga_state.go
```diff
@@ -246,24 +246,20 @@ func (state *SagaState) addTaskData(taskId string, msgType SagaMessageType, data
 }
 
 /*
- * Applies the supplied message to the supplied sagaState.  Does not mutate supplied Saga State
- * Instead returns a new SagaState which has the update applied to it
+ * Applies the supplied message to the supplied sagaState.
+ * Mutates state only if msg is a valid transition.
```
CaitieM20: Is this true?  Couldn't it mutate and return an error?  That was the reason for the copy in the first place

saga/saga_state_test.go
```diff
@@ -55,26 +55,3 @@ func TestSagaState_Copy(t *testing.T) {
 		t.Error(fmt.Sprintf("Copy Should Preserve SagaId"))
 	}
 }
-
-func TestSagaState_SagaStateNotMutatedDuringUpdate(t *testing.T) {
```
CaitieM20: this should be replaced with a test that ensures that an update resulting in an error does not result in a mutation

saga/saga.go
CaitieM20: +1

dbentley: Done.

saga/saga_state.go
dbentley: I believe the code currently mutates only after checking, but that's a hard thing to ensure we maintain, so I've removed the comment.

saga/saga_state_test.go
dbentley: The tests in saga_test.go already check for this (and were very useful in diagnosing what was going on).

CaitieM20: awesome past caitie did us both a solid, lol

PR #200, opened 2017-01-14
sched: include unstarted tasks in status

dbentley:
Improve logging for task completion or error.

Improve cmds.go for setup-cloud-scoot: logging and dying after an earlier cmd died.


scootapi/setup/cmds.go
```diff
@@ -131,12 +132,18 @@ func (c *Cmds) Command(path string, arg ...string) *exec.Cmd {
...
+		return fmt.Errorf("killed; cannot start new cmds")
+	}
 	log.Println("Starting", cmd.Args)
 	err := cmd.Start()
 	if err == nil {
 		go func() {
+			pid := cmd.Process.Pid
 			cmd.Wait()
-			log.Printf("Cmd %v finished", cmd.Path)
+			log.Printf("Cmd %v (%v) finished", pid, cmd.Path)
```
jasonbs10: Does it make sense to print pids for each started process if we care about pid here?

sched/scheduler/stateful_scheduler.go
```diff
@@ -286,11 +286,13 @@ func (s *statefulScheduler) scheduleTasks() {
...
 			},
 			func(err error) {
-				log.Println("Ending task", taskId, " command:", strings.Join(taskDef.Argv, " "))
 				// update the jobState
 				if err == nil {
+					log.Println("Ending task", taskId, " command:", strings.Join(taskDef.Argv, " "))
+
 					jobState.taskCompleted(taskId)
 				} else {
+					log.Println("Error running task", taskId, " command:", strings.Join(taskDef.Argv, " "))
```
dbentley: Done.

scootapi/setup/cmds.go
dbentley: Done.

common/thrifthelpers/thrift_serializer.go
```diff
@@ -39,6 +39,8 @@ func BinaryDeserialize(targetStruct thrift.TStruct, sourceBytes []byte) (err err
 		return nil
 	}
 	d := thrift.NewTDeserializer()
+	// NB(dbentley): this seems to have pathological behavior on some strings. E.g.,
+	// a random 34 bytes took 45 seconds to decode.
```
JeanetteBruno: is the delay consistent or intermittent? (is it gc?) -it would be good to include this info in the comment

saga/saga_generators.go
```diff
@@ -146,10 +149,10 @@ func GenId() gopter.Gen {
 	}
 }
 
-// Generator for a Valid Saga State
-func GenSagaState() gopter.Gen {
+// Generator for a Valid Saga State, with a job def if includeJob
+func GenSagaState(includeJob bool) gopter.Gen {
```
CaitieM20: Is there a reason we wouldn't want the job?  Why not just always have it add it, then the test is more realistic

PR #201, opened 2017-01-16
Added memory constraint option to runner.Command. DX1703.

jasonbs10:
This does best effort monitoring of a runner command and kills it if memory usage exceeds a given limit. It measures all rss memory used by the command and all descendent processes (ballpark accuracy hopefully).

As far as I can tell, rlimit will not work for us since the limit essentially resets for every every child process. And the proper way to do this, cgroups etc, is not something we have control over.

go/j/DX1703

### DIFF
--- a/binaries/scheduler/config/config.go
+++ b/binaries/scheduler/config/config.go
@@ -161,7 +161,7 @@ func AssetNames() []string {
 
 // _bindata is a table, holding each asset generator, mapped to its name.
 var _bindata = map[string]func() (*asset, error){
-	"config/local.local": configLocalLocal,
+	"config/local.local":  configLocalLocal,
 	"config/local.memory": configLocalMemory,
 }
 
@@ -204,9 +204,10 @@ type bintree struct {
 	Func     func() (*asset, error)
 	Children map[string]*bintree
 }
+
 var _bintree = &bintree{nil, map[string]*bintree{
 	"config": &bintree{nil, map[string]*bintree{
-		"local.local": &bintree{configLocalLocal, map[string]*bintree{}},
+		"local.local":  &bintree{configLocalLocal, map[string]*bintree{}},
 		"local.memory": &bintree{configLocalMemory, map[string]*bintree{}},
 	}},
 }}
@@ -257,4 +258,3 @@ func _filePath(dir, name string) string {
 	cannonicalName := strings.Replace(name, "\\", "/", -1)
 	return filepath.Join(append([]string{dir}, strings.Split(cannonicalName, "/")...)...)
 }
-
--- a/binaries/workerserver/main.go
+++ b/binaries/workerserver/main.go
@@ -13,7 +13,7 @@ import (
 	"github.com/scootdev/scoot/common/stats"
 	"github.com/scootdev/scoot/config/jsonconfig"
 	"github.com/scootdev/scoot/os/temp"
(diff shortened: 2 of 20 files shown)


dbentley:
What would you think of moving this into execer? It seems like something that belongs most in os_execer, e.g.?

Why do this actively? Would rlimit solve this without requiring active monitoring?

jasonbs10:
My understanding is that rlimit resets the counter for allowed memory for each new child. This was the simplest way I could think of to aggregate the memory consumption for the process group. Is that incorrect?
I'll think about moving it all to execer. Is that a blocker for you though?


		_____________________________

dbentley:
It would be weird if you could get around an rlimit by just starting a whole bunch of processes...

Moving it into execer isn't blocking... Is there a hurry to this? Have we been bit by this recently? Is it in our current scrum?

jasonbs10:
I can't find anything online to indicate rlimit would work for us, mostly the opposite. See for example:http://coldattic.info/shvedsky/pro/blogs/a-foo-walks-into-a-bar/posts/40

I did this because I saw a bunch of tickets about this being a biggish problem on iron, so this should help us debug our way to 100% test coverage. Also it was a self contained task that didn't require undue effort. 



On Sun, Jan 15, 2017 at 6:58 PM -0800, "Dan Bentley" <notifications@github.com> wrote:












It would be weird if you could get around an rlimit by just starting a whole bunch of processes...


Moving it into execer isn't blocking... Is there a hurry to this? Have we been bit by this recently? Is it in our current scrum?

JeanetteBruno:
(sorry to take so long to review)
Good Feature
  
comment 1:
wrt: should it be in runner vs. execer: I think execer:  
my reasoning: I envision also adding cpu cap -> I think we will need pid to do this -> only os_execer has pid -> both mem cap and cpu cap should be in the same location => keep it in execer.

comment 2:
Will we want to extend it with mem cap/command?  If so we should create a (low priority) jira ticket for a mem cap/command feature. 

comment 3:
How is iron solving this issue?  is it killing processes that take up too much memory (do we know what this cap is?) Will this change make scoot behave the same as iron, or will it give us a list of tests that we shouldn't expect to align with iron?  We'll need to update Drew's change to supply this option - create a jira ticket to make sure we don't forget to do this?

dgassaway:
I agree with the execer vs runner comments.  This looks like a good baseline that we could use to collect other data about commands in the future, so that is good.

I was looking for an alternative or workaround to the rlimit "vulnerability" but it seems like the multiprocess thing is, well, a thing.  It would be nice if there were user cgroups on mesos tasks we could subgroup for commands but I don't think that is going to happen.

Re: @JeanetteBruno comment 3: Iron doesn't solve this right now.  The node just gets killed by mesos, has no chance to do anything gracefully, and IronClient just loses contact with it and has to retry or reschedule or whatever it does when a node goes down.  These are annoying to track down, because you have to look through the logs via aurora to determine what happened, and this would already be way ahead of that by 1) making it more obvious and 2) not losing the worker.

LGTM, only not clicking approve because I'm not too fluent on runner/execer code lately.

jasonbs10:
I think this is unlikely to require more functionality in the near-term. We can address needed changes when/if we decide to add support for concurrent runs to workerserver.

Are there any action items I'm missing?

PR #203, opened 2017-01-18
TwitterServer displays stats from sidecar on metrics endpoint

CaitieM20:
We want to also collect stats from our sidecar services.  Since there is only one Absorber (daemon that collects stats by querying stats endpoint and then pushes to twitter obs) per aurora deploy we want to also have the stats endpoint of the main service, query stats of its sidecar services and return those as well.  This way all stats will be collected by the single Absorber and pushed to Twitter Obs. 

I modified TwitterServer to take in a list of sidecar ports.  If specified their stats endpoints are also queried and returned when querying the stats endpoint.  

I tested this by modifying the scheduler to list the workers as absorbers, then running
./pants run ./binaries/setup-cloud-scoot/main.go --strategy local.local
./pants run ./binaries/scootapi/main.go run_smoke_test 5 (to generate stats)

Then querying http://localhost:9091/admin/metrics.json, stats from scheduler and worker were returned as one json object.


common/endpoints/endpoints.go
```diff
@@ -60,9 +73,56 @@ func (s *TwitterServer) statsHandler(w http.ResponseWriter, r *http.Request) {
...
+				log.Printf("ERROR: Getting Stats for Sidecar at Port %v, returned an error %v", port, err)
+			}
+		}(p)
+	}
+
+	// wait until we have all the stats then combine
+	wg.Wait()
+	close(sidecarStatsCh)
+	for statsRsp := range sidecarStatsCh {
+		json.Unmarshal(statsRsp, &allStats)
```
jasonbs10: Arrays don't keep existing elements but maps do? I'm surprised enough by this behavior that it may warrant a comment despite it being a std lib.

common/endpoints/endpoints.go
```diff
@@ -60,9 +73,56 @@ func (s *TwitterServer) statsHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	// get stats for the sidecars
+	sidecarStatsCh := make(chan []byte, len(s.SidecarStatPorts))
+	var wg sync.WaitGroup
+	for _, p := range s.SidecarStatPorts {
+		wg.Add(1)
+		go func(port int) {
+			defer wg.Done()
+			// sidecars must be running on the same machine and also be twitter servers
+			rsp, err := http.Get(fmt.Sprintf("http://localhost:%v/admin/metrics.json", port))
```
jasonbs10: It might be worth adding a timeout here even though we expect it to just work.

jasonbs10:
Actually, before you merge can you take a look at an example aurora file and its multiple Stat declarations? Does this manage to collate stats from multiple processes?

See:
https://cgit.twitter.biz/gazebo/tree/mesos/gazebo.aurora

JeanetteBruno:
Now that I'm starting to wear a Tool Insights hat.... will we be using TI to collect metrics?  What is the relationship between these stats and Tool Insights stats (duplication? more refined stats here than from Tool Insights? ...)
I think I'm recommending we have an (initial) overall design to the stats that will be collected for Scoot.

common/endpoints/endpoints.go
CaitieM20: yup it just adds the new keys, i'll add a comment

common/endpoints/endpoints.go
CaitieM20: yes good idea, done

common/endpoints/endpoints.go
CaitieM20: yup arrays just add new elements to the list, added a comment

CaitieM20:
@jasonbs10 I'll try that today it looks like it does.  I didn't know that multiple absorbers were supported (may be a new feature)

@JeanetteBruno ToolInsights is for collecting metrics from developer laptops, we don't have a pipe that does that.  For services running on machines in twitter DCs, and in external zones, like AWS we use the Twitter Observability stack.  [Observability at Twitter Part 1](https://blog.twitter.com/2016/observability-at-twitter-technical-overview-part-i) & [Observability at Twitter Part 2](https://blog.twitter.com/2016/observability-at-twitter-technical-overview-part-ii)
Ping me if you want links to internal docs as well.

CaitieM20:
@jasonbs10 The double aurora undocumented trick worked!  going to close this pull request and just update the aurora file!

PR #204, opened 2017-01-19
snapshot/git/gitdb: refactor to meet improved API/design

dbentley:
(as per design doc)

This doesn't change any functionality, but sets us up to be consistent.


snapshot/git/gitdb/local_data.go
```diff
@@ -2,24 +2,26 @@ package gitdb
...
-	"github.com/scootdev/scoot/snapshot/git/repo"
 )
 
 const localIDText = "local"
 const localIDFmt = "%s-%s-%s"
 
-// localValue holds a reference to a value that is in the local DB
-type localValue struct {
+type localBackend struct {
+	db *DB
```
jasonbs10: Where is db set and is a specific impl of db required here?

snapshot/git/gitdb/backends.go
```diff
@@ -0,0 +1,62 @@
...
+	if !kinds[kind] {
+		return nil, fmt.Errorf("invalid kind: %s", kind)
+	}
+
+	switch backendType {
+	case localIDText:
+		return db.local.parseID(id, kind, parts[2:])
+	case streamIDText:
+		return db.stream.parseID(id, kind, parts[2:])
+	default:
```
jasonbs10: TODO: bundlestoreBackend / gittagBackend?

snapshot/git/gitdb/README.md
```diff
@@ -4,32 +4,32 @@ GitDB is an implementation of ScootDB (snapshot.DB) that stores local values in
...
-## Schemes
-GitDB uses different Schemes to identify, upload and download Values depending on their provenance. (Like how Chrome treats http: URLs different from file: URLs.) Local storage is the same among Kinds (an object in the Git ODB), but download and upload are different. Schemes are orthogonal to Value Kinds (like how html vs. jpg is orthogonal to http: vs. file:). GitDB encodes the Scheme and other information in the Value IDs.
+## Backends
+GitDB uses different Backends to identify, upload and download Snapshots.
 
-## Value ID format
-GitDB uses Value IDs that encode the scheme, kind and per-scheme data, separated by '-':
-* _<scheme>-<kind>(-<additional information>)+_
+## Snapshot ID format
+GitDB uses Value IDs that encode the backend, kind and per-backend data, separated by '-':
```
dbentley: Done.

snapshot/git/gitdb/backends.go
```diff
@@ -0,0 +1,62 @@
...
+	"strings"
+
+	"github.com/scootdev/scoot/snapshot"
+)
+
+type backend interface {
+	parseID(id string, kind string, parts []string) (snap, error)
+}
+
+type uploader interface {
```
dbentley: Yeah, it's weird, but the alternatives are weirder.

Option 1: Download is on backend, not snap.

Look at checkout.go, where it downloads before checkout: it would be weird to have to get the right backend to pass right after you parsed it.

Option 2: Upload on Value.
Well, it then has to figure out what the right place to upload is, etc.

So, because upload and download are the reverses of one another, they start from opposite places.

snapshot/git/gitdb/backends.go
```diff
@@ -0,0 +1,62 @@
...
+	upload(id snapshot.ID) (snapshot.ID, error)
+	backend
+}
+
+// GitDB uses several different Backends to store Snapshots.
+// snap is the common interface for a Snapshot (identified by an ID)
+type snap interface {
```
dbentley: snapID?

I'm happy to delete snapshot/interface.go for now, along with the FS code.... But maybe that's a separate PR?

This isn't just a generic snapshot, this is the implementation of snapshot in GitDB, so we could call it gitDBSnap, but it's already in package gitdb.

snapshot/git/gitdb/backends.go
```diff
@@ -0,0 +1,62 @@
...
+	ID() snapshot.ID
+
+	Kind() snapKind
```
dbentley: Really? You don't think that's too long? snap is unambiguously snapshot, no?

snapshot/git/gitdb/db.go
```diff
@@ -50,7 +51,9 @@ type DB struct {
 	dataRepo  *repo.Repository
 	tmp       *temp.TempDir
 	checkouts map[string]bool // checkouts stores bare checkouts, but not the git worktree
-	stream    *StreamConfig
+	local     *localBackend
+	stream    *streamBackend
+	remote    uploader // TODO(dbentley): implement uploader
```
dbentley: How's that?

snapshot/git/gitdb/backends.go
dbentley: Eh, I'd rather not commit to a plan in code here; let's just implement it over the next short period?

snapshot/git/gitdb/db.go
```diff
@@ -74,18 +77,21 @@ func (db *DB) loop() {
...
 			req.resultCh <- idAndError{id: id, err: err}
 		case checkoutReq:
 			path, err := db.checkout(req.id)
 			req.resultCh <- stringAndError{str: path, err: err}
 		case releaseCheckoutReq:
 			req.resultCh <- db.releaseCheckout(req.path)
-		case downloadReq:
-			id, err := db.download(req.id)
-			req.resultCh <- idAndError{id: id, err: err}
 		default:
```
dbentley: download is functional: it's now called as part of checkout.

snapshot/git/gitdb/local_data.go
dbentley: Unneeded; removed.

jasonbs10:
I have some thoughts forming about the gitdb design. It seems possible and desirable to do away with a bunch of types -  localSnap/streamSnap/localBackend/streamBackend/backend/snapshot - and minimize 'kind'.

Can we instead focus on the actions we're taking:
Creator.IngestDir(dir) -> tree sha, err
Creator.IngestCommit(commitish) -> commit sha, err

snapshot.ID = [bundlestore|stream]-sha-additionalInfo
-- Do we need 'kind' when bundle is not yet local? And when it's local can't we inspect the object? 

Uploader.Upload(sha) -> snapshot.ID
 -- The uploader impl uses a specific backend. We can have a wrapper with multiple Uploaders.
Downloader.Download(snapshot.ID) -> err
-- Downloader impl supports a specific backend. We'd have a wrapper with multiple Downloaders.
Reader.Checkout(snapshot.ID) -> dir, kind, err

- Inspect the local sha and do the right thing on upload/checkout.
- DB becomes the facade for Reader/Creator/Uploader/Downloader.

What other actions do we need?

snapshot/db.go
```diff
@@ -4,43 +4,28 @@ import (
...
+// ID identifies a Snapshot in DB. (Cf. doc.go for explanation of Snapshot) Opaque to the client.
 type ID string
 
-// DB is the Scoot Database, allowing creation, distribution, and export of Values.
-type DB interface {
+// Creator allows creating new Snapshots.
+type Creator interface {
 	// Ingest
 
 	// IngestDir ingests a directory directly.
```
JeanetteBruno: these comments are defining the terms with the term, eg. IngestDir ingests a directory... can we modify it to define ingest?  eg: IngestDir creates a new snapshot containing a copy of the files in dir.  ?The root of the snapshot will be dir?  

Also - what is FSSnapshot?  where can I find the definition for this?

This comment also applies to the definition of IngestCommit

snapshot/doc.go
```diff
@@ -2,13 +2,12 @@
 package snapshot offers access to Scoot data.
 Scoot data is largely snapshots (hence the package name).
```
JeanetteBruno: Can a snapshot be anything other than a set of files?  (and I think of 'Scoot data' as a lot more than a snapshot - the term Scoot data in my mind includes the job definitions, saga data, snapshots, etc.)
Alternate definition:
'A Scoot snapshot is a set of files that will be made local to the environment in which a scoot command is run and/or it is the set of files that are preserved after a Scoot command is finished.'

snapshot/doc.go
```diff
@@ -2,13 +2,12 @@
...
 
-The main entry point is the interface DB. DB holds Values which can be a Snapshot or a SnapshotWithHistory.
+The main entry point is the interface DB. DB holds Snapshots which can be an FSSnapshot or a GitCommitSnapshot.
 
-A Snapshot is a snapshot of filesystem state (like a tar file).
+An FSSnapshot is a snapshot of filesystem state (like a tar file).
 
-A SnapshotWithHistory is a Snapshot and also a commit message and an optional list of parent SnapshotWithHistory's.
-(This definition corresponds with a git commit, which is convenient for creating git repos.)
+A GitCommitSnapshot mirrors a git commit: a Snapshot, commit metadata, and an optional list of parent GitCommitSnapshots.
```
JeanetteBruno: not sure how to interpret 'mirrors a git commit': 
Is a GitCommitSnapshot a copy of the selected files as they exist at a targeted git commit or is GitCommitSnapshot a copy of the git change history  (git db) as it exists at a commit? (It is my understanding that it is the copy of the files, not the git db).

snapshot/db.go
```diff
@@ -4,43 +4,28 @@ import (
...
-
-	// Download makes sure the Value id is downloaded, returning an ID that can be used
-	// on this computer or an error
-	Download(id ID) (ID, error)
-
-	// Export
-
-	// Checkout puts the Value identified by id in the local filesystem, returning
+// Reader allows reading data from existing Snapshots
+type Reader interface {
```
JeanetteBruno: on the fence as to whether this should be called Checkouter vs Reader.... if all methods use the word 'checkout', then I would lean toward Checkouter.

snapshot/git/gitdb/backends.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"fmt"
+	"strings"
+
+	snap "github.com/scootdev/scoot/snapshot"
+)
+
+// backend allows getting a snapshot for an ID, which can then be used to download the ID
+type backend interface {
+	parseID(id string, kind string, parts []string) (snapshot, error)
```
JeanetteBruno: should kind be constrained to a 'backendType' struct or interface? (be more explicit about what is allowed)

snapshot/db.go
dbentley: Great suggestion; done.

snapshot/doc.go
dbentley: Updated a bit...

1) In the Doc, we agreed that Snapshot can be both FSSnapshot and GitCommitSnapshot, so this reflects that.

2) I described that you can use them as the input to a Task, and we intend to store them as the output, but not that that's the exclusive use (you can use Scoot to distribute Snapshots without using CloudExec at all!)

snapshot/doc.go
dbentley: It also includes the history, because various tools that want to use Scoot for data do need the history (this was covered in the SnapshotDB/GitDB design doc)

snapshot/db.go
dbentley: They will not all use Reader forever (Cf. the Design Doc)

snapshot/git/gitdb/backends.go
dbentley: Done.

dbentley:
Merged, but still happy to change things

PR #205, opened 2017-01-20
Initial prep for incorporating gitdb into workerserver.

jasonbs10:
I did preliminary testing and this appears to work for checking out a revision at worker and uploading the result. Although, some of the logging is questionable in that it's not quite exactly doing what I thought it would...

Also, we may need to rework the gitdb cli wrt the changes I made there.

------Sequence----
% create a new empty repo and start a worker that'll use that repo for checkouts:
go run binaries/workerserver/main.go --repo=/tmp/scootClient --bundlestore=/tmp

% create a new repo with a couple commits, cd there then ingest one of those commits:
go run ~/scoot/binaries/scoot-snapshot-db/main.go create ingest_git_commit --commit COMMITISH --bundlestore /tmp

% run 'ls' on the worker with the bundle name printed out by ingesting above:
go run binaries/workercl/main.go run --addr=localhost:9090 --snapshotid SNAPSHOT ls

% Unbundle the the result bundle printed by workerserver:
git bundle unbundle BUNDLE_PATH

% Checkout the sha printed by unbundle, then verify contents of STDOUT/STDERR files.
git checkout SHA


runner/runners/invoke.go
```diff
@@ -131,17 +134,16 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 
 	switch st.State {
 	case execer.COMPLETE:
-		srcToDest := map[string]string{
-			stdout.AsFile(): "STDOUT",
-			stderr.AsFile(): "STDERR",
+		tmp, err := temp.NewTempDir("", "invoke")
```
JeanetteBruno: side comment: we want the temp directory to always resolve to os.TempDir() - should we remove the directory param from temp.NewTempDir()?

dgassaway: If this is running on a worker that runs on Mesos and I think it is, we want to make sure we _never_ use the system temp dir, and be sure that everything we create is in our task sandbox ("./" more or less). By passing "" we are using os.TempDir, which is bad and we should scrub these from our repo.

jasonbs10: My preferred solution is to set $TMPDIR to the desired location in our aurora deploy. The code shouldn't have to know that there are system limitations. WDYT?

dgassaway: SGTM

JeanetteBruno: right - it was my understanding that we set $TMPDIR in aurora and os.TEMPDIR() uses $TMPDIR when it is available.  So our current implementations are safe.    Just recommending that we eliminate the parameter since we really don't want it to be set to anything besides ""/os.TEMPDIR().  (I only realized just recently realized that sending in a non-"" dir would be bad.)

jasonbs10:
I'm choosing the path of least resistance where I can and I think inverting the adapter would be more trouble, admittedly minor, than it's worth at this point. Once we're fully satisfied with the state of db/gitdb then we can make the switch.

I'll probably pull in your changes and do some additional plumbing in this PR. Is there anything that needs to be changed in its current state?

dbentley:
Why have the repo flag if workerserver doesn't today? That's my big concern...

jasonbs10:
I did preliminary testing and this appears to work for checking out a revision at worker and uploading the result. Although, some of the logging is questionable in that it's not quite exactly doing what I thought it would...

------Sequence----
# create a new empty repo and start a worker that'll use that repo for checkouts:
go run binaries/workerserver/main.go --repo=/tmp/scootClient --bundlestore=/tmp

# create a new repo with a couple commits, cd there then ingest one of those commits:
go run ~/scoot/binaries/scoot-snapshot-db/main.go create ingest_git_commit --commit COMMITISH --bundlestore /tmp

# run 'ls' on the worker with the bundle name printed out by ingesting above:
go run binaries/workercl/main.go run --addr=localhost:9090 --snapshotid SNAPSHOT ls

# Unbundle the the result bundle printed by workerserver:
git bundle unbundle BUNDLE_PATH

# Checkout the sha printed by unbundle, then verify contents of stdout/stderr files.
git checkout SHA

runner/runners/invoke.go
dbentley: This discussion is all correct, but also I think it might be easier to have Invoker also hold a tmp instead of needing to create a new one right here.

runner/runners/invoke.go
```diff
@@ -133,17 +136,16 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
...
-		// if cmd.SnapshotPlan != nil {
-		// 	for src, dest := range cmd.SnapshotPlan {
-		// 		srcToDest[checkout.Path()+"/"+src] = dest // manually concat to preserve src *exactly* as provided.
-		// 	}
-		// }
-		snapshotID, err := inv.filer.IngestMap(srcToDest)
+		defer os.RemoveAll(tmp.Dir)
+		outPath := stdout.AsFile()
+		errPath := stderr.AsFile()
+		os.Link(outPath, filepath.Join(tmp.Dir, "STDOUT"))
```
dbentley: we haven't talked much about the semantic of links (either hard or soft) in IngestDir, so I'd rather, for now, just copy (and yes, that'll be a slight performance hit)

snapshot/cli/cli.go
```diff
@@ -41,13 +41,14 @@ import (
...
 
+	"github.com/scootdev/scoot/os/temp"
 	"github.com/scootdev/scoot/snapshot"
 	"github.com/scootdev/scoot/snapshot/git/repo"
 )
 
 type DBInjector interface {
 	// TODO(dbentley): we probably want a way to register flags
-	Inject() (snapshot.DB, error)
+	Inject(storeDir *temp.TempDir) (snapshot.DB, error)
```
dbentley: Instead of this, add register flags, and then it can happen all in main!

snapshot/cli/cli.go
```diff
@@ -80,23 +77,34 @@ func MakeDBCLI(injector DBInjector) *cobra.Command {
...
 	return cmd
 }
 
-func (c *ingestGitCommitCommand) run(db snapshot.DB, _ *cobra.Command, _ []string) error {
+func (c *ingestGitCommitCommand) run(injector DBInjector, _ *cobra.Command, _ []string) error {
+	var storeDir *temp.TempDir
+	if c.store != "" {
+		storeDir = &temp.TempDir{Dir: c.store}
+	}
+	db, err := injector.Inject(storeDir)
```
dbentley: Yeah, this can all go in code in main...

snapshot/git/gitdb/git_tags.go
```diff
@@ -27,6 +28,10 @@ type tagsBackend struct {
 const tagsIDText = "tags"
 
 func (b *tagsBackend) parseID(id snap.ID, kind snapshotKind, extraParts []string) (snapshot, error) {
+	if b.cfg == nil {
+		return nil, errors.New("Stream backend not initialized.")
```
dbentley: tags.

snapshot/git/gitdb/bundlestore.go
```diff
@@ -26,6 +27,10 @@ const bundlestoreIDText = "bs"
 // "bs-gc-<bundle>-<stream>-<sha>"
 
 func (b *bundlestoreBackend) parseID(id snap.ID, kind snapshotKind, extraParts []string) (snapshot, error) {
+	if b.cfg == nil {
+		return nil, errors.New("Stream backend not initialized.")
```
dbentley: bundlestore

snapshot/git/repo/repo.go
```diff
@@ -75,15 +76,26 @@ func validateSha(sha string) (string, error) {
...
 	}
 	topLevel = strings.Replace(topLevel, "\n", "", -1)
 	log.Println("git.NewRepository:", dir, topLevel)
-	repo.dir = topLevel
-	return repo, nil
+	r.dir = topLevel
+	return r, nil
+}
+
+// Try to initialized a new git repo in the given directory.
```
dbentley: s/initialized/initialize/

snapshot/cli/cli.go
dbentley: Cf. https://github.com/scootdev/scoot/commit/330ab36b0132c3349aa6d4721a36b7fa40f5eb1a

binaries/apiserver/main.go
```diff
@@ -0,0 +1,68 @@
...
+				scootapi.CloudScootClientConfig{
+					Addr:   *backendAddr,
+					Dialer: dialer.NewSimpleDialer(transportFactory, protocolFactory),
+				})
+		},
+	)
+	go bundlestore.RunServer(bag, schema, configText)
+
+	// Start a scheduler frontend in the current goroutine and skip the observability server.
+	bag, schema = frontend.Defaults()
```
dbentley: So I think this will make these not share anything, including a stats receiver, so we'll essentially lose observability...

Can I suggest that we make it a bundlestore server *first*, and separate the work of making it be an API server? (Also allows us to sidestep review of that code, which I think isn't resilient to, e.g., scheduler restarts, which is the whole point)

config/scootconfig/workers_modules.go
```diff
@@ -69,7 +70,13 @@ type WorkersLocalConfig struct {
...
 
 func (c *WorkersLocalConfig) Install(bag *ice.MagicBag) {
-	bag.Put(func() worker.WorkerFactory {
-		return workers.MakeInmemoryWorker
+	bag.Put(func(tmp *temp.TempDir) worker.WorkerFactory {
+		return InmemoryWorkerFactory(tmp)
 	})
 }
+
+func InmemoryWorkerFactory(tmp *temp.TempDir) worker.WorkerFactory {
```
dbentley: might be easier to have WorkerFactory be a function that takes a tmp as well?

jasonbs10: I thought about it, and while it would've been easier, it's not needed by thrift worker and I think it's cleaner to curry per-type vars in this case.

PR #206, opened 2017-01-20
snapshot/git/gitdb: implement tags backend

dbentley:
Tags backend uploads and downloads snapshots as git tags.

Enable Auto-Uploading in GitDB. When a Snapshot is created, it can be auto-uploaded to a remote (that is an uploader).

Change some (internal-to-GitDB) code to make more sense and be more composable. (e.g., now we return a snapshot instead of a snap.ID)

Expand db_test to have more DBs in the system to test distribution.

rename {local,stream}Snap -> {local,stream}Snapshot (because I forgot to when I renamed snap -> snapshot; oops!)


snapshot/git/gitdb/db.go
```diff
@@ -23,16 +23,34 @@ var kinds = map[snapshotKind]bool{
...
 		checkouts: make(map[string]bool),
 		local:     &localBackend{},
 		stream:    &streamBackend{cfg: stream},
+		tags:      &tagsBackend{cfg: tags},
+	}
+
+	switch autoUploadDest {
+	case AutoUploadNone:
+	case AutoUploadTags:
+		result.remote = result.tags
```
jasonbs10: Can this from from remote -> autoUploadBackend?

snapshot/git/gitdb/git_tags.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	snap "github.com/scootdev/scoot/snapshot"
+)
+
+type TagsConfig struct {
+	Name   string
```
dbentley: Argh, the IDs were too long because we were using Remote instead of Name; thanks; fixed.

snapshot/git/gitdb/git_tags.go
```diff
@@ -0,0 +1,92 @@
...
+
+type tagsBackend struct {
+	cfg *TagsConfig
+}
+
+const tagsIDText = "tags"
+
+// tags-sss-<sha>
+
+func (b *tagsBackend) parseID(id snap.ID, kind snapshotKind, parts []string) (snapshot, error) {
```
dbentley: Updated (in all backends)

snapshot/git/gitdb/db_test.go
```diff
@@ -119,26 +117,61 @@ func TestStream(t *testing.T) {
...
 
+func TestTags(t *testing.T) {
+	// Create a commit in external, then ingest into authorDB, then check it out in our regular db
+
+	externalCommitID, err := commitText(fixture.tags, "tags_first")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	id, err := fixture.authorDB.IngestGitCommit(fixture.tags, externalCommitID)
```
dbentley: The public (exported) method still returns an ID; it's ingestGitCommit that now returns a snapshot.

snapshot/git/gitdb/db_test.go
```diff
@@ -228,27 +261,79 @@ func setup() (*dbFixture, error) {
 		return nil, err
 	}
 
+	tags, err := createRepo(tmp, "tags-repo")
```
dbentley: Fixed (though in a slightly different way)

You really don't trust the tmp reaper, huh? I expect that a Unix system will clean its tmp dir...

snapshot/git/gitdb/db.go
dbentley: I'm sorry, I don't understand.

jasonbs10: s/Can this/Can this be renamed/

PR #207, opened 2017-01-21
snapshot/cli: cli to snapshot db

dbentley:
scoot-snapshot-db create ingest_git_commit 545c88d71d40a49ebdfb1d268c724110793330d2

gives stdout of:

local-gc-545c88d71d40a49ebdfb1d268c724110793330d2

(don't yet have uploading turned on in this branch)

### DIFF
--- a/binaries/scoot-snapshot-db/main.go
+++ b/binaries/scoot-snapshot-db/main.go
@@ -35,7 +35,7 @@ func (i *injector) Inject() (snapshot.DB, error) {
 	dataRepo, err := repo.NewRepository(wd)
 	if err != nil {
 		return nil, fmt.Errorf(
-			"cannot create a repo in wd %v; scoot-snapshot-db must be run in a git repo", wd, err)
+			"cannot create a repo in wd %v; scoot-snapshot-db must be run in a git repo: %v", wd, err)
 	}
 
 	return gitdb.MakeDB(dataRepo, tempDir, nil), nil
--- a/binaries/scoot-snapshot-db/main.go
+++ b/binaries/scoot-snapshot-db/main.go
@@ -1,0 +1,44 @@
+package main
+
+import (
+	"log"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/scootdev/scoot/os/temp"
+	"github.com/scootdev/scoot/snapshot"
+	"github.com/scootdev/scoot/snapshot/cli"
+	"github.com/scootdev/scoot/snapshot/git/gitdb"
+	"github.com/scootdev/scoot/snapshot/git/repo"
+)
+
+func main() {
+	inj := &injector{}
+	cmd := cli.MakeDBCLI(inj)
+	if err := cmd.Execute(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+type injector struct{}
+
+func (i *injector) RegisterFlags(rootCmd *cobra.Command) {}
+
+func (i *injector) Inject() (snapshot.DB, error) {
+	tempDir, err := temp.TempDirDefault()
+	if err != nil {
+		return nil, err
+	}
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
+
+	dataRepo, err := repo.NewRepository(wd)
+	if err != nil {
+		return nil, err
+	}
+
+	return gitdb.MakeDB(dataRepo, tempDir, nil), nil
+}
--- a/binaries/scoot-snapshot-db/main.go
+++ b/binaries/scoot-snapshot-db/main.go
@@ -1,10 +1,9 @@
 package main
 
 import (
+	"fmt"
 	"log"
 	"os"
-
-	"github.com/spf13/cobra"
 
 	"github.com/scootdev/scoot/os/temp"
 	"github.com/scootdev/scoot/snapshot"
@@ -23,8 +22,6 @@ func main() {
 
 type injector struct{}
 
-func (i *injector) RegisterFlags(rootCmd *cobra.Command) {}
-
 func (i *injector) Inject() (snapshot.DB, error) {
 	tempDir, err := temp.TempDirDefault()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


snapshot/cli/cli.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	createCmd := &cobra.Command{
+		Use:   "create",
+		Short: "create a snapshot",
+	}
+	rootCmd.AddCommand(createCmd)
+	parentCmd = createCmd
+
+	add(&ingestGitCommitCommand{})
```
jasonbs10: Add() should probably take cobra parentCmd as a param and return the new cobra cmd.

snapshot/cli/cli.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func MakeDBCLI(injector DBInjector) *cobra.Command {
+	rootCmd := &cobra.Command{
+		Use:   "scoot-snapshot-db",
+		Short: "scoot snapshot db CLI",
+	}
+	parentCmd := rootCmd
+	add := func(subCmd dbCommand) {
+		cmd := subCmd.register()
```
jasonbs10: Kind of confusing to switch between dbCommand and cobra.Command without tagging the vars as such.

binaries/scoot-snapshot-db/main.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/scootdev/scoot/os/temp"
+	"github.com/scootdev/scoot/snapshot"
+	"github.com/scootdev/scoot/snapshot/cli"
+	"github.com/scootdev/scoot/snapshot/git/gitdb"
+	"github.com/scootdev/scoot/snapshot/git/repo"
+)
+
+func main() {
```
jasonbs10: There should be a help string stating that cwd must be a the git repo (even if it's implied since there's no other way to specify it)

binaries/scoot-snapshot-db/main.go
```diff
@@ -0,0 +1,44 @@
...
+	inj := &injector{}
+	cmd := cli.MakeDBCLI(inj)
+	if err := cmd.Execute(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+type injector struct{}
+
+func (i *injector) RegisterFlags(rootCmd *cobra.Command) {}
```
jasonbs10: I believe this is an implementation review and if that's the case, we should remove things that aren't obviously needed (or add a TODO if you wish).

snapshot/cli/cli.go
```diff
@@ -0,0 +1,84 @@
...
+	cmd := &cobra.Command{
+		Use:   "ingest_git_commit",
+		Short: "ingest a git commit",
+	}
+	cmd.Flags().StringVar(&c.commit, "commit", "", "commit to ingest")
+	return cmd
+}
+
+func (c *ingestGitCommitCommand) run(db snapshot.DB, _ *cobra.Command, _ []string) error {
+	wd, err := os.Getwd()
```
jasonbs10: Might be better to construct this command with a repoDir.

binaries/scoot-snapshot-db/main.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (i *injector) Inject() (snapshot.DB, error) {
```
jasonbs10: Why do we need an injector? We may want to delay db construction but we can't we instead pass the necessary params into MakeDBCLI()?

binaries/scoot-snapshot-db/main.go
dbentley: Done.

binaries/scoot-snapshot-db/main.go
dbentley: So, I maybe buried the lede:

In all our other CLI tools, we've had a problem trying to mix args. So we've had to setup first, then call the CLI, and trying to combine has gotten messy (especially with the --addr flags).

This is nifty because it allows us to do everything in cobra: the injector can register flags (though doesn't right now), and then those get parsed naturally by cobra.

So, at some point (and soon), we're going to want the injector to have args, and it's going to be able to; yay!

To summarize:
the injector doesn't just delay db construction; it allows getting args from the CLI, using them to make objects, and *then* calling into the CLI's shared logic, across our open-source and closed-source binaries.

binaries/scoot-snapshot-db/main.go
dbentley: Hrm, not sure exactly how to it that in...

I added:

	dataRepo, err := repo.NewRepository(wd)
	if err != nil {
		return nil, fmt.Errorf(
			"cannot create a repo in wd %v; scoot-snapshot-db must be run in a git repo", wd, err)
	}


WDYT?

snapshot/cli/cli.go
dbentley: OK, I started writing a github PR comment about when what happens, and realized many of your comments were because this is poorly explained. I just left a big godoc package comment on the top of cli.go; why don't you take a look and we can go from there?

snapshot/cli/cli.go
dbentley: OK, take parentCmd as a param; not sure we need the returned cobra cmd (will we ever have a non-leaf cmd that also have behavior?)

snapshot/cli/cli.go
dbentley: Yup... Any idea how best to tag them?

dbentley:
OK, I've updated with a lengthy doc comment and a few other things.

I am proud of what good things there are, that I've explained much more in the lengthy doc comment.

A few of your comments I agree with and am not sure how best to fix. I'm planning to merge and hope we can improve it together...

snapshot/cli/cli.go
jasonbs10: Giving all the cobra commands a cobra postfix isn't too horrible.

binaries/scoot-snapshot-db/main.go
jasonbs10: That should be fine actually.

PR #209, opened 2017-01-22
snapshot/git/gitdb: implement bundlestore backend

dbentley:


snapshot/git/gitdb/bundlestore.go
```diff
@@ -0,0 +1,200 @@
...
+// git bundle create 545c88d71d40a49ebdfb1d268c724110793330d2..3060a3a519888957e13df75ffd09ea50f97dd03b
+// instead, we have to write a temporary ref
+const bundlestoreTempRef = "reserved_scoot/bundlestore/__temp_for_writing"
+
+func (b *bundlestoreBackend) uploadLocalSnapshot(s *localSnapshot, db *DB) (sn snapshot, err error) {
+	// the sha of the commit we're going to use as the ref
+	commitSha := s.sha
+
+	// the revList to create the bundle
+	// unless we can find a better merge base, we'll just include the commit
```
jasonbs10: What is the meaning of 'better' here vs finding a merge base at all?

snapshot/git/gitdb/bundlestore.go
```diff
@@ -0,0 +1,200 @@
...
+	case kindGitCommitSnapshot:
+		if db.stream.cfg != nil && db.stream.cfg.RefSpec != "" {
+			streamHead, err := db.dataRepo.Run("rev-parse", db.stream.cfg.RefSpec)
+			if err != nil {
+				return nil, err
+			}
+
+			mergeBase, err := db.dataRepo.RunSha("merge-base", streamHead, commitSha)
+
+			// if err != nil, it just means we don't have a merge-base
```
jasonbs10: Elaborate on what it means to not have a merge-base. We have a desired s.sha (*are we assuming sha is validated beforehand / should we?*) and if it doesn't share an ancestor with streamHead, we create a bundle with only s.sha? When would this happen?

If we have a bunch of exclusively local commits and there's no ancestor in common with streamHead, isn't the generated bundle useless?

snapshot/git/gitdb/bundlestore.go
```diff
@@ -0,0 +1,200 @@
...
+		return nil, err
+	}
+
+	f, err := os.Open(bundleFilename)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	if err := b.cfg.Store.Write(s.sha, f); err != nil {
```
jasonbs10: Why do we want to reference the same resource as both ID in some places and sha here?

snapshot/git/gitdb/bundlestore.go
```diff
@@ -0,0 +1,200 @@
...
+
+	if err := validSha(sha); err != nil {
+		return nil, err
+	}
+
+	return &bundlestoreSnapshot{kind: kind, sha: sha, bundleName: bundleName, streamName: streamName}, nil
+}
+
+func (b *bundlestoreBackend) upload(s snapshot, db *DB) (snapshot, error) {
+	// We only have to upload a localSnapshot
```
jasonbs10: Is this strictly true? Do we never want to upload stream -> bundlestore? See stream question below.

snapshot/git/gitdb/bundlestore.go
```diff
@@ -0,0 +1,200 @@
...
+			mergeBase, err := db.dataRepo.RunSha("merge-base", streamHead, commitSha)
+
+			// if err != nil, it just means we don't have a merge-base
+			if err == nil {
+				revList = fmt.Sprintf("%s..%s", mergeBase, bundlestoreTempRef)
+				streamName = db.stream.cfg.Name
+			}
+
+		}
+	case kindFSSnapshot:
```
jasonbs10: Can you clarify the details of fsSnapshot and CommitSnapshot? What history is the worker expected to have in order to handle the generated bundle? How much gets deltafied? Etc.

snapshot/git/gitdb/bundlestore.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	// we couldn't unbundle
+	// see if it's because we're missing prereqs
+	exitError := err.(*exec.ExitError)
+	if exitError == nil || !strings.Contains(string(exitError.Stderr), "error: Repository lacks these prerequisite commits:") {
+		return err
+	}
+
+	// we are missing prereqs, so let's try updating the stream that's the basis of the bundle
```
jasonbs10: What are the failure scenarios here?

snapshot/git/gitdb/bundlestore.go
dbentley: TODO'ed.

snapshot/git/gitdb/bundlestore.go
```diff
@@ -0,0 +1,200 @@
...
+
+	return &bundlestoreSnapshot{kind: kind, sha: sha, bundleName: bundleName, streamName: streamName}, nil
+}
+
+func (b *bundlestoreBackend) upload(s snapshot, db *DB) (snapshot, error) {
+	// We only have to upload a localSnapshot
+	switch s := s.(type) {
+	case *tagsSnapshot:
+		return s, nil
+	case *streamSnapshot:
```
dbentley: A stream updates; tags don't.

Commits in a stream are related; commits in different tags aren't.

This means that you can just update a stream.

You create a streamSnapshot by pushing something new to master, then creating the ID, and using that.

snapshot/git/gitdb/bundlestore.go
```diff
@@ -0,0 +1,200 @@
...
+	case *bundlestoreSnapshot:
+		return s, nil
+	case *localSnapshot:
+		return b.uploadLocalSnapshot(s, db)
+	default:
+		return nil, fmt.Errorf("cannot upload %v: unknown type %T", s, s)
+	}
+}
+
+// git bundle create takes a rev list; it requires that it include a postiive ref
```
dbentley: Updated.

snapshot/git/gitdb/bundlestore.go
dbentley: Removed.

snapshot/git/gitdb/bundlestore.go
dbentley: The generated bundle would be self-sufficient, just ginormous.

snapshot/git/gitdb/bundlestore.go
dbentley: Commented.

snapshot/git/gitdb/bundlestore.go
dbentley: Where is it ID?

snapshot/git/gitdb/bundlestore.go
dbentley: Commented.

snapshot/git/gitdb/stream.go
```diff
@@ -65,14 +65,21 @@ func (s *streamSnapshot) Download(db *DB) error {
...
-	if _, err := db.dataRepo.Run("fetch", db.stream.cfg.Remote); err != nil {
+	if err := db.stream.updateStream(s.streamName, db); err != nil {
 		return err
 	}
 
 	return db.shaPresent(s.SHA())
 }
+
+// updateStream updates the named stream
+func (b *streamBackend) updateStream(name string, db *DB) error {
```
dbentley: Thanks; updated to use its own.

snapshot/git/gitdb/bundlestore.go
jasonbs10: This seems to be one of the few places that refers to snapshot data without %s-%s-%s. It's one thing to checkout a raw sha1, but I'm not sure it makes sense to do the same elsewhere.

snapshot/git/gitdb/bundlestore.go
jasonbs10: This is good but I also meant the failure scenarios for stream fetch failure. Pretty much just configuration error or git server error?

snapshot/git/gitdb/bundlestore.go
jasonbs10: Still not seeing why tags can't be folded into the stream concept. Tags are a stream that doesn't advance and can't be updated with new data. But they both do the same core thing which is push/pull from repos.

I'm not sure you addressed stream uploading - what's pushing to master in your last sentence? And what's creating the ID? What is the user doing directly with git and also via our CLI?

dbentley: Tags vs. Streams:

We will have the stream in every GitDB. But each tag will only be in some GitDBs. So when we create a bundle, it will base off a stream, which it assumes is easy to get, but not a tag.

Stream uploading:

something outside scoot pushes to master. On the macbook, they'll fetch outside scoot (just call git fetch), and then when they upload, GitDB will realize if the commit is on a stream and if so use that. (that's to be implemented)

snapshot/git/gitdb/bundlestore.go
dbentley: Hoisted out a "makeBundleName" function; wdyt?

snapshot/git/gitdb/bundlestore.go
dbentley: or you're offline or....

The second unbundle would fail because the git server is out-of-date, or offline, or...

I've added some comments.

And yeah, I think this is a higher level of review and I'm not sure I can imagine everything right now.

PR #210, opened 2017-01-23
scootapi/client - Flag for get status to output as json

dgassaway:
Add option for more parseable format


dgassaway:
I thought the same thing- it is better than the json, and it leaves us free
to improve it later with a RunStatus String() function.

PR #213, opened 2017-01-26
snapshot: cross-package changes to get gitdb working end-to-end

dbentley:
snapshotDB:
  add ExportGitCommit command (in db.go, cli, and GitDB)

snapshot/bundlestore:
  better error-checking for uploading
  require bundles to have names like bs-<sha>.bundle

snapshot/filer:
  release checkout when we're done

gitdb/bundlestore:
  pass -c core.packobjectedgesonlyshallow=0 to avoid a bug in our internal git
  fix an error with not parsing a SHA


snapshot/git/gitdb/bundlestore.go
```diff
@@ -132,7 +132,7 @@ func (b *bundlestoreBackend) uploadLocalSnapshot(s *localSnapshot, db *DB) (sn s
 	bundleFilename := path.Join(d.Dir, bundleName)
 
 	// create the bundle
```
jasonbs10: Add the ticket here.

snapshot/bundlestore/http_store.go
```diff
@@ -62,7 +63,21 @@ func (s *httpStore) Write(name string, data io.Reader) error {
...
-	log.Printf("Posted %s, err: %v", uri, err)
-	return err
+	resp, err := s.client.Post(uri, "text/plain", data)
+	log.Printf("Posted %s", uri)
+	if err != nil {
+		log.Printf("Error posting %s, err: %v", uri, err)
+		return err
+	}
+
+	if resp.StatusCode == http.StatusOK {
```
jasonbs10: Did you merge this from my other PR or add it manually? I remember needing to close resp.Body.

snapshot/bundlestore/server.go
```diff
@@ -99,9 +99,10 @@ func (s *Server) HandleDownload(w http.ResponseWriter, req *http.Request) {
 func (s *Server) checkBundleName(name string) (bool, error) {
 	// Matches 3 dash delimited strings and an optional path postfix.
 	// Looks for the first two, then a third which may have additional dashes, and then a path.
```
jasonbs10: Update comment.

snapshot/git/gitdb/checkout.go
```diff
@@ -112,3 +113,52 @@ func (db *DB) releaseCheckout(path string) error {
...
+		return "", err
+	}
+
+	if _, err := db.dataRepo.Run("update-ref", "-d", tempRef); err != nil {
+		return "", err
+	}
+
+	if _, err := externalRepo.Run("update-ref", "-d", tempRef); err != nil {
+		return "", err
+	}
```
jasonbs10: This whole block is very similar to ingestion. Can you make a common fn?

snapshot/cli/cli.go
```diff
@@ -119,3 +127,36 @@ func (c *ingestGitCommitCommand) run(db snapshot.DB, _ *cobra.Command, _ []strin
...
 	return nil
 }
+
+type exportGitCommitCommand struct {
+	id string
+}
+
+func (c *exportGitCommitCommand) register() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "export_git_commit",
```
jasonbs10: I feel like this and ingest have awkward naming since it's trying to be more general than the current main.go requires. Can we get something closer to the intent which is 'import from store' and 'export to store'?

snapshot/git/gitdb/bundlestore.go
dbentley: The ticket is internal, so I'd rather not link it in open source?

snapshot/bundlestore/http_store.go
dbentley: Now merged. Thx.

snapshot/bundlestore/server.go
dbentley: Removed.

snapshot/git/gitdb/checkout.go
dbentley: Good idea. Done.

snapshot/cli/cli.go
dbentley: From store? It's not from store; this is how you get something into scoot or out at all...

snapshot/git/gitdb/bundlestore.go
jasonbs10: That's fine, this needs some comment though.

snapshot/cli/cli.go
jasonbs10: Yes, but the dataRepo and targetRepo are the same for cli, and if the point isn't to interact with store/scoot these operations aren't very meaningful. Am I misreading something?

dbentley: Oh, I see, yeah, you're right...

I hope that eventually scoot-snapshot-db will just be transparently talking to a daemon, so I think the semantics here are reasonable, if awkward for now.

WDYT?

snapshot/git/gitdb/bundlestore.go
dbentley: Duh, yeah, fixed.

snapshot/cli/cli.go
dbentley: to_git_commit

PR #214, opened 2017-01-26
{scoot,worker}api: include generated SnapshotID in thrift

dbentley:


workerapi/api.go
```diff
@@ -98,9 +98,19 @@ func ThriftRunStatusToDomain(thrift *worker.RunStatus) runner.RunStatus {
 	if thrift.ExitCode != nil {
 		domain.ExitCode = int(*thrift.ExitCode)
 	}
+	if thrift.SnapshotId != nil {
+		domain.SnapshotID = *thrift.SnapshotId
+	}
 	return domain
 }
 
+func copyString(s string) *string {
```
jasonbs10: Are we being consistent elsewhere wrt nil and empty string when doing thrift <---> domain? Is this important other than the fact that it makes api_test.go easier?

dbentley: It is important, because if we don't copy like this, we open ourselves to bugs more easily. (Because taking the address of a struct can mean your code changes out from under you if the struct is mutated)

jasonbs10: Do you mind double checking for this PR that the other thrift types are being converted correctly?

dbentley: I looked and couldn't find any.

PR #215, opened 2017-01-27
First pass at apiserver bundlestore caching.

jasonbs10:
This seems to produce reasonable bundlestore cache stats for both local and peer cache hits.

TODO: might need more testing and an architecture graphic.


snapshot/bundlestore/groupcache.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/golang/groupcache"
+	"github.com/scootdev/scoot/cloud/cluster"
+)
+
+type PeerFetcher interface {
+	Fetch() ([]cluster.Node, error)
+}
+
+type GroupcacheConfig struct {
```
JeanetteBruno: comments/docs somewhere would be helpful....need to understand the functionality this code is trying to provide, and what will be using it.

I'm assuming this functionality is to support caching groups of ?snapshots? ?
What is the grouping criteria - where is a group defined?

This GroupcacheConfig the configuration for a group or an entry in the group?
- is Name the name of the Cache that contains a group of things/?snapshots? or the name of one of the things in the group?
- is Memory_btyes - does this imply that the cache contents are stored in memory? then what is AddrSelf and Endpoint?
- are peers other groupCaches or other entries in the cache that this entry is in?

JeanetteBruno:
I created a design doc page from the design alternatives doc.....if you go to the 'Backend Design' link (on the table of contents)  you will see a (rather empty) component architecture graphic.  If you know what the components are, and have the time to fill in the picture, that might help us to better review the implementation.

snapshot/bundlestore/groupcache.go
```diff
@@ -0,0 +1,93 @@
...
+	// The HTTPPool constructor will register as a global PeerPicker on our behalf.
+	poolOpts := &groupcache.HTTPPoolOptions{BasePath: cfg.Endpoint}
+	pool := groupcache.NewHTTPPoolOpts(cfg.AddrSelf, poolOpts)
+	pool.Set(peers...)
+
+	// Create the cache which knows how to retreive the underlying bundle data.
+	var cache = groupcache.NewGroup(cfg.Name, cfg.Memory_bytes, groupcache.GetterFunc(
+		func(ctx groupcache.Context, bundleName string, dest groupcache.Sink) error {
+			log.Print("Not cached, fetching bundle: ", bundleName)
+			if reader, err := underlying.OpenForRead(bundleName); err != nil {
```
dgassaway: This if/else if/else block reads strangely - is this a Go thing? It's not obvious right away that the successful code path should execute OpenForRead then ReadAll then SetBytes. Unless this is some idiom I would rather have the admittedly verbose if err != nil checks broken out.

snapshot/bundlestore/groupcache.go
```diff
@@ -0,0 +1,93 @@
...
+		return nil, err
+	}
+	log.Print("cache: ", s.cache.CacheStats(groupcache.MainCache), s.cache.Stats)
+	return ioutil.NopCloser(bytes.NewReader(data)), nil
+
+}
+
+func (s *groupcacheStore) Exists(name string) (bool, error) {
+	log.Print("Exists() checking for cached bundle: ", name)
+	//TODO: what if it exists but we get an err? Can we get existence without also getting all data?
```
dgassaway: Not sure if this would be a real efficiency improvement, but what about a TruncatingByteSliceSink with zero length?

binaries/apiserver/main.go
```diff
@@ -6,18 +6,21 @@ import (
...
 	"github.com/scootdev/scoot/snapshot/bundlestore"
 	"github.com/scootdev/scoot/snapshot/git/gitdb"
 	"github.com/scootdev/scoot/snapshot/snapshots"
 )
 
 func main() {
-	httpAddr := flag.String("http_addr", "localhost:11101", "http addr to serve on")
+	httpAddr := flag.String("http_addr", scootapi.DefaultApiBundlestore_HTTP, "'host:port' addr to serve http on")
 	configFlag := flag.String("config", "{}", "API Server Config (either a filename like local.local or JSON text")
+	tmpDir := flag.String("tmp", "", "Use this existing dir as the root tmp dir for file read/write.")
```
dbentley: Why not just use $TMPDIR?

binaries/apiserver/main.go
```diff
@@ -29,25 +32,59 @@ func main() {
 		log.Fatal(err)
 	}
 
+	type StoreAndHandler struct {
```
dbentley: I don't know why you need this type; it seems really clumsy. What are you trying to do with it?

binaries/workerserver/main.go
```diff
@@ -36,31 +43,64 @@ func main() {
...
-
-		func(s stats.StatsReceiver, handlers map[string]http.Handler) *endpoints.TwitterServer {
-			return endpoints.NewTwitterServer(endpoints.Addr(*httpAddr), s, handlers)
+		func() (*repo.Repository, error) {
+			return repo.NewRepository(*repoDir)
 		},
-
-		func(tmpDir *temp.TempDir) (snapshot.Filer, error) {
-			if db, err := setup.NewGitDB(tmpDir, *repoDir, *storeHandle); err != nil {
+		func(tmpDir *temp.TempDir) (runners.HttpOutputCreator, error) {
```
dbentley: Is there a reason to do this now, too? It seems unrelated to the change...

sched/worker/workers/makers.go
```diff
@@ -8,21 +8,22 @@ import (
 	"github.com/scootdev/scoot/runner/execer/execers"
 	"github.com/scootdev/scoot/runner/runners"
 	"github.com/scootdev/scoot/sched/worker"
-	"github.com/scootdev/scoot/snapshot"
+	"github.com/scootdev/scoot/snapshot/snapshots"
 )
 
 // Makes a worker suitable for using as an in-memory worker.
-func MakeInmemoryWorker(node cluster.Node, tmp *temp.TempDir, filer snapshot.Filer) worker.Worker {
+// TODO: these are useful for unit tests but really shoudldn't be used for integration tests...
```
dbentley: Errr, I disagree with this. I think they can be very useful for integration tests, because they limit the scope of what we're using. Can I fix these in another PR?

snapshot/bundlestore/groupcache.go
```diff
@@ -0,0 +1,123 @@
...
+	Fetcher      PeerFetcher
+}
+
+// Add in-memory caching to the given store.
+func MakeGroupcacheStore(underlying Store, cfg *GroupcacheConfig) (Store, http.Handler, error) {
+	// Create and initialize peer group.
+	// The HTTPPool constructor will register as a global PeerPicker on our behalf.
+	poolOpts := &groupcache.HTTPPoolOptions{BasePath: cfg.Endpoint}
+	pool := groupcache.NewHTTPPoolOpts("http://"+cfg.AddrSelf, poolOpts)
+	go loop(cfg.Fetcher, pool)
```
dbentley: How does this goroutine stop, fwiw?

snapshot/bundlestore/groupcache.go
```diff
@@ -0,0 +1,123 @@
...
+	Endpoint     string
+	Fetcher      PeerFetcher
+}
+
+// Add in-memory caching to the given store.
+func MakeGroupcacheStore(underlying Store, cfg *GroupcacheConfig) (Store, http.Handler, error) {
+	// Create and initialize peer group.
+	// The HTTPPool constructor will register as a global PeerPicker on our behalf.
+	poolOpts := &groupcache.HTTPPoolOptions{BasePath: cfg.Endpoint}
+	pool := groupcache.NewHTTPPoolOpts("http://"+cfg.AddrSelf, poolOpts)
```
dbentley: So it's configure to populate itself by calling itself? How does that do it?

jasonbs10:
I'm merging this to unblock. I'll revisit this code and add comments in a later PR.

PR #216, opened 2017-01-27
snapshot/snapshots/server.go: serve snapshot viewing

dbentley:
We want to serve viewing snapshot data. This replaces browse_store.

This means that our apiserver now wants to serve several methods, so rework the injection there. Now we will just serve via endpoints, which creates an HTTP Server. We do this by injecting (in main.go) a map[string]http.Handler. For us, our two Handlers (for now) are bundlestore.Server and snapshots.ViewServer.

Add file reading to Snapshot.DB interface, and implement in GitDB.

Add file reading and ingest dir to CLI.

Remove browse_store


snapshot/bundlestore/README.md
```diff
@@ -11,13 +11,7 @@ in memory, bundles on disk, and bundles located behind an http server. Further,
 wrap these stores in a parent store to add additional business logic to our handling.
 
 ## Bundle name conventions
-For now names look like '%s-%s-%s', where the third string can include additional dashes.
-A path may be appended to this name and is recognized to start at the first occurence of '/'.
-This path is only allowed when reading a bundle, otherwise it's considered an error if included.
-
-The '/' is not allowed in bundle names except at the end where it signifies the start
```
jasonbs10: We should undelete/relocate some of this. It's only clear in subsection of the code that bundles shouldn't have '/' in the name. Also keep the nil basis comment even if you've added appropriate error cases already.

binaries/apiserver/main.go
```diff
@@ -28,17 +29,22 @@ func main() {
 		log.Fatal(err)
 	}
 
-	// Start a new goroutine for bundlestore server as well as the observability server.
-	bag, schema := bundlestore.Defaults()
+	bag := ice.NewMagicBag()
+	schema := jsonconfig.EmptySchema()
+	bag.InstallModule(temp.Module())
```
dbentley: I think it'll go too stale, ultimately, if you want to see what's in a module, it's easier to go look at it than try and comment on every use.

binaries/scoot-snapshot-db/main.go
```diff
@@ -49,15 +44,8 @@ func (i *injector) Inject() (snapshot.DB, error) {
...
-		storeTmp, err := tempDir.TempDir("bundlestore")
-		if err != nil {
-			return nil, err
-		}
-		i.storeDir = storeTmp.Dir
-	}
-
-	store, err := bundlestore.MakeFileStore(i.storeDir)
+	_, api := scootapi.GetScootapiAddr()
+	store, err := bundlestore.MakeHTTPStore(fmt.Sprintf("http://%s/bundle/", api)), nil
```
dbentley: Done (and moved it to locate.go)

common/endpoints/setup.go
```diff
@@ -0,0 +1,51 @@
...
+
+type module struct{}
+
+// Install installs functions for serving HTTP
+func (m module) Install(b *ice.MagicBag) {
+	b.PutMany(
+		func(scope StatScope) stats.StatsReceiver {
+			return MakeStatsReceiver(scope).Precision(time.Millisecond)
+		},
+		NewTwitterServer,
```
dbentley: I agree in general; I'd say ObservableServer?

In the interest of not getting hung up on naming, especially given how little this is actually touching that function, I've TODO'ed it. That work for you?

common/endpoints/setup.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		log.Fatal("Error configuring Scoot API: ", err)
+	}
+
+	// Initialize Objects Based on Config Settings
+	bag.InstallModule(mod)
+
+	// Run Servers
+	var server *TwitterServer
+	err = bag.Extract(&server)
```
dbentley: I think reflection might not, and also others need both servers still, because thrift isn't http.

os/temp/temp_dir.go
```diff
@@ -75,3 +77,13 @@ func TempDirHere() (*TempDir, error) {
 
 	return NewTempDir(cwd, "scoot-tmp-")
 }
+
```
dbentley: I figured that this was so minimal that it would be preferable to keep this as a one-file package. WDYT?

scootapi/setup/workers.go
```diff
@@ -118,6 +118,9 @@ func NewGitDB(tmpDir *temp.TempDir, repoDir, storeAddr string) (snapshot.DB, err
 		}
 	} else {
 		r, err = repo.NewRepository(repoDir)
+		if err != nil {
```
dbentley: Yes, it probably should, but I really don't want to pull on that string in this PR.

As to style, golang goes the other way:
https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

snapshot/bundlestore/setup.go
```diff
@@ -1,82 +1,16 @@
...
 )
 
-type servers struct {
-	bs  *Server
-	obs *endpoints.TwitterServer
-}
-
-func makeServers(bs *Server, obs *endpoints.TwitterServer) servers {
-	return servers{bs, obs}
+func Module() ice.Module {
```
dbentley: Done.

snapshot/cli/cli.go
```diff
@@ -119,3 +128,51 @@ func (c *ingestGitCommitCommand) run(db snapshot.DB, _ *cobra.Command, _ []strin
...
+}
+
+type catCommand struct {
+	id string
+}
+
+func (c *catCommand) register() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "cat",
+		Short: "catenate file from an FSSnapshot to stdout",
```
dbentley: ~/work/src/github.com/scootdev/scoot [dbentley/view_server] 02:28:39 0 $ man cat | head -n5

CAT(1)                    BSD General Commands Manual                   CAT(1)

NAME
     cat -- concatenate and print files

snapshot/cli/cli.go
```diff
@@ -119,3 +128,51 @@ func (c *ingestGitCommitCommand) run(db snapshot.DB, _ *cobra.Command, _ []strin
...
+	}
+	cmd.Flags().StringVar(&c.id, "id", "", "Snapshot ID to read from")
+	return cmd
+}
+
+func (c *catCommand) run(db snapshot.DB, _ *cobra.Command, filenames []string) error {
+	id := snapshot.ID(c.id)
+	for _, filename := range filenames {
```
dbentley: It was to emulate the behavior of cat, to make this easier to script and compose.

snapshot/git/gitdb/setup.go
```diff
@@ -0,0 +1,54 @@
...
+		func() *StreamConfig {
+			return nil
+		},
+		func() *TagsConfig {
+			return nil
+		},
+		func() AutoUploadDest {
+			return AutoUploadBundlestore
+		},
+		MakeDBNewRepo,
```
dbentley: That may be true now, but I'm not sure it always will be.

And no, I think not necessarily defaults for everything...

I'd rather not try to overfit on too little data rn.

snapshot/snapshots/setup.go
```diff
@@ -0,0 +1,17 @@
+package snapshots
+
+import (
+	"github.com/scootdev/scoot/ice"
+)
+
+// Module returns a module to allow serving Snapshots
```
dbentley: Done.

snapshot/bundlestore/README.md
dbentley: Moved to main.go of apiserver, because that's not a limitation of bundlestore package but of our configuration.

common/endpoints/setup.go
jasonbs10: Extract already takes an interface{} so it seems like we could pass 'server *TwitterServer' into a utility function for extraction?

os/temp/temp_dir.go
jasonbs10: Best judgement I guess, when something is too big it has to be split out, this is tiny though.

common/endpoints/setup.go
jasonbs10: Yep

snapshot/cli/cli.go
jasonbs10: Right, but that's concatenate? I've never seen catenate before, just commenting that it may look weird to some.

dbentley: Dammit.

Yes, catenate is an obscure and dated synonym for concatenate which is notable now only for being the source of the name "cat". I've updated the comment to use concatenate.

PR #217, opened 2017-01-31
common/resolver: resolve an address or url

dbentley:


scootapi/locate.go
```diff
@@ -24,6 +24,9 @@ func GetScootapiAddrPath() string {
 func GetScootapiAddr() (sched string, api string, err error) {
 	data, err := ioutil.ReadFile(GetScootapiAddrPath())
 	if err != nil {
+		if os.IsNotExist(err) {
+			return "", "", nil
```
jasonbs10: When do we want empty strings *and* nil?

dbentley: The same way ReadFile returns an error, but not EOF, this shouldn't error if the file doesn't exist, because it's trying to read there. It *should* error if, I dunno, it fails to read because your hard drive died in the middle.

common/dialer/resolver.go
```diff
@@ -0,0 +1,62 @@
...
+
+// NewCompositeResolves creates a new CompositeResolve that resolves by looking through delegates (in order)
+func NewCompositeResolver(dels ...Resolver) *CompositeResolver {
+	return &CompositeResolver{dels: dels}
+}
+
+// Resolve resolves by resolving, in order, via delegates
+func (r *CompositeResolver) Resolve() (string, error) {
+	for _, r := range r.dels {
+		if s, err := r.Resolve(); s != "" || err != nil {
```
dgassaway: Behavior not totally clear above - Resolvers are expected to return ""/nil if they didn't resolve?  And we don't want to fall through to the next resolver if a prior one errors?

dbentley: Commented.

PR #218, opened 2017-02-01
Groupcache should take cluster instead of fetcher.

jasonbs10:


snapshot/bundlestore/groupcache.go
```diff
@@ -61,26 +59,25 @@ func MakeGroupcacheStore(underlying Store, cfg *GroupcacheConfig) (Store, http.H
...
-	for {
-		peerNodes, err := fetcher.Fetch()
-		if err != nil {
-			log.Print("Unable to fetch groupcache peers: ", err)
-			continue
-		}
+// Loop will listen for cluster updates and create a list of peer addresses to update groupcache.
+// Cluster is expected to include the current node.
+func loop(c *cluster.Cluster, pool *groupcache.HTTPPool) {
+	toPeers := func(nodes []cluster.Node) []string {
```
dbentley: I'd make this a function at the top level if you aren't using any variables witin...

PR #219, opened 2017-02-01
Added apiserver/bundlestore stats.

jasonbs10:


snapshot/bundlestore/groupcache.go
```diff
@@ -56,62 +61,90 @@ func MakeGroupcacheStore(underlying Store, cfg *GroupcacheConfig) (Store, http.H
...
 	for {
 		select {
 		case <-sub.Updates:
-			pool.Set(toPeers(c.Members())...)
+			pool.Set(toPeers(c.Members(), stat)...)
 		}
 	}
 }
 
+func updateCacheStats(cache *groupcache.Group, stat stats.StatsReceiver) {
```
dbentley: I'm sorry, I don't understand this and it looks werd... Shouldn't updating it on the stat be enough? Why this middleman function?

PR #220, opened 2017-02-01
gitdb: handle error better

dbentley:
scootapi: fix a watch_job bug

### DIFF
--- a/scootapi/client/watch_job_cmd.go
+++ b/scootapi/client/watch_job_cmd.go
@@ -75,12 +75,17 @@ func PrintJobStatus(jobStatus *scoot.JobStatus) {
 	for taskId, taskStatus := range jobStatus.TaskStatus {
 		fmt.Printf("\tTask %s {\n", taskId)
 		fmt.Printf("\t\tStatus: %s\n", taskStatus.String())
-		runStatus := jobStatus.TaskData[taskId]
-		if runStatus.OutUri != nil {
-			fmt.Printf("\t\tStdout: %v\n", *runStatus.OutUri)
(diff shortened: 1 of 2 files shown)


scootapi/client/watch_job_cmd.go
```diff
@@ -75,12 +75,17 @@ func PrintJobStatus(jobStatus *scoot.JobStatus) {
 	for taskId, taskStatus := range jobStatus.TaskStatus {
 		fmt.Printf("\tTask %s {\n", taskId)
 		fmt.Printf("\t\tStatus: %s\n", taskStatus.String())
-		runStatus := jobStatus.TaskData[taskId]
-		if runStatus.OutUri != nil {
-			fmt.Printf("\t\tStdout: %v\n", *runStatus.OutUri)
-		}
-		if runStatus.ErrUri != nil {
-			fmt.Printf("\t\tStderr: %v\n", *runStatus.ErrUri)
+		runStatus, ok := jobStatus.TaskData[taskId]
```
jasonbs10: Are we fine with task data being null instead of PENDING?

dbentley:
Yes, we should fix the server, *and* we should fix the client to not error, because who knows what happens on the server?

scootapi/client/watch_job_cmd.go
dbentley: Fixed that up.

jasonbs10:
Hmm, trick question? *We* know what's supposed to happen on the server. I'd say this is panic()-worthy but we probably do want to err towards best effort.

PR #221, opened 2017-02-02
snapshot/git: a few fixes

dbentley:
In bundlestore.go, if the we're asked to ingest a comit that's on the branch
in the stream, then return a snap from the stream (instead of trying to
upload)

create.go: when we were adding env variables to our run of git, we were
actually not using any of the existing environment; oops.

### DIFF
--- a/snapshot/git/gitdb/bundlestore.go
+++ b/snapshot/git/gitdb/bundlestore.go
@@ -96,6 +96,10 @@ func (b *bundlestoreBackend) uploadLocalSnapshot(s *localSnapshot, db *DB) (sn s
 
 			mergeBase, err := db.dataRepo.RunSha("merge-base", streamHead, commitSha)
 
+			if mergeBase == commitSha {
+				return &streamSnapshot{sha: commitSha, kind: kindGitCommitSnapshot, streamName: db.stream.cfg.Name}, nil
+			}
+
 			// if err != nil, it just means we don't have a merge-base
 			if err == nil {
 				revList = fmt.Sprintf("%s..%s", mergeBase, bundlestoreTempRef)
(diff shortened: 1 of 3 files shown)


snapshot/git/gitdb/db_test.go
```diff
@@ -337,12 +337,26 @@ func TestBundlestore(t *testing.T) {
 
 	consumerDB := MakeDBFromRepo(consumerDataRepo, fixture.tmp, streamCfg, nil, bundleCfg, AutoUploadBundlestore)
 
+	upstreamMaster, err := fixture.upstream.RunSha("rev-parse", "master")
```
jasonbs10: It's getting onerous to read through longer tests like these. Can you sprinkle in some comments? And in particular, how does this test the bundlestore.go change?

snapshot/git/gitdb/bundlestore.go
```diff
@@ -96,6 +96,14 @@ func (b *bundlestoreBackend) uploadLocalSnapshot(s *localSnapshot, db *DB) (sn s
 
 			mergeBase, err := db.dataRepo.RunSha("merge-base", streamHead, commitSha)
 
+			if mergeBase == commitSha {
```
jasonbs10: When would this happen? I may be misunderstanding, but only if the user committed something to master and then later decided to revisit that revision? And this works because only SQ can merge to master?

snapshot/git/gitdb/db_test.go
dbentley: commented.

This tests by trying to ingest master.

snapshot/git/gitdb/bundlestore.go
dbentley: Commented.

I just tried to test that master worked passed, and it error'ed. Whoops!

PR #222, opened 2017-02-07
sched/scheduler: update status of tasks when they start running

dbentley:
We want to be able to see the status of tasks as they run (so we can, e.g., follow the output as it goes)

This requires using runner.Service directly in task_runner.go

Thus, we can get rid of Worker. Yay.

This change got big because it requires config changes.


sched/scheduler/task_runner_test.go
```diff
@@ -58,7 +99,7 @@ func Test_runTaskAndLog_FailedToLogStartTask(t *testing.T) {
 	sagaCoord := saga.MakeSagaCoordinator(sagaLogMock)
 	s, _ := sagaCoord.MakeSaga("job1", nil)
 
-	err := runTaskAndLog(s, workers.MakeSimWorker(tmp), "task1", task, false, stats.NilStatsReceiver())
+	err := testTaskRunner(s, workers.MakeDoneWorker(tmp), "task1", task, false).run()
```
jasonbs10: Does the type of worker matter here?

sched/scheduler/task_runner.go
```diff
@@ -1,78 +1,150 @@
...
-
-	// TODO(dbentley): Update to use a runner.Service instead of a worker.Worker.
-	// This will allow us to wait with the proper timeout and also wait for
-	// stdoutRef and stderrRef to appear so we can include it in our reply to the user.
-
+func (r *taskRunner) run() error {
+	log.Println("Starting task", r.taskId, " command:", strings.Join(r.task.Argv, " "))
 	// Log StartTask Message to SagaLog
-	err := saga.StartTask(taskId, nil)
+	err := r.saga.StartTask(r.taskId, nil)
```
dbentley: Good idea; done.

sched/scheduler/task_runner.go
```diff
@@ -1,78 +1,150 @@
...
+	endTime := time.Now().Add(cmd.Timeout).Add(r.runnerOverhead)
+
+	st, err := r.runner.Run(&cmd)
+	if err != nil || st.State.IsDone() {
+		return st, err
+	}
+
+	id := st.RunID
+
+	st, err = r.queryWithTimeout(id, endTime, true)
```
dbentley: Done.

sched/scheduler/task_runner.go
```diff
@@ -1,78 +1,150 @@
...
+		st = runner.RunStatus{
+			RunID: id,
+			State: runner.TIMEDOUT,
+		}
+	}
+	return st, nil
+}
+
+func (r *taskRunner) logTaskStatus(st runner.RunStatus, msgType saga.SagaMessageType) error {
+	log.Println("Finishing", st)
```
dbentley: Done.

scootapi/server/job_status.go
```diff
@@ -122,6 +125,10 @@ func workerRunStatusToScootRunStatus(resultsFromSaga []byte) (*scoot.RunStatus,
 		return nil, err
 	}
 
+	log.Println("Ahh", workerRunStatus.SnapshotId)
```
dbentley: Done.

binaries/scheduler/config/local.local
```diff
@@ -12,7 +12,9 @@
     "Type": "stateful",
     "MaxRetriesPerTask" : 2,
     "DebugMode" : false,
-    "RecoverJobsOnStartup" : true
+    "RecoverJobsOnStartup" : true,
+    "DefaultTaskTimeoutMs" : 1800000,
+    "RunnerOverheadMs": 600000
```
dbentley: 10 minutes of overhead accounts for you hitting a worker that's just starting up and needs to initialize.

sched/worker/workers/makers.go
```diff
@@ -5,25 +5,28 @@ import (
...
 	"github.com/scootdev/scoot/runner/execer/execers"
 	"github.com/scootdev/scoot/runner/runners"
-	"github.com/scootdev/scoot/sched/worker"
 	"github.com/scootdev/scoot/snapshot/snapshots"
 )
 
 // Makes a worker suitable for using as an in-memory worker.
-// TODO: these are useful for unit tests but really shoudldn't be used for integration tests...
-func MakeInmemoryWorker(node cluster.Node, tmp *temp.TempDir) worker.Worker {
+func MakeInmemoryWorker(node cluster.Node, tmp *temp.TempDir) runner.Service {
```
dbentley: How would you recommend?

config/scootconfig/workers_modules.go
```diff
@@ -70,13 +61,13 @@ type WorkersLocalConfig struct {
 }
 
 func (c *WorkersLocalConfig) Install(bag *ice.MagicBag) {
-	bag.Put(func(tmp *temp.TempDir) worker.WorkerFactory {
+	bag.Put(func(tmp *temp.TempDir) func(cluster.Node) runner.Service {
 		return InmemoryWorkerFactory(tmp)
```
dbentley: What naming would be better?

runner/execer/execers/done.go
```diff
@@ -3,26 +3,32 @@ package execers
 import "github.com/scootdev/scoot/runner/execer"
 
 // Creates a new doneExecer.
-func NewDoneExecer() execer.Execer {
-	return &doneExecer{}
+func NewDoneExecer() *DoneExecer {
+	return &DoneExecer{
+		State: execer.COMPLETE,
```
dbentley: Because it's needed in stateful_scheduler_test.go, in the test TaskGetsMarkedCompletedAfterMaxRetriesFailedRuns, where we want to make sure that a failed run (on the worker) gets considered an error in the scheduler.

sched/generators.go
```diff
@@ -82,7 +82,7 @@ func GenRandomTask(rng *rand.Rand) TaskDefinition {
 		envVarsMap[fmt.Sprintf("env%d", j)] = testhelpers.GenRandomAlphaNumericString(rng)
 	}
 
-	timeout := time.Duration(rng.Int63n(1000))
+	timeout := time.Duration(rng.Int63n(1000)) * time.Second
```
dbentley: Before, the timeout would always be <1000ns, which is 1 us, which is too short for anything to succeed.

sched/scheduler/stateful_scheduler_test.go
```diff
@@ -215,6 +264,7 @@ func Test_StatefulScheduler_JobRunsToCompletion(t *testing.T) {
 
 	// add additional saga data
 	sagaLogMock.EXPECT().LogMessage(saga.MakeStartTaskMessage(jobId, taskId, nil))
+	sagaLogMock.EXPECT().LogMessage(TaskMessageMatcher{Type: &sagaStartTask, JobId: "job1", TaskId: "task1", Data: gomock.Any()}).AnyTimes()
```
dbentley: It's a race condition whether it sees the process while it's running or if the first poll sees it after it's done, so it's 0 or 1. I've changed it to MaxTimes(1)

sched/scheduler/task_runner_test.go
```diff
@@ -35,16 +49,43 @@ func Test_runTaskAndLog_Successful(t *testing.T) {
 	sagaLogMock := saga.NewMockSagaLog(mockCtrl)
 	sagaLogMock.EXPECT().StartSaga("job1", nil)
 	sagaLogMock.EXPECT().LogMessage(saga.MakeStartTaskMessage("job1", "task1", nil))
+	sagaLogMock.EXPECT().LogMessage(TaskMessageMatcher{Type: &sagaStartTask, JobId: "job1", TaskId: "task1", Data: gomock.Any()}).AnyTimes()
```
dbentley: Yeah, updating the tests took about 3 hours of digging around.

sched/scheduler/task_runner_test.go
dbentley: Yup; with simworker everything fails (because the generated jobs have nonsense commands)

binaries/scheduler/config/local.local
jasonbs10: This seems like a TODO until we can report worker readiness to the scheduler then?

config/scootconfig/workers_modules.go
jasonbs10: Just saying we have half worker and half service terminology. It's not confusing enough to change anything though.

sched/worker/workers/makers.go
jasonbs10: If snapshotId is set then workers should return an appropriate error? I'm not sure if there's a good way to implement my suggestion.

PR #224, opened 2017-03-20
added ability to use log.Fatal(), log.Trace() and log.Info().  This i…

JeanetteBruno:
…s the initial implementation, it writes log files to scoot-tmp-log<xxx>/log.txt.  We can extend it with different loggers and externalize the ability to set the log level

would like initial review of the logging approach.  
- note the changes to task_runner.go (around line 54) were just to make the code more readable.

### DIFF
--- a/binaries/scheduler/config/config.go
+++ b/binaries/scheduler/config/config.go
@@ -85,7 +85,7 @@ func configConfigGo() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "config/config.go", size: 0, mode: os.FileMode(420), modTime: time.Unix(1486507161, 0)}
+	info := bindataFileInfo{name: "config/config.go", size: 0, mode: os.FileMode(420), modTime: time.Unix(1490022286, 0)}
 	a := &asset{bytes: bytes, info: info}
 	return a, nil
 }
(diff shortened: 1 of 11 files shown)


dgassaway:
I'm a little confused on the usage for the different levels.  I would assume all of the logs that are being changed to trace here are info logs that we would want recorded on a normal production system.  Also wondering why there's no error level - assuming that it's possible we will encounter recoverable errors and will want to be able to ID those by level.

dgassaway:
Also confused that the levels seem not to be (lowest -> highest): trace -> info -> fatal.

On the file, I'd assume logs would be output to ./<svc-name>.log.

It seems like we are doing a lot of wheel-inventing here by writing a small logging lib vs using an open source one so I'm really not following this.

JeanetteBruno:
The logging levels are documented at: https://confluence.twitter.biz/display/EE/Development+Practices.  
- Fatal - 'something very bad has happened that we cannot recover from'
- Trace - 'I want to be able to follow the processing flow'
- Info  - 'I want to print minor details to help me debug some new code and I don't want to have to remember to go back and remove print statements'

We can reverse the order of the levels
We can write to ???.log, but  I'm not a big fan of using the type extension, till we support a configurable file name what would we call it? log.log?

Yes - we are creating our own little utiility...I thought we said we didn't want to use logrus... golang's log only offers Fatal and Print.

jasonbs10:
My understanding was that no one objected to deferring this work for the time being (I moved the corresponding ticket to tech debt early this morning). I agree with Drew that we want to limit custom code where public options may suffice. I'd suggest that if we want to make exceptions to that rule of thumb we need to hash out a quick design doc and create an interface PR first.

JeanetteBruno:
abandoning see dx1991/logContent branch

PR #225, opened 2017-03-20
Poll indefinitely if there is no timeout

ryancouto:
Tasks have been timing out without a timeout being set. This is because the timeout default value is 0. A timeout of 0 should be read as absence of a timeout, not as a 0 second timeout. If there is no timeout, then we should poll indefinitely.

### DIFF
--- a/runner/runners/polling.go
+++ b/runner/runners/polling.go
@@ -34,18 +34,14 @@ func (r *PollingStatusQuerier) QueryNow(q runner.Query) ([]runner.RunStatus, err
 // Query returns all RunStatus'es matching q, waiting as described by w
 func (r *PollingStatusQuerier) Query(q runner.Query, wait runner.Wait) ([]runner.RunStatus, error) {
 	end := time.Now().Add(wait.Timeout)
-	for {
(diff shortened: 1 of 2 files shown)


runner/runners/polling.go
```diff
@@ -33,19 +33,19 @@ func (r *PollingStatusQuerier) QueryNow(q runner.Query) ([]runner.RunStatus, err
...
 func (r *PollingStatusQuerier) Query(q runner.Query, wait runner.Wait) ([]runner.RunStatus, error) {
-	if wait.Timeout == 0 {
-		return r.del.QueryNow(q)
-	}
 	end := time.Now().Add(wait.Timeout)
-	for time.Now().Before(end) {
-		st, err := r.del.QueryNow(q)
-		if err != nil || len(st) > 0 {
-			return st, err
+	for {
```
dgassaway: Do we need the switch? Can we just use the switch compound conditional for the for loop (and leave the return nil/nil outside of it)?

ryancouto: You're right - brain fart.

PR #226, opened 2017-03-20
updated log statements

JeanetteBruno:
[schedulerLog.txt](https://github.com/scootdev/scoot/files/856578/schedulerLog.txt)
 the attached file contains the logging output from scheduler with this change (plus commented out log statement in fetcher.go in twitter source)

### DIFF
--- a/runner/runners/invoke.go
+++ b/runner/runners/invoke.go
@@ -57,7 +57,7 @@ type checkoutAndError struct {
 // Run will enforce cmd's Timeout, and will abort cmd if abortCh is signaled.
 // Run will not return until the process is not running.
 func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struct{}, updateCh chan runner.RunStatus) (r runner.RunStatus) {
-	log.Printf("run. id:%v, cmd: %+v", id, cmd)
+	log.Printf("run. id: %v, cmd: %+v", id, cmd)
 	defer func() {
 		updateCh <- r
 		close(updateCh)
@@ -70,7 +70,7 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 		if cmd.SnapshotID == "" {
 			//TODO: we don't want this logic to live here, these decisisions should be made at a higher level.
 			if len(cmd.Argv) > 0 && cmd.Argv[0] != execers.UseSimExecerArg {
-				log.Printf("RunID=%s has no snapshotID! Using a nop-checkout initialized with cwd.\n", id)
+				log.Printf("RunID:%s has no snapshotID! Using a nop-checkout initialized with cwd.\n", id)
 			}
 			checkout := gitfiler.MakeUnmanagedCheckout("", "./")
 			checkoutCh <- checkoutAndError{checkout, nil}
@@ -98,7 +98,7 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 
 	defer checkout.Release()
 
-	log.Printf("runner/runners/invoke.go: checkout done. id %v cmd: %+v checkout: %v", id, cmd, checkout.Path())
+	log.Printf("checkout done. id: %v, cmd: %+v, checkout: %v", id, cmd, checkout.Path())
 
 	stdout, err := inv.output.Create(fmt.Sprintf("%s-stdout", id))
 	if err != nil {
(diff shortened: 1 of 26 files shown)


snapshot/git/gitdb/bundlestore.go
```diff
@@ -207,15 +209,15 @@ func (s *bundlestoreSnapshot) Download(db *DB) error {
 	// unbundle optimistically
 	// this will succeed if we have all of the prerequisite objects
 
-	log.Print("Return if unbundling gets us our sha, else continue. ", s.SHA())
 	if _, err = db.dataRepo.Run("bundle", "unbundle", filename); err == nil {
+		log.Print("Unbundling got the sha:%s, returning from Download()", s.SHA())
```
ryancouto: log.Printf

runner/runners/invoke.go
```diff
@@ -57,7 +57,7 @@ type checkoutAndError struct {
 // Run will enforce cmd's Timeout, and will abort cmd if abortCh is signaled.
 // Run will not return until the process is not running.
 func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struct{}, updateCh chan runner.RunStatus) (r runner.RunStatus) {
-	log.Printf("runner/runners/invoke.go: run. id %v cmd %+v", id, cmd)
+	log.Printf("run. id:%v, cmd: %+v", id, cmd)
```
dgassaway: total nit - space after the ':' in this line and many of the other new/updated lines

snapshot/git/gitdb/bundlestore.go
```diff
@@ -191,11 +191,13 @@ func (s *bundlestoreSnapshot) Kind() snapshotKind { return s.kind }
 func (s *bundlestoreSnapshot) SHA() string        { return s.sha }
 
 func (s *bundlestoreSnapshot) Download(db *DB) error {
-	log.Print("Return if we already have this sha, else continue by downloading from bundlestore. ", s.SHA())
+	log.Printf("Downloading sha:%s", s.SHA())
```
jasonbs10: While these changes look clearer, they may not entirely address the original intent of these prints - we want to be very explicit that if an error is printed by the functions in some of these steps, that's ok/expected and the viewer can safely ignore the error.

jasonbs10:
My previous comment is hidden as outdated in case you miss it. I think the biggest remaining thing is making sure that we always print the task result, probably in runner/runners/invoke.go so we always know what happened, and possibly also in workerapi/server/server.go  so we know what got sent/received. I don't know how frequently Query() gets called though, maybe too often to print.

jasonbs10:
Oh, also, can you instrument queue.go:Run() etc to help us debug seemingly spurious QueueFullMsg's?

JeanetteBruno:
[remoteSmokeTestLog.txt](https://github.com/scootdev/scoot/files/859327/remoteSmokeTestLog.txt)
[localSmokeTestRun.txt](https://github.com/scootdev/scoot/files/859332/localSmokeTestRun.txt)

runner/runners/queue.go
```diff
@@ -24,7 +26,11 @@ func NewQueueRunner(
 	statusManager := NewStatusManager()
 	inv := NewInvoker(exec, filer, output, tmp)
 	controller := &QueueController{statusManager: statusManager, inv: inv, capacity: capacity}
-	return &Service{controller, statusManager, statusManager}
+	runner := &Service{controller, statusManager, statusManager}
+	log.Println("************** runner definition")
```
jasonbs10: I'd try just one of these blocks located in ice/eval.go which prints the values map. If that works we can replace the other blocks that are sprinkled throughout.

runner/runners/polling.go
```diff
@@ -3,21 +3,31 @@ package runners
...
+	log.Println(runnerDesc)
+	return runner
 }
 
 // NewPollingService creates a new Service from a Controller, a StatusEraser, and a StatusQueryNower.
 // (This is a convenience function over NewPollingStatusQuerier
 func NewPollingService(c runner.Controller, e runner.StatusEraser, nower runner.StatusQueryNower, period time.Duration) runner.Service {
 	q := NewPollingStatusQuerier(nower, period)
-	return &Service{c, q, e}
+	runner := &Service{c, q, e}
```
jasonbs10: We don't need this one since it calls NewPollingStatusQuerier.

I don't think it's worth the verbosity to print this every time the function is called, especially if we can do it once for every type on startup after parsing and evaluating the config. More to the point, it looks like the repeated prints may currently be devoid of actionable info.

JeanetteBruno:
[localRunSpewingIce.txt](https://github.com/scootdev/scoot/files/859436/localRunSpewingIce.txt)

JeanetteBruno:
[localRunLog.txt](https://github.com/scootdev/scoot/files/859666/localRunLog.txt)

PR #229, opened 2017-03-27
Jschiller/reuse workspace

jasonbs10:
Still needs tests but otherwise ready for review I believe.

### DIFF
--- a/runner/runners/invoke.go
+++ b/runner/runners/invoke.go
@@ -47,15 +47,11 @@ func (inv *Invoker) Run(cmd *runner.Command, id runner.RunID) (abortCh chan<- st
 	return abortChFull, updateChFull
 }
 
-type checkoutAndError struct {
-	checkout snapshot.Checkout
-	err      error
-}
-
 // Run runs cmd as run id returning the final ProcessStatus
 // Run will send updates the process is running to updateCh.
 // Run will enforce cmd's Timeout, and will abort cmd if abortCh is signaled.
 // Run will not return until the process is not running.
+// NOTE: kind of gnarly since our filer implementation (gitdb) currently mutates the same worktree on every op.
 func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struct{}, updateCh chan runner.RunStatus) (r runner.RunStatus) {
 	log.Printf("run. id: %v, cmd: %+v", id, cmd)
 	defer func() {
@@ -63,42 +59,48 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 		close(updateCh)
 	}()
 
-	checkoutCh := make(chan checkoutAndError)
-	var checkout snapshot.Checkout
-	var err error
+	var co snapshot.Checkout
+	checkoutCh := make(chan error)
 	go func() {
 		if cmd.SnapshotID == "" {
-			//TODO: we don't want this logic to live here, these decisisions should be made at a higher level.
+			//TODO: we don't want this logic to live here, these decisions should be made at a higher level.
 			if len(cmd.Argv) > 0 && cmd.Argv[0] != execers.UseSimExecerArg {
-				log.Printf("RunID:%s has no snapshotID! Using a nop-checkout initialized with cwd.\n", id)
+				log.Printf("RunID:%s has no snapshotID! Using a nop-checkout initialized with tmpDir.\n", id)
 			}
-			checkout := gitfiler.MakeUnmanagedCheckout("", "./")
-			checkoutCh <- checkoutAndError{checkout, nil}
+			if tmp, err := inv.tmp.TempDir("invoke_nop_checkout"); err != nil {
+				checkoutCh <- err
+			} else {
+				co = gitfiler.MakeUnmanagedCheckout(string(id), tmp.Dir)
+				checkoutCh <- nil
+			}
 		} else {
-			checkout, err = inv.filer.Checkout(cmd.SnapshotID)
-			checkoutCh <- checkoutAndError{checkout, err}
+			//NOTE: given the current gitdb impl, this checkout will block until the previous checkout is released.
+			log.Printf("RunID:%s checking out snapshotID:%s", id, cmd.SnapshotID)
+			var err error
+			co, err = inv.filer.Checkout(cmd.SnapshotID)
+			checkoutCh <- err
 		}
 	}()
 
 	select {
 	case <-abortCh:
 		go func() {
-			checkoutAndErr := <-checkoutCh
-			if checkoutAndErr.err != nil {
+			if err := <-checkoutCh; err != nil {
+				// If there was an error there should be no lingering gitdb locks, so return.
 				return
 			}
-			checkoutAndErr.checkout.Release()
+			// If there was no error then we need to release this checkout.
+			co.Release()
 		}()
 		return runner.AbortStatus(id)
-	case checkoutAndErr := <-checkoutCh:
-		if checkout, err = checkoutAndErr.checkout, checkoutAndErr.err; err != nil {
+	case err := <-checkoutCh:
+		if err != nil {
 			return runner.ErrorStatus(id, err)
 		}
+		// Checkout is ok, continue with run and when finished release checkout.
+		defer co.Release()
 	}
-
-	defer checkout.Release()
-
-	log.Printf("checkout done. id: %v, cmd: %+v, checkout: %v", id, cmd, checkout.Path())
+	log.Printf("checkout done. id: %v, cmd: %+v, checkout: %v", id, cmd, co.Path())
 
 	stdout, err := inv.output.Create(fmt.Sprintf("%s-stdout", id))
 	if err != nil {
@@ -114,7 +116,7 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 
 	p, err := inv.exec.Exec(execer.Command{
 		Argv:   cmd.Argv,
-		Dir:    checkout.Path(),
+		Dir:    co.Path(),
 		Stdout: stdout,
 		Stderr: stderr,
 	})
@@ -146,7 +148,7 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 	case st = <-processCh:
 	}
 
-	log.Printf("run done. id: %v, status: %+v, cmd: %+v, checkout: %v", id, st, cmd, checkout.Path())
+	log.Printf("run done. id: %v, status: %+v, cmd: %+v, checkout: %v", id, st, cmd, co.Path())
 
 	switch st.State {
 	case execer.COMPLETE:
--- a/sched/scheduler/cluster_state.go
+++ b/sched/scheduler/cluster_state.go
@@ -1,8 +1,9 @@
 package scheduler
 
 import (
-	"github.com/scootdev/scoot/cloud/cluster"
 	"log"
+
+	"github.com/scootdev/scoot/cloud/cluster"
 )
 
 const noTask = ""
@@ -12,12 +13,23 @@ const noTask = ""
 type clusterState struct {
 	updateCh chan []cluster.NodeUpdate
 	nodes    map[cluster.NodeId]*nodeState
+	affinity map[string]*nodeSet
+}
+
+type nodeSet struct {
+	idle map[cluster.NodeId]cluster.Node
+	busy map[cluster.NodeId]cluster.Node
+}
+
+func newNodeSet() *nodeSet {
+	return &nodeSet{idle: map[cluster.NodeId]cluster.Node{}, busy: map[cluster.NodeId]cluster.Node{}}
 }
 
 // The State of A Node in the Cluster
 type nodeState struct {
 	node        cluster.Node
 	runningTask string
+	snapshotID  string
 }
 
 // Initializes a Node State for the specified Node
@@ -25,30 +37,40 @@ func newNodeState(node cluster.Node) *nodeState {
 	return &nodeState{
 		node:        node,
 		runningTask: noTask,
+		snapshotID:  "",
 	}
 }
 
 // Creates a New State Distributor with the initial nodes, and which updates
 // nodes added or removed based on the supplied channel.
 func newClusterState(initial []cluster.Node, updateCh chan []cluster.NodeUpdate) *clusterState {
-
 	nodes := make(map[cluster.NodeId]*nodeState)
+	affinity := map[string]*nodeSet{"": newNodeSet()}
 	for _, n := range initial {
 		nodes[n.Id()] = newNodeState(n)
+		affinity[""].idle[n.Id()] = n
 	}
 
-	cs := &clusterState{
+	return &clusterState{
 		updateCh: updateCh,
 		nodes:    nodes,
+		affinity: affinity,
 	}
-
-	return cs
 }
 
 // Update ClusterState to reflect that a task has been scheduled on a
 // particular node
-func (c *clusterState) taskScheduled(nodeId cluster.NodeId, taskId string) {
+func (c *clusterState) taskScheduled(nodeId cluster.NodeId, taskId string, snapshotID string) {
 	ns := c.nodes[nodeId]
+
+	delete(c.affinity[ns.snapshotID].idle, nodeId)
+	delete(c.affinity[ns.snapshotID].busy, nodeId)
+	if _, ok := c.affinity[snapshotID]; !ok {
+		c.affinity[snapshotID] = newNodeSet()
+	}
+	c.affinity[snapshotID].idle[nodeId] = ns.node
+
+	ns.snapshotID = snapshotID
 	ns.runningTask = taskId
 }
 
@@ -88,11 +110,16 @@ func (c *clusterState) update(updates []cluster.NodeUpdate) {
 			// add the node if it doesn't already exist
 			if _, ok := c.nodes[update.Node.Id()]; !ok {
 				c.nodes[update.Node.Id()] = newNodeState(update.Node)
+				c.affinity[""].idle[update.Node.Id()] = update.Node
 				log.Printf("Added node: %+v, now have %d nodes\n", update.Node, len(c.nodes))
 			}
 		case cluster.NodeRemoved:
-			delete(c.nodes, update.Id)
-			log.Printf("Removed nodeId: %s, now have %d nodes\n", string(update.Id), len(c.nodes))
+			log.Printf("Removed nodeId: %s (%v), now have %d nodes\n", string(update.Id), c.nodes[update.Id], len(c.nodes))
+			if nodeState, ok := c.nodes[update.Id]; ok {
+				delete(c.affinity[nodeState.snapshotID].idle, update.Id)
+				delete(c.affinity[nodeState.snapshotID].busy, update.Id)
+				delete(c.nodes, update.Id)
+			}
 		}
 	}
 
--- a/sched/scheduler/cluster_state.go
+++ b/sched/scheduler/cluster_state.go
@@ -13,7 +13,7 @@ const noTask = ""
 type clusterState struct {
 	updateCh   chan []cluster.NodeUpdate
 	nodes      map[cluster.NodeId]*nodeState
-	nodeGroups map[string]*nodeGroup //key is a snapshotID.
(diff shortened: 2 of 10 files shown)


sched/scheduler/cluster_state.go
```diff
@@ -12,43 +13,64 @@ const noTask = ""
...
+	delete(c.affinity[ns.snapshotID].idle, nodeId)
+	delete(c.affinity[ns.snapshotID].busy, nodeId)
+	if _, ok := c.affinity[snapshotID]; !ok {
+		c.affinity[snapshotID] = newNodeSet()
+	}
+	c.affinity[snapshotID].idle[nodeId] = ns.node
+
+	ns.snapshotID = snapshotID
 	ns.runningTask = taskId
 }
```
JeanetteBruno: when the task is complete, should we be moving the node to the snapshot's idle map?  in taskCompleted() ...
c.affinity[snapshotId].idle[nodeId] = ns.node
delete(c.affinity[snapshotId].busy, nodeId]

sched/scheduler/cluster_state_test.go
```diff
@@ -53,7 +54,7 @@ func Test_ClusterState_DuplicateNodeAdd(t *testing.T) {
 	cl := makeTestCluster("node1")
```
JeanetteBruno: shouldn't we have some tests with non-nil snapshotID?

sched/scheduler/task_scheduler.go
```diff
@@ -9,27 +9,50 @@ type taskAssignments struct {
...
-
-	// for now just assign any task to any available node
+// Returns a list of taskAssigments of task to available node.
+// Also returns a modified copy of clusterState.Affinity for the caller to apply (so this remains a pure fn).
+// Does best effort scheduling which tries to assign tasks to nodes already primed for similar tasks.
+// Not all tasks are guaranteed to be scheduled.
+func getTaskAssignments(cs *clusterState, tasks []*taskState) ([]taskAssignments, map[string]*nodeSet) {
 	var tas []taskAssignments
-	taskIndex := 0
+	affinity := map[string]*nodeSet{}
```
JeanetteBruno: nodeGroups

sched/scheduler/task_scheduler.go
```diff
@@ -9,27 +9,50 @@ type taskAssignments struct {
...
-				tas = append(tas, taskAssignments{
-					node: nodeState.node,
-					task: tasks[taskIndex],
-				})
-				taskIndex++
-			} else {
-				// we've scheduled all tasks
-				break
+	// Loop over all snapshotIDs looking for an idle node. Prefer, in order:
+	// - Hot node for the given snapshotID
```
JeanetteBruno: not sure what 'Hot' means... idle? (same comment for line 36)

sched/scheduler/stateful_scheduler.go
```diff
@@ -270,7 +265,8 @@ func (s *statefulScheduler) scheduleTasks() {
 	}
 
 	// Calculate a list of Tasks to Node Assignments & start running all those jobs
-	taskAssignments := getTaskAssignments(s.clusterState, unscheduledTasks)
+	taskAssignments, affinity := getTaskAssignments(s.clusterState, unscheduledTasks)
```
JeanetteBruno: if we make getTaskAssignments into getTaskAssignment and move into the scheduling loop below (make that loop, loop over unscheduledTasks), I think we can eliminate the logic for copying the nodeSet in getTaskAssignment.  getTaskAssignment can just look for an idle node with that snapshotId, otherwise and idle node with snapshotID "", otherwise, search through all of clusterState's nodesets for the first Idle node.

sched/scheduler/cluster_state.go
```diff
@@ -12,43 +13,64 @@ const noTask = ""
...
+		affinity: affinity,
 	}
-
-	return cs
 }
 
 // Update ClusterState to reflect that a task has been scheduled on a
 // particular node
-func (c *clusterState) taskScheduled(nodeId cluster.NodeId, taskId string) {
+func (c *clusterState) taskScheduled(nodeId cluster.NodeId, taskId string, snapshotID string) {
```
JeanetteBruno: add to comment that snapshotId must be the snapshotID from the task's definition.

scootapi/client/run_job_cmd.go
```diff
@@ -22,8 +25,9 @@ func (c *runJobCmd) registerFlags() *cobra.Command {
 		Use:   "run_job",
 		Short: "run a job",
 	}
-	r.Flags().StringVar(&c.snapshotId, "snapshot_id", scoot.TaskDefinition_SnapshotId_DEFAULT, "snapshot ID to run job against")
-	r.Flags().StringVar(&c.jobFilePath, "job_def", "", "JSON file to read jobs from")
+	r.Flags().StringVar(&c.snapshotId, "stream_name", "sm", "If passing snapshot_id=SHA, this is the global UID for the associated repo.")
```
JeanetteBruno: we need to be outlining/starting to create user docs for this feature.

snapshot/git/gitdb/checkout.go
```diff
@@ -51,7 +52,17 @@ func (db *DB) checkout(id snap.ID) (path string, err error) {
 		return db.checkoutFSSnapshot(v.SHA())
 	case kindGitCommitSnapshot:
 		// For GitCommitSnapshot's, we use dataRepo's work tree.
-		return db.checkoutGitCommitSnapshot(v.SHA())
+		if id == db.currentSnapID {
```
JeanetteBruno: why doesn't this logic apply to FSSnapshot? (above)

scootapi/client/run_job_cmd.go
ryancouto: should this be &c.stream_name?

sched/scheduler/cluster_state.go
```diff
@@ -12,43 +13,64 @@ const noTask = ""
 type clusterState struct {
 	updateCh chan []cluster.NodeUpdate
 	nodes    map[cluster.NodeId]*nodeState
+	affinity map[string]*nodeSet
```
ryancouto: is the string key a snapshot id? if so can you add a comment?

sched/scheduler/task_scheduler.go
```diff
@@ -9,27 +9,50 @@ type taskAssignments struct {
...
 	var tas []taskAssignments
-	taskIndex := 0
+	affinity := map[string]*nodeSet{}
+	snapshotIds := []string{}
+
+	// Create a copy of affinity to modify based on new scheduling.
+	for snapId, set := range cs.affinity {
+		affinity[snapId] = newNodeSet()
+		for nodeId, node := range set.idle {
+			affinity[snapId].idle[nodeId] = node
```
ryancouto: is there a race condition between this and

sched/scheduler/cluster_state.go#66
`delete(c.affinity[nodeState.snapshotID].idle, update.Id)`

sched/scheduler/task_scheduler.go
```diff
@@ -9,27 +9,50 @@ type taskAssignments struct {
...
+	// - New untouched node (or node run with an empty snapshotID)
+	// - Any node that's hot for a different snapshotID.
+	// Note: tasks are expected to have a valid snapshotID but if it's empty this should still work.
+Loop:
+	for _, task := range tasks {
+		snapIds := append([]string{task.Def.SnapshotID, ""}, snapshotIds...)
+		for _, snapId := range snapIds {
+			if set, ok := affinity[snapId]; ok {
+				for nodeId, node := range set.idle {
+					tas = append(tas, taskAssignments{node: node, task: task})
```
ryancouto: not a huge fan of tas as a var name, confusing with tasks and task right there. taskAssigns?

sched/scheduler/task_scheduler.go
```diff
@@ -9,27 +9,50 @@ type taskAssignments struct {
...
-				taskIndex++
-			} else {
-				// we've scheduled all tasks
-				break
+	// Loop over all snapshotIDs looking for an idle node. Prefer, in order:
+	// - Hot node for the given snapshotID
+	// - New untouched node (or node run with an empty snapshotID)
+	// - Any node that's hot for a different snapshotID.
+	// Note: tasks are expected to have a valid snapshotID but if it's empty this should still work.
+Loop:
```
dgassaway: this seems weird

runner/runners/invoke.go
```diff
@@ -47,58 +47,60 @@ func (inv *Invoker) Run(cmd *runner.Command, id runner.RunID) (abortCh chan<- st
...
 
-	checkoutCh := make(chan checkoutAndError)
-	var checkout snapshot.Checkout
-	var err error
+	var co snapshot.Checkout
+	checkoutCh := make(chan error)
 	go func() {
 		if cmd.SnapshotID == "" {
-			//TODO: we don't want this logic to live here, these decisisions should be made at a higher level.
+			//TODO: we don't want this logic to live here, these decisions should be made at a higher level.
```
jasonbs10: Let's decide if we want to allow empty snapshotID in scootapi before deleting this (I'm leaning towards requiring them despite slightly impairing ease of use).

sched/scheduler/cluster_state.go
jasonbs10: Done.

sched/scheduler/cluster_state.go
```diff
@@ -12,43 +13,64 @@ const noTask = ""
...
+		snapshotID:  "",
 	}
 }
 
 // Creates a New State Distributor with the initial nodes, and which updates
 // nodes added or removed based on the supplied channel.
 func newClusterState(initial []cluster.Node, updateCh chan []cluster.NodeUpdate) *clusterState {
-
 	nodes := make(map[cluster.NodeId]*nodeState)
+	affinity := map[string]*nodeSet{"": newNodeSet()}
```
jasonbs10: Done.

sched/scheduler/cluster_state.go
jasonbs10: Done.

sched/scheduler/cluster_state.go
```diff
@@ -12,43 +13,64 @@ const noTask = ""
...
 }
 
 // Update ClusterState to reflect that a task has been scheduled on a
 // particular node
-func (c *clusterState) taskScheduled(nodeId cluster.NodeId, taskId string) {
+func (c *clusterState) taskScheduled(nodeId cluster.NodeId, taskId string, snapshotID string) {
 	ns := c.nodes[nodeId]
+
+	delete(c.affinity[ns.snapshotID].idle, nodeId)
+	delete(c.affinity[ns.snapshotID].busy, nodeId)
```
jasonbs10: It's not an error but I believe it's redundant wrt the delete in in clusterState.update(). The underlying issue is that I've been seeing nodes re-added to our cluster with a new shardID and then a lagging delete of the old shardID.

sched/scheduler/cluster_state.go
jasonbs10: Thanks, I forgot to fill that in.

sched/scheduler/cluster_state_test.go
jasonbs10: Yep, TODO per my commit comment. I won't merge without them.

sched/scheduler/task_scheduler.go
jasonbs10: Done.

sched/scheduler/task_scheduler.go
jasonbs10: Commented.

sched/scheduler/stateful_scheduler.go
jasonbs10: We can revisit this later. The idea here is that task_scheduler needs to know all the tasks at once to do basic scheduling (or it needs to take additional params for # of tasks remaining with the same jobID, # of jobs, # of remaining tasks with different jobID, etc).

scootapi/client/run_job_cmd.go
jasonbs10: I added this for dx-team ease-of-use, so it's more likely a candidate for internal docs. When we move beyond CI+sickle as our only customer we'll want to do more user facing docs.

jasonbs10: Doh, good catch!

sched/scheduler/task_scheduler.go
jasonbs10: There shouldn't be any race conditions since we do everything sequentially; there's a single goroutine which loops doing statefulScheduler.step().

sched/scheduler/task_scheduler.go
jasonbs10: Done, renamed to 'assignments'.

sched/scheduler/cluster_state.go
```diff
@@ -12,43 +13,64 @@ const noTask = ""
...
+	if _, ok := c.affinity[snapshotID]; !ok {
+		c.affinity[snapshotID] = newNodeSet()
+	}
+	c.affinity[snapshotID].idle[nodeId] = ns.node
```
jasonbs10: Yep, thanks.

sched/scheduler/task_scheduler.go
jasonbs10: I made minor edits to the comments. Can you elaborate on how I can make this clearer?

snapshot/git/gitdb/checkout.go
jasonbs10: We aren't using git_tags.go - it was created as a backup in case bundlestore wasn't working out. I'd like to delete it after we hit MVP.

We don't want to clean up the last run since we're trying to preserve the build cache for tasks with the same snapshotID.

sched/scheduler/cluster_state.go
```diff
@@ -13,16 +13,16 @@ const noTask = ""
 type clusterState struct {
 	updateCh chan []cluster.NodeUpdate
 	nodes    map[cluster.NodeId]*nodeState
-	affinity map[string]*nodeSet
+	affinity map[string]*nodeGroups //key is a snapshotID.
```
JeanetteBruno: Sorry, I wasn't clear.... I was recommending changing affinity to nodeGroups.
the nodeGroups type should be singular (nodeGroup)... 

type clusterState struct {
...
    nodeGroups map[string]*nodeGroup // key is snapshotID.
}

type nodeGroup struct {
...
}

sched/scheduler/task_scheduler.go
```diff
@@ -13,37 +13,36 @@ type taskAssignments struct {
...
 
 	// Loop over all snapshotIDs looking for an idle node. Prefer, in order:
-	// - Hot node for the given snapshotID
-	// - New untouched node (or node run with an empty snapshotID)
-	// - Any node that's hot for a different snapshotID.
-	// Note: tasks are expected to have a valid snapshotID but if it's empty this should still work.
+	// - Hot node for the given snapshotID (one whose last task shared the same snapshotID).
+	// - New untouched node (or node whose last task used an empty snapshotID)
+	// - Any node from the idle pools of nodes associated with other snapshotIDs.
 Loop:
```
JeanetteBruno: I would leave labels out unless they are used (they are too closely associated with the goto hell I lived through way back when!)

sched/scheduler/cluster_state.go
jasonbs10: Done.

sched/scheduler/task_scheduler.go
jasonbs10: This is kosher here. People have a knee-jerk reaction to goto's and the like but they have a proper place in clarifying intent for deeply nested loops.

runner/runners/invoke.go
JeanetteBruno: I agree we should require snapshotId.   The execers.UseSimExecerArg test brings special test behavior into the code - This is an anti-pattern. We should update the tests to have a noop checkouter when SimExecer is used and provide a dummy snapshotId.  (if this can't be done now, we should create a techdebt jira issue and put the issue number in the comment).

Also will we have a task with no cmd.Argv?

jasonbs10: I'll create a ticket. I originally thought we could/should reference our internal tickets here but Dan disagreed. We should poll the group again to decide.

I can't think of a reasonable case for an empty Argv off the top of my head other than an escape hatch for sending env vars to workers to communicate out-of-band logic.

sched/scheduler/task_scheduler.go
```diff
@@ -9,27 +9,49 @@ type taskAssignments struct {
...
-				taskIndex++
-			} else {
-				// we've scheduled all tasks
-				break
+	// Loop over all snapshotIDs looking for an idle node. Prefer, in order:
+	// - Hot node for the given snapshotID (one whose last task shared the same snapshotID).
+	// - New untouched node (or node whose last task used an empty snapshotID)
+	// - Any node from the idle pools of nodes associated with other snapshotIDs.
+Loop:
+	for _, task := range tasks {
```
JeanetteBruno: I this loop behaved as follows we could eliminate the copy above and the function only has to return the taskAssignments.

for each task
     if cs.affinity[task.snapshotID].idle is not empty then assign the task to the first idle node and continue to the next task
     if cs.affinity[""].idle is not empty then assign the task to the first idle node and continue to the next task
     for each key in cs.affinity 
            if key == "" or key == task.snapshotId continue to the next key
            if cs.affinity[key].idle is not empty then assign the task to the first idle node and break out of the key loop, continuing on to the next task assignment

sched/scheduler/task_scheduler.go
```diff
@@ -9,27 +9,49 @@ type taskAssignments struct {
...
-				})
-				taskIndex++
-			} else {
-				// we've scheduled all tasks
-				break
+	// Loop over all snapshotIDs looking for an idle node. Prefer, in order:
+	// - Hot node for the given snapshotID (one whose last task shared the same snapshotID).
+	// - New untouched node (or node whose last task used an empty snapshotID)
+	// - Any node from the idle pools of nodes associated with other snapshotIDs.
+Loop:
```
JeanetteBruno: can eliminate the label with logic change below (I also am wary of labels, they are too closely associated with the goto hell I lived in way back when!)

sched/scheduler/task_scheduler.go
jasonbs10: I'm probably going to punt on this unless the following applies:

I suggest we set our criteria for code quality such that we're not spending time on what could be ascribed to preference. We should accept code where:
1) Intent is clear
2) Behavior isn't suspicious wrt unintended side effects

We've spent a considerable # of dev hours in the past doing ultimately unnecessary things like runner refactor and going forward I'd like to avoid allocating any time that way, even for something as small as this section of code.

What does everyone else think?

sched/scheduler/task_scheduler.go
```diff
@@ -9,27 +9,49 @@ type taskAssignments struct {
...
+// Not all tasks are guaranteed to be scheduled.
+func getTaskAssignments(cs *clusterState, tasks []*taskState) ([]taskAssignments, map[string]*nodeGroup) {
+	var assignments []taskAssignments
+	nodeGroups := map[string]*nodeGroup{}
+	snapshotIds := []string{}
 
-	// for now just assign any task to any available node
-	var tas []taskAssignments
-	taskIndex := 0
+	// Create a copy of nodeGroups to modify based on new scheduling.
```
JeanetteBruno: we can eliminate this logic - see the comment below

sched/scheduler/task_scheduler.go
jasonbs10: (Offline conversation revealed that this change reduces node thrashing so I'll go ahead and make the change).

ryancouto: Agreed on the 3 criteria. I do understand the aversion to labels, and I feel similarly, but I wouldn't want to explicitly rule them out. They can be useful in some cases.

runner/runners/invoke.go
ryancouto: I'm fine with referencing internal tickets as long as we're the only people working on this.

sched/scheduler/task_scheduler.go
dgassaway: I'd prefer we don't use labels but I don't think refactoring this loop is worth the effort right now.  To the above heuristics we should add:
4. Adhere to language idioms and best practices

jasonbs10: I'll make the change since it's a fairly easy way to avoid costly node thrashing.

PR #230, opened 2017-03-30
rcouto/log15

ryancouto:
This replaces the stdlib logging we have used up to this point with a wrapper for github.com/inconshreveable/log15 as described within their godoc (https://godoc.org/github.com/inconshreveable/log15#hdr-Library_Use).

log15 has 5 levels we can use, Debug, Info, Warn, Error, Crit. 

I'm of the opinion that we only really need Crit for fatal errors, Info, for user facing log statements, and Debug for logs that we, as developers, care about, but I'm open to extending our API if we feel that we would benefit from including the Warn and Error levels (https://dave.cheney.net/2015/11/05/lets-talk-about-logging).

I replaced all log.Print(f/ln) statements with log.Info, this is in all likelihood not an accurate representation of the appropriate level for all of the logging statements, but that will require a much more thorough pass through at a later point.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -57,6 +57,10 @@
 	path = vendor/github.com/scootdev/groupcache
 	url = https://github.com/scootdev/groupcache
 	branch = 85fde1c5c36988481c30fc26605cd83f21cc94d1
+[submodule "vendor/github.com/scootdev/groupcache"]
+	path = vendor/github.com/inconshreveable/log15
+	url = https://github.com/inconshreveable/log15
+	branch = 39bacc234bf1afd0b68573e95b45871f67ba2cd4
 
 #
 # Everything after here is a transitive dependency that is not required to use scoot.
--- a/.gitmodules
+++ b/.gitmodules
@@ -203,3 +203,11 @@
 	path = vendor/github.com/pmezard/go-difflib
 	url = https://github.com/pmezard/go-difflib.git
 	branch = 792786c7400a136282c1664665ae0a8db921c6c2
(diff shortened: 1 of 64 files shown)


ryancouto:
Question for whoever reviews this - should all remaining fmt package calls be changed to log?

jasonbs10:
Yes, fmt calls (excluding Sprintf family) should be converted imo.

jasonbs10:
I think that blog piece has the right intentions but assumes on-size-fits-all. I agree that Crit/Fatal is something to be avoided. I feel Error is useful as a <bold>Info</bold> so super important lines don't drown in a sea of Info. Info is where most of our logging should be. Debug should be sparing and is something that gives us the option, which we may not ultimately exercise, to log more to disk and less to LogLens.

ryancouto:
I agree on the one-size-fits-all assumption, just wanted to show my thought process. I'm fine with removing Crit from ours and adding in Error. Additionally, w.r.t. fmt calls, going to leave fmt.Errorf in as is, as we sometimes use the returned error value.

ryancouto:
@dgassaway @jasonbs10 , @JeanetteBruno and I have been talking about the log PR I posted today. It seems that log15 doesn't provide any string formatting options provided by the stdlib, (e.g. Printf, Infof, etc.) but logrus does. Thinking that in the interest of not changing around the format of our logged statements too much, it might be better to go with logrus.

jasonbs10:
I think that's the whole point of structured logging... maybe print some example lines and see if they're the same level of effort to parse useful information out of them.

PR #231, opened 2017-03-30
rcouto/logrus

ryancouto:
This PR is the alternative to https://github.com/scootdev/scoot/pull/230, using logrus instead of log15. There is no structured logging included or modifications to logged statements, just import path changes and fmt replacements.


binaries/apiserver/main.go
```diff
@@ -25,7 +25,7 @@ func main() {
 	configFlag := flag.String("config", "{}", "API Server Config (either a filename like local.local or JSON text")
 	flag.Parse()
 
-	log.SetFlags(log.LstdFlags | log.LUTC | log.Lshortfile)
+	// log.SetFlags(log.LstdFlags | log.LUTC | log.Lshortfile)
```
dgassaway: Prefer deletion, or better: setup with Logrus' SetFormatter(TextFormatter), SetLevel, etc

jasonbs10: We should definitely continue to include file:Line# in this PR.

saga/saga_generators.go
```diff
@@ -39,7 +40,7 @@ func genSagaState(genParams *gopter.GenParameters, includeJob bool) *SagaState {
 	state, err := makeSagaState(sagaId, job)
 
 	if err != nil {
-		fmt.Println(err)
+		log.Println(err)
```
JeanetteBruno: change Print to Info?

binaries/apiserver/main.go
JeanetteBruno: +1 Using SetFormatter(TextFormatter)

async/Mailbox_test.go
```diff
@@ -58,7 +58,7 @@ func storeValue_withMailbox(num int) error {
 			successfulWrites++
 		}
 		returnedWrites++
-		fmt.Println("completedWrites", returnedWrites)
+		log.Println("completedWrites", returnedWrites)
```
JeanetteBruno: do we want to convert all log.Print to log.Info?  (I also saw one in sagaGenerators... not sure if there are more lurking out there.

binaries/apiserver/main.go
ryancouto: Should've said this in the summary, but I was planning to include file:Line# in the subsequent Twitter source phab, I think it makes sense to leave OSS as barebones as possible and do all custom configuration within source.

async/Mailbox_test.go
ryancouto: Yes, thank you.

PR #232, opened 2017-04-03
First pass at scheduler resiliency.

jasonbs10:
This adds the concept of lost and flaky workers, plus retrying scheduler->worker requests when we see a thrift error.

I'm still fixing a couple bugs and adding tests, so this PR is for early feedback.

From convo with jbruno:
[7:23 AM] Jason Schiller: lost means it's been reported lost by aurora and it *may* be the case that this was a blip and it'll report healthy again
[7:23 AM] Jason Schiller: flaky means aurora hasn't reported anything but we've experienced issues at the connection level.
[7:24 AM] Jason Schiller: if we never log info about lost nodes then we can assume that when aurora lists the available nodes it's always accurate, and then we can remove lost node logic I think

And for task_runner retrying, we make multiple attempts to run() and try to handle the case where a request was processed but there was an issue with the response. We also retry querying after a run has started but there's no mutable state involved on the worker side for that.


sched/scheduler/stateful_scheduler.go
```diff
@@ -315,12 +315,13 @@ func (s *statefulScheduler) scheduleTasks() {
...
-					log.Println("Error running job:", jobId, ", task:", taskId, " command:", strings.Join(taskDef.Argv, " "), retry)
+					log.Println("Error running job:", jobId, ", task:", taskId,
+						" command:", strings.Join(taskDef.Argv, " "), retry, " err:", err)
 					jobState.errorRunningTask(taskId, err)
 				}
 
-				// update cluster state that this node is now free
-				s.clusterState.taskCompleted(nodeId, taskId)
+				// update cluster state that this node is now free and if we had a non-domain (ex: thrift) error.
+				s.clusterState.taskCompleted(nodeId, taskId, (err != nil))
```
jasonbs10: Yeah, should there be another place?

ryancouto: Nope, just making sure I wasn't missing something

sched/scheduler/cluster_state.go
```diff
@@ -106,23 +134,58 @@ func (c *clusterState) updateCluster() {
...
 		switch update.UpdateType {
 		case cluster.NodeAdded:
-			// add the node if it doesn't already exist
-			if _, ok := c.nodes[update.Node.Id()]; !ok {
-				c.nodes[update.Node.Id()] = newNodeState(update.Node)
-				c.nodeGroups[""].idle[update.Node.Id()] = update.Node
-				log.Infof("Added node: %+v, now have %d nodes\n", update.Node, len(c.nodes))
+			if ns, ok := c.lostNodes[update.Id]; ok {
+				// This node was lost earlier (assuming id is unique to a running node instance), we can recover it now.
+				ns.timeLost = nilTime
```
ryancouto: Seems weird to have both nilTime and time.Time{} being used for assignment.

sched/scheduler/task_runner.go
```diff
@@ -78,32 +81,70 @@ func (r *taskRunner) run() error {
...
+	id = st.RunID
 
-	id := st.RunID
-
+	log.Infof("Query(running) for job:%s, taskId:%s", r.jobId, taskId)
 	// Wait for the process to start running
-	st, err = r.queryWithTimeout(id, endTime, true)
-	if err != nil || st.State.IsDone() {
-		return st, err
+	elapsedRetryTime = 0
```
dgassaway: this and 3rd assignment below - consistency vs set to time.Duration(0) above?

sched/scheduler/task_runner.go
```diff
@@ -78,32 +81,70 @@ func (r *taskRunner) run() error {
...
+			time.Sleep(DefaultRetryInterval)
+			elapsedRetryTime += DefaultRetryInterval
+			continue
+		} else if err != nil || st.State.IsDone() {
+			return st, err
+		}
+		break
 	}
 
 	// It's running, but not done, so we want to log a second StartTask that includes
```
dgassaway: Suggest folding the 3rd loop into this one to get rid of duplication, and use a e.g. if notStarted flag + check for task log.  Also with this as-is, if the task exits quickly enough and the first poll in the 2nd loop gets IsDone, start task will never be logged?  Lastly would prefer for loop w/ conditionals vs manual control statements but willing to defer to preference

workerapi/server/server.go
```diff
@@ -105,32 +118,49 @@ func (h *handler) QueryWorker() (*worker.WorkerStatus, error) {
...
 	defer h.stat.Latency("runLatency_ms").Time().Stop()
 	h.stat.Counter("runs").Inc(1)
-	log.Infof("Worker Running:%s", cmd)
+	log.Infof("Worker trying to run cmd: %v", cmd)
 
 	h.updateTimeLastRpc()
 	c := domain.ThriftRunCommandToDomain(cmd)
-	process, err := h.run.Run(c)
+	status, err := h.run.Run(c)
+	//Check if this is a dup retry for an already running command and if so get its status.
```
dgassaway: Is this check adequate? What if argv is the same but snapshot id or env is different?

scootapi/setup/workers.go
```diff
@@ -97,6 +98,7 @@ func (s *LocalWorkersStrategy) StartupWorkers() (string, error) {
 		thriftPort := s.nextPort
 		s.nextPort++
 		if err := s.cmds.Start(bin,
+			"--loglevel", "info",
```
JeanetteBruno: don't understand why we have ln 101 and 104.

jasonbs10: Thank you for catching that. This was a bad merge.

sched/scheduler/cluster_state.go
```diff
@@ -103,26 +129,64 @@ func (c *clusterState) updateCluster() {
...
+				delete(c.suspendedNodes, update.Id)
+				log.Infof("Recovered suspended node %v (%v), now have %d healthy nodes, %d suspended nodes",
+					update.Id, ns, len(c.nodes), len(c.suspendedNodes))
+			} else if ns, ok := c.nodes[update.Id]; !ok {
+				// This is a new unrecognized node, add it to the cluster.
+				c.nodes[update.Id] = newNodeState(update.Node)
+				c.nodeGroups[""].idle[update.Id] = c.nodes[update.Id]
+				log.Infof("Added new node: %v (%+v), now have %d nodes", update.Id, update.Node, len(c.nodes))
+			} else {
+				// This node is already present, log this spurious add.
```
JeanetteBruno: is this a symptom of a bigger issue?  Should we log it as an 'ignored' error - use Error level, but indicate that this error was ignored.  Same with ln 170 below.

jasonbs10: This really shouldn't happen and some codebases might assert() here. For our needs, I wouldn't panic or use error unless it's for something that might cause a customer-facing issue.

PR #233, opened 2017-04-04
set log level flag

ryancouto:
Allows binaries to take a flag to set their log level. This way we can disable debug for noisy logs in travisCI, etc.


binaries/setup-cloud-scoot/main.go
```diff
@@ -31,13 +38,13 @@ func main() {
 
 	workersCfg := &setup.WorkerConfig{Count: *workersFlag, RepoDir: *repoDir}
 	sched := map[string]setup.SchedulerStrategy{
-		"local.memory": setup.NewLocalMemory(workersCfg, builder, cmds),
-		"local.local":  setup.NewLocalLocal(workersCfg, builder, cmds),
+		"local.memory": setup.NewLocalMemory(workersCfg, builder, cmds, level),
```
jasonbs10: Can you move level into the WorkerConfig struct instead? Same for scheduler and apiserver.

scootapi/setup/api.go
```diff
@@ -26,17 +26,19 @@ type ApiConfig struct {
...
-func NewLocalApiStrategy(apiCfg *ApiConfig, builder Builder, cmds *Cmds) *LocalApiStrategy {
+func NewLocalApiStrategy(apiCfg *ApiConfig, builder Builder, cmds *Cmds, level log.Level) *LocalApiStrategy {
 	return &LocalApiStrategy{
-		apiCfg:  apiCfg,
-		builder: builder,
-		cmds:    cmds,
+		apiCfg:   apiCfg,
+		builder:  builder,
+		cmds:     cmds,
+		logLevel: level,
```
jasonbs10: For all of these strategies, can you move "level" into the higher level *Config struct and then access it as localApiStrategy.apiCfg.level?

tests/testhelpers/clusterHelpers.go
```diff
@@ -22,10 +22,10 @@ func CreateLocalTestCluster() (*setup.Cmds, error) {
 	builder := setup.NewGoBuilder(clusterCmds)
 	go func() {
 		sched := map[string]setup.SchedulerStrategy{
-			"local.local": setup.NewLocalLocal(nil, builder, clusterCmds),
+			"local.local": setup.NewLocalLocal(nil, builder, clusterCmds, log.DebugLevel),
```
jasonbs10: Is DebugLevel the right choice here? I don't actually know, but we for sure don't want debug in TravisCI.

binaries/setup-cloud-scoot/main.go
ryancouto: There isn't a SchedulerConfig, so the scheduler's log level will be set through the WorkerConfig.

tests/testhelpers/clusterHelpers.go
ryancouto: No, you're right. Info is a better default.

jasonbs10:
lgtm

PR #234, opened 2017-04-06
updating run job and status to print results to stdout so that Sickle…

JeanetteBruno:
… can pick them up

### DIFF
--- a/scootapi/client/get_status_cmd.go
+++ b/scootapi/client/get_status_cmd.go
@@ -51,8 +51,10 @@ func (c *getStatusCmd) run(cl *simpleCLIClient, cmd *cobra.Command, args []strin
 			return fmt.Errorf("Error converting status to JSON: %v", err.Error())
 		}
 		log.Infof("%s\n", asJson)
+		fmt.Printf("%s\n", asJson) // must also go to stdout so Sickle can find the results
 	} else {
 		log.Info("Job Status:", status)
+		fmt.Printf("Job Status:", status) // must also go to stdout so Sickle can find the results
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


jasonbs10:
Can you explain why this change is technically necessary? Sickle code looks like it should pick up both stderr/stdout so these prints seem redundant. If there's a problem I'd expect us to resolve it within sickle itself.

scootapi/client/get_status_cmd.go
```diff
@@ -51,8 +51,10 @@ func (c *getStatusCmd) run(cl *simpleCLIClient, cmd *cobra.Command, args []strin
 			return fmt.Errorf("Error converting status to JSON: %v", err.Error())
 		}
 		log.Infof("%s\n", asJson)
+		fmt.Printf("%s\n", asJson) // must also go to stdout so Sickle can find the results
 	} else {
 		log.Info("Job Status:", status)
+		fmt.Printf("Job Status:", status) // must also go to stdout so Sickle can find the results
```
dgassaway: Println

scootapi/client/run_job_cmd.go
```diff
@@ -98,8 +98,8 @@ func (c *runJobCmd) run(cl *simpleCLIClient, cmd *cobra.Command, args []string)
 		}
 	}
 
-	log.Info(jobId.ID)
-	log.Infof("JobID:%s\n", jobId.ID)
+	fmt.Println(jobId.ID) // must go to std out so Sickle can pick up the results
+	log.Info("JobID:%s\n", jobId.ID)
```
dgassaway: With Info vs Infof, can drop the \n?

PR #237, opened 2017-04-11
First pass at scheduler tracking worker init status.

jasonbs10:
Looking for feedback on this approach and any general suggestions. Tests to come later.

The basic idea here is that new nodes should be initially tracked as suspended nodes and moved to the healthy node pool once they've completely started up. This is mostly plumbing.

### DIFF
--- a/runner/runners/status_manager.go
+++ b/runner/runners/status_manager.go
@@ -22,7 +22,7 @@ func NewStatusManager() *StatusManager {
 type StatusManager struct {
 	mu        sync.Mutex
 	runs      map[runner.RunID]runner.RunStatus
-	service   runner.ServiceStatus
+	svcStatus runner.ServiceStatus
 	nextRunID int64
 	listeners []queryAndCh
 }
@@ -51,15 +51,15 @@ func (s *StatusManager) NewRun() (runner.RunStatus, error) {
 }
 
 // Update the overall service status independent of run status.
-func (s *StatusManager) UpdateService(service runner.ServiceStatus) error {
+func (s *StatusManager) UpdateService(svcStatus runner.ServiceStatus) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	s.service = service
+	s.svcStatus = svcStatus
 	return nil
 }
 
-// Update writes a new status.
+// Update writes a new status for a run.
 // It enforces several rules:
 //   cannot change a status once it is Done
 //   cannot erase Stdout/Stderr Refs
@@ -102,7 +102,7 @@ func (s *StatusManager) Update(newStatus runner.RunStatus) error {
 func (s *StatusManager) Query(q runner.Query, wait runner.Wait) ([]runner.RunStatus, runner.ServiceStatus, error) {
 	current, future, err := s.queryAndListen(q, wait.Timeout != 0)
 	if err != nil || len(current) > 0 || wait.Timeout == 0 {
-		return current, s.service, err
+		return current, s.svcStatus, err
 	}
 
 	var timeout <-chan time.Time
@@ -114,9 +114,9 @@ func (s *StatusManager) Query(q runner.Query, wait runner.Wait) ([]runner.RunSta
 
 	select {
 	case st := <-future:
-		return []runner.RunStatus{st}, s.service, nil
+		return []runner.RunStatus{st}, s.svcStatus, nil
 	case <-timeout:
-		return nil, s.service, nil
+		return nil, s.svcStatus, nil
 	}
 }
 
--- a/runner/runners/status_manager.go
+++ b/runner/runners/status_manager.go
@@ -22,6 +22,7 @@ func NewStatusManager() *StatusManager {
 type StatusManager struct {
 	mu        sync.Mutex
 	runs      map[runner.RunID]runner.RunStatus
+	service   runner.ServiceStatus
 	nextRunID int64
 	listeners []queryAndCh
 }
@@ -49,6 +50,15 @@ func (s *StatusManager) NewRun() (runner.RunStatus, error) {
 	return st, nil
 }
 
+// Update the overall service status independent of run status.
+func (s *StatusManager) UpdateService(service runner.ServiceStatus) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	s.service = service
+	return nil
+}
+
 // Update writes a new status.
 // It enforces several rules:
 //   cannot change a status once it is Done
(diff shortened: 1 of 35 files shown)


sched/scheduler/cluster_state.go
```diff
@@ -137,62 +146,95 @@ func (c *clusterState) updateCluster() {
...
+
 			} else {
 				// We don't know about this node, log spurious remove.
 				log.Infof("Cannot remove unknown node: %v", update.Id)
 			}
 		}
 	}
 
-	// Clean up lost nodes that haven't recovered in time, and put flaky nodes back in rotation.
+	// Clean up lost nodes that haven't recovered in time, add flaky nodes back into rotation after some time,
```
JeanetteBruno: I think this allows nodes that are stuck initializing to be stuck forever...also not sure why we are reinstating flaky based on time... alternative approach 

Have an other loop (in scheduler?)  that is checking all suspendedNodes for good response to 'StatusAll', if get healthy status, reinstate the node, at end of loop drop all nodes that have been suspended too long.

sched/scheduler/stateful_scheduler.go
```diff
@@ -107,6 +107,23 @@ func NewStatefulScheduler(
...
 ) *statefulScheduler {
 
+	nodeReadyFn := func(node cluster.Node) bool {
+		run := rf(node)
+		st, err := run.StatusAll()
+		if err != nil {
+			return false
+		}
+		if len(st) == 1 && st[0].Error == runner.NoRunnersMsg {
+			//TODO(jschiller) amend protocol to be more direct than this array of one with a particular error status.
```
JeanetteBruno: agreed - recommend adding a healthCheck() or Status() API - returning  starting, running, idle, or error (not necessarily in this pr)

snapshot/git/gitdb/db.go
```diff
@@ -34,17 +37,18 @@ const (
...
 // MakeDBNewRepo makes a gitDB that uses a new DB, populated by initer
 func MakeDBNewRepo(initer RepoIniter, tmp *temp.TempDir, stream *StreamConfig,
-	tags *TagsConfig, bundles *BundlestoreConfig, autoUploadDest AutoUploadDest) *DB {
+	tags *TagsConfig, bundles *BundlestoreConfig, autoUploadDest AutoUploadDest) (*DB, InitCh) {
 	return makeDB(nil, initer, tmp, stream, tags, bundles, autoUploadDest)
 }
 
 func makeDB(dataRepo *repo.Repository, initer RepoIniter, tmp *temp.TempDir, stream *StreamConfig,
-	tags *TagsConfig, bundles *BundlestoreConfig, autoUploadDest AutoUploadDest) *DB {
+	tags *TagsConfig, bundles *BundlestoreConfig, autoUploadDest AutoUploadDest) (*DB, InitCh) {
```
JeanetteBruno: don't need to return InitCh, its in *DB already, and it doesn't look like it's used anywhere.

snapshot/git/gitdb/db.go
```diff
@@ -13,6 +13,9 @@ import (
 // kind instead of type because type is a keyword
 type snapshotKind string
 
+// Returned by MakeDBNewRepo(), gets closed as soon as the repo is initialized.
+type InitCh <-chan struct{}
```
JeanetteBruno: name InitDoneCh?

snapshot/git/gitdb/setup.go
```diff
@@ -34,7 +35,12 @@ func (m module) Install(b *ice.MagicBag) {
 			return AutoUploadBundlestore
 		},
 		MakeDBNewRepo,
-		func(db *DB) snap.DB { return db },
+		func(db *DB) snap.DB {
+			return db
+		},
+		func(ic InitCh) runner.InitCh {
```
JeanetteBruno: rename to InitDoneCh?

snapshot/git/gitdb/db.go
jasonbs10: It's closed in gitdb.db.init and listened to in workerapi/server/server.go, and DB is an opaque type so it may be available in DB but it isn't available to any other packages.

dgassaway:
I'm fine with this approach to check initializing only for now, but agreed w/ @JeanetteBruno for reconciling on the various states later.

snapshot/git/gitdb/db.go
jasonbs10: Done.

snapshot/git/gitdb/setup.go
jasonbs10: Done.

sched/scheduler/cluster_state.go
jasonbs10: We can try to condense some of this state in another PR. I think the bigger problem is that I'm making multiple RPC calls in the main loop...

sched/scheduler/stateful_scheduler.go
jasonbs10: Done, added the concept of ServiceStatus.

runner/runners/status_manager.go
```diff
@@ -22,6 +22,7 @@ func NewStatusManager() *StatusManager {
 type StatusManager struct {
 	mu        sync.Mutex
 	runs      map[runner.RunID]runner.RunStatus
+	service   runner.ServiceStatus
```
JeanetteBruno: serviceStatus or just status

runner/runners/status_manager.go
```diff
@@ -49,13 +50,23 @@ func (s *StatusManager) NewRun() (runner.RunStatus, error) {
 	return st, nil
 }
 
+// Update the overall service status independent of run status.
+func (s *StatusManager) UpdateService(service runner.ServiceStatus) error {
```
JeanetteBruno: serviceStatus or just status

runner/runners/status_manager.go
```diff
@@ -49,13 +50,23 @@ func (s *StatusManager) NewRun() (runner.RunStatus, error) {
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	s.service = service
+	return nil
+}
+
 // Update writes a new status.
```
JeanetteBruno: Update writes a new status for a Run.

runner/status.go
```diff
@@ -151,3 +151,12 @@ func PreparingStatus(runID RunID) (r RunStatus) {
 	r.State = PREPARING
 	return r
 }
+
+// This is for overall runner status, just 'initialized' status for now.
+type ServiceStatus struct {
```
JeanetteBruno: do we want to have a struct with member variable or int type and use enum (iota) values?

jasonbs10: I think it's fine this way until we have more specific needs (and we already track if a node is idle, though I guess there's some value to making that check explicit).

JeanetteBruno: ok

PR #238, opened 2017-04-14
Set verbose log output to debug

ryancouto:
and set default level to info instead of returning on error parsing

### DIFF
--- a/binaries/apiserver/main.go
+++ b/binaries/apiserver/main.go
@@ -32,7 +32,7 @@ func main() {
 	level, err := log.ParseLevel(*logLevelFlag)
 	if err != nil {
 		log.Error(err)
-		return
+		level = "info"
 	}
 	log.SetLevel(level)
 
--- a/binaries/apiserver/main.go
+++ b/binaries/apiserver/main.go
@@ -32,7 +32,7 @@ func main() {
 	level, err := log.ParseLevel(*logLevelFlag)
 	if err != nil {
 		log.Error(err)
-		level = log.InfoLevel
+		return
 	}
 	log.SetLevel(level)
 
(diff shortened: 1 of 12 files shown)


binaries/apiserver/main.go
```diff
@@ -32,7 +32,7 @@ func main() {
 	level, err := log.ParseLevel(*logLevelFlag)
 	if err != nil {
 		log.Error(err)
-		return
+		level = log.InfoLevel
```
jasonbs10: Why do we want to continue starting up if there's an error parsing the level? Startup is the perfect time to bomb due to errors.

JeanetteBruno: +1.   Also - does the logrus Fatal trigger a call to os.Exit(1) (as the log package does?)   The logrus docs just say it logs a message at the Fatal level doesn't mention os.Exit(1).... the next question is do we want it to exit?

ryancouto: Reverting. And it does Exit(1), 
https://github.com/sirupsen/logrus/blob/master/entry.go#L164
https://github.com/sirupsen/logrus/blob/547e984ad93a90192134fc0fdb57a468edb514fe/alt_exit.go#L49

JeanetteBruno:
lgtm

ice/eval.go
```diff
@@ -36,8 +36,8 @@ func (bag *MagicBag) Extract(dest interface{}) (result error) {
 	targetVal.Set(reflect.Value(eval.construct(targetType)))
 
 	s := spew.Sdump(dest)
-	log.Info("************* Extracted:")
-	log.Info(s)
+	log.Debug("************* Extracted:")
```
jasonbs10: I deleted this in my latest PR so we should just drop this change.

PR #239, opened 2017-04-17
scootapi/job_status - initialize TaskData and TaskStatus fields when job is not started

dgassaway:
* Init JobStatus fields
* remove unrelated unnecessary if block

Not totally sure how to repro

### DIFF
--- a/scootapi/server/job_status.go
+++ b/scootapi/server/job_status.go
@@ -16,13 +16,11 @@ func GetJobStatus(jobId string, sc s.SagaCoordinator) (*scoot.JobStatus, error) 
 		js.ID = ""
 		js.Status = scoot.Status_NOT_STARTED
 
-		if err != nil {
-			switch err.(type) {
-			case s.InvalidRequestError:
-				err = scoot.NewInvalidRequest()
-			case s.InternalLogError:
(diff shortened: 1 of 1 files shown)


scootapi/server/job_status.go
```diff
@@ -33,6 +31,8 @@ func GetJobStatus(jobId string, sc s.SagaCoordinator) (*scoot.JobStatus, error)
 		js := scoot.NewJobStatus()
 		js.ID = jobId
 		js.Status = scoot.Status_NOT_STARTED
+		js.TaskStatus = make(map[string]scoot.Status)
+		js.TaskData = make(map[string]*scoot.RunStatus)
```
jasonbs10: Can you also loop through the task data and add NOT_STARTED?

dgassaway: I don't think it has access to what the tasks are - state from GetSagaState here is nil, and the sagaState is what convertSagaStateToJobStatus relies on to update the task statuses.

jasonbs10: Nm, this is a bigger change then. I think a StartSaga needs to be logged before a jobID is returned so I'm not sure why we're even hitting this block of code...

PR #242, opened 2017-04-25
runner: pass through jobID and taskID

ryancouto:
https://jira.twitter.biz/browse/DX-2151


ryancouto:
@JeanetteBruno I thought that was another story. Not sure if we want that to be unique to each client, request, or task. Additionally not sure if we want to allow users to provide their own request tag or if it should be generated similar to a JobID.

@dgassaway agreed, I'm wondering if maybe they should be optional fields to mitigate risk of future changes.

JeanetteBruno:
I created DX-2156 for the external tag concept.  (This tag would be unique to each job request.)

PR #244, opened 2017-04-26
bundlestore: Propagate TTL config across client and server.

jasonbs10:


snapshot/bundlestore/store.go
```diff
@@ -2,8 +2,18 @@ package bundlestore
 
 import (
 	"io"
+	"time"
 )
 
+const DefaultTTL time.Duration = time.Hour * 24 * 30 //30 days.
+const DefaultTTLKey string = "x-scoot-expires"
```
dgassaway: Comment about this key? as awkward as it is in this repo

snapshot/bundlestore/http_store.go
```diff
@@ -60,13 +61,29 @@ func (s *httpStore) Exists(name string) (bool, error) {
...
+			return nil, err
+		}
+		req.Header.Set("Content-Type", "text/plain")
+		if ttl == nil {
+			ttl = &TTLConfig{DefaultTTL, DefaultTTLKey}
+		}
+		if ttl.TTL != 0 {
+			req.Header[ttl.TTLKey] = []string{ttl.TTL.String()}
+		}
+		return http.DefaultClient.Do(req)
```
ryancouto: Why use http.DefaultClient instead of s.client?

snapshot/bundlestore/server.go
```diff
@@ -2,22 +2,28 @@ package bundlestore
...
 	store Store
+	ttl   *TTLConfig
 	stat  stats.StatsReceiver
 }
 
-func MakeServer(s Store, stat stats.StatsReceiver) *Server {
-	return &Server{s, stat.Scope("bundlestoreServer")}
+// Make a new server that delegates to an underlying store.
+// TTL may be nil, in which case defaults are applied downstream.
+// TTL duration may be overriden by request headers, but we always pass this ttl key to the store.
```
JeanetteBruno: I tripped over 'but we always pass this ttl key to the store'... interpreted 'this' as ttl in argument - which implies that the server's ttl overrides ttl in request header....
remove  'but we always.... ' from comment

snapshot/bundlestore/file_store.go
```diff
@@ -47,7 +48,7 @@ func (s *FileStore) Exists(name string) (bool, error) {
 	return false, err
 }
 
-func (s *FileStore) Write(name string, data io.Reader) error {
+func (s *FileStore) Write(name string, data io.Reader, ttl *TTLConfig) error {
```
JeanetteBruno: We should put a comment in this file that file_store ignores the ttl.

PR #246, opened 2017-04-28
Jschiller/bughunt

jasonbs10:
Added a bunch of logging and fixed a number of issues including:
- worker panic
- incorrect ttl date format
- mem_cap reporting
- queue.go initDoneCh omission
- inconsistent scheduler retry policy
- sagalog end task failure handling


runner/execer/os/os_execer.go
```diff
@@ -82,8 +83,12 @@ type osProcess struct {
 
 // Periodically check to make sure memory constraints are respected.
 func (p *osProcess) monitorMem(memCap execer.Memory, stat stats.StatsReceiver) {
-	memTicker := time.NewTicker(100 * time.Millisecond)
+	prevThreshold := 0.0   // The last reported (usage / mem_cap)
+	reportThreshold := 0.0 // Start reporting once we hit this utilization threshold.
```
JeanetteBruno: shouldn't this be higher? (75%, 80%)

JeanetteBruno: 1. shouldn't this be higher (to reduce log messages) - (60%, 70% or 80%?) - also if we increase this threshold we should probably decrease the increment to something like 5%...
2. do we want to report if/when usage goes back under the threshold?

runner/runners/invoke.go
```diff
@@ -154,7 +155,7 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 	case st = <-processCh:
 	}
 
-	log.Infof("run done. id: %v, status: %+v, cmd: %+v, checkout: %v", id, st, cmd, co.Path())
+	log.Infof("run done. id: %v, status: %+v, checkout: %v", id, st, co.Path())
```
JeanetteBruno: I have a utility to split sickle and scheduler logs into lists of successes/failures - if we think it would be useful for worker logs we should change this to match (or at least be closer to scheduler or sickle format:
run done:  <cmd>
Status: COMPLETE
Exit Code: <exit code>
Stdout: <stdout ID/URL>
Stderr: <stderr ID/URL>

runner/runners/queue.go
```diff
@@ -72,10 +72,17 @@ func (c *QueueController) Run(cmd *runner.Command) (runner.RunStatus, error) {
...
+		ready = true
+	default:
+		ready = (c.initDoneCh == nil)
+	}
+	log.Infof("Trying to run, ready=%t, available slots:%d/%d, currentRun:%s cmd:%v",
+		ready, c.capacity+1-numCmds, c.capacity+1, c.runningID, cmd)
 
-	if numCmds > c.capacity {
+	if !ready || numCmds > c.capacity {
 		return runner.RunStatus{}, fmt.Errorf(QueueFullMsg)
```
JeanetteBruno: sending queue full msg when the worker has not finished initializing.  This will mask the issues that there is something wrong with the scheduler's waiting for worker's to finish initializing before sending a run request to a worker.  Recommend we split the errors into 'worker not initialized yet' vs 'queue full'

sched/scheduler/stateful_scheduler.go
```diff
@@ -351,25 +353,41 @@ func (s *statefulScheduler) scheduleTasks() {
...
+					} else {
+						jobState.errorRunningTask(taskId, err)
+					}
+					log.Info(msg, jobId, ", task:", taskId, " err:", taskErr, " cmd:", taskDef.Argv)
+
+					// If the task completed succesfully but sagalog failed, start a goroutine to retry until it succeeds.
+					if taskErr.sagaErr != nil && taskErr.runnerErr == nil && taskErr.resultErr == nil {
+						log.Info(msg, jobId, ", task:", taskId, " -> starting goroutine to handle failed sagalog. ")
+						go func() {
+							for err := errors.New(""); err != nil; err = run.logTaskStatus(&taskErr.st, saga.EndTask) {
```
JeanetteBruno: running forever till success seems dangerous... should we have TODO about putting a cap on it and sending an alert?

runner/runners/invoke.go
ryancouto: +1, I'm all for standardizing output

sched/scheduler/stateful_scheduler.go
dgassaway: Yeah, how can we do this in the background? Isn't this a requirement to proceed safely?

ryancouto: +1, I'd lean towards just throwing a timeout in there for now and TODOing the alert

snapshot/bundlestore/store.go
```diff
@@ -14,9 +29,10 @@ type StoreRead interface {
 }
 
 // Write operations on store, limited to a one-shot writing operation since bundles are immutable.
+// If ttl config is nil then the store will use its defaults.
 type StoreWrite interface {
 	// Does a streaming write of the given bundle. There is no concept of partial writes (partial=failed).
-	Write(name string, data io.Reader) error
+	Write(name string, data io.Reader, ttl *TTLValue) error
```
JeanetteBruno: is this span to keep alive or expiration date/time? (same as prior question)

snapshot/bundlestore/server.go
```diff
@@ -58,7 +67,28 @@ func (s *Server) HandleUpload(w http.ResponseWriter, req *http.Request) {
 		fmt.Fprintf(w, "Bundle %s already exists\n", bundleName)
 	}
 
-	if err := s.store.Write(bundleName, bundleData); err != nil {
+	// Get ttl if defaults were provided during Server construction or if it comes in this request header.
+	var ttl *TTLValue
```
JeanetteBruno: is ttl in the expiration date or span to keep alive?  If its expiration date we should call it that.

runner/execer/os/os_execer.go
jasonbs10: I reworked this, please re-review.

runner/runners/invoke.go
jasonbs10: I'm punting on this, it can be done as part of a more concerted effort later on if necessary.

runner/runners/queue.go
jasonbs10: Reworked this section to fix some bugs.

sched/scheduler/stateful_scheduler.go
jasonbs10: There are two options here as I see it - keep trying until we manually intervene, or panic since we can't communicate status to our clients. I'm doing the first thing here since it seems the better of the two (also it really doesn't matter for Q2 since we use in-memory sagalog).

runner/execer/os/os_execer.go
```diff
@@ -82,8 +83,12 @@ type osProcess struct {
...
+	reportIncrement := .1  // Report every time we increase utilization threshold thi amount.
```
jasonbs10: Reworked.

snapshot/bundlestore/store.go
jasonbs10: TTLValue uses time.Time so it self-documents as expiration time which I think is sufficient.

PR #248, opened 2017-04-28
scootapi/scoot - Add timeouts to scootapi thrift, support in client run_job

dgassaway:
https://jira.twitter.biz/browse/DX-2130
* add timeouts to job/task def in scootapi
* job def timeout applies to tasks, unless specified in task def
* add equivalent fields to client run_job structs
* fix 2 `go vet` unspecified field errors

### DIFF
--- a/scootapi/client/run_job_cmd.go
+++ b/scootapi/client/run_job_cmd.go
@@ -55,7 +55,7 @@ func (c *runJobCmd) run(cl *simpleCLIClient, cmd *cobra.Command, args []string) 
 			log.Info("No snapshotID provided - cmd will be run in an empty tmpdir.")
 		} else if !strings.Contains(c.snapshotId, "-") {
 			//this is not a bundleID, assume it's a sha that's available upstream. Cf. snapshot/git/gitdb/README.md
-			streamId := fmt.Sprintf("stream-swh-%s-%s", c.streamName, c.snapshotId)
+			streamId := fmt.Sprintf("stream-gc-%s-%s", c.streamName, c.snapshotId)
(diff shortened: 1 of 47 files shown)


scootapi/client/run_job_cmd.go
```diff
@@ -33,13 +33,15 @@ func (c *runJobCmd) registerFlags() *cobra.Command {
 
 // Types to handle JobDefinitions from JSON files
 type JobDef struct {
-	Tasks map[string]TaskDef
+	Tasks     map[string]TaskDef
+	TimeoutMs int32
```
jasonbs10: We should take this out until we're ready to implement it.

dgassaway:
I think this might actually be covered by scootapi/server/translate_prop_test.go, but I also manually verified it

scootapi/client/run_job_cmd.go
dgassaway: Not following this

jasonbs10: Task timeout appears to flow through but job timeout is dropped in scheduler. We probably shouldn't add things unless we have a concrete timeline for using them.

dgassaway: Ah gotcha - these get resolved in scootapi/server/run_job.go - task timeout will be used, if it's not set then job timeout will be used

jasonbs10: Does that conform to expected behavior? Job timeout should either be called defaultTaskTimeout in this case or it should be used to limit total job time imo.

dgassaway: Agreed on defaultTaskTimeout - it's just a shorthand vs including the field in each task. Limiting job time would be trickier, and different use case.

PR #249, opened 2017-04-28
vendor: Use scoot copy of thrift

ryancouto:
Apache deleted their thrift repo today, github.com/apache/thrift. There is now a copy of their repo located at github.com/scootdev/thrift. This PR replaces all references to the former with the latter. It also required a couple changes to deps.sh, as we have transitive dependencies on github.com/apache/thrift.

### DIFF
--- a/deps.sh
+++ b/deps.sh
@@ -15,15 +15,20 @@ GOPATH_NEW="$(mktemp -d -t TEMP.XXXXXXX)"
 export GOPATH="${GOPATH_NEW}"
 
 get_deps() {
-    cd "${GOPATH}/src/$1"
-    for need in $(go list -f '{{join .Deps "\n"}}' ./... | \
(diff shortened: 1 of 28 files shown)


deps.sh
```diff
@@ -15,15 +15,20 @@ GOPATH_NEW="$(mktemp -d -t TEMP.XXXXXXX)"
 export GOPATH="${GOPATH_NEW}"
 
 get_deps() {
```
JeanetteBruno: should we have a comment explaining why we have to replace the apache thrift references with scoot thrift references (is it in case a vendor requires thrift?)

jasonbs10: This PR was aborted. Is this comment pertinent to what we currently do?

JeanetteBruno: I hadn't realized the pr was aborted.

PR #252, opened 2017-05-09
sched: Initial plumbing for new scheduler strategy.

jasonbs10:
Please skim this diff for obvious flaws (besides lack of code comments...)

-Added priority, tag, basis, and requestor fields to scootapi.
-Plumbing for priorities 0,1 where idle nodes are assigned to tasks.
-Partial plumbing for priorities 2,3 namely the ability to kill a task, but nothing that exercises it.

See: https://docs.google.com/document/d/1Jbdz_96n2Bq1e4_TMX_HAOj14P8jTAmN9XXtLU68XV0

### DIFF
--- a/sched/definitions.go
+++ b/sched/definitions.go
@@ -45,7 +45,7 @@ type JobDefinition struct {
 	Requestor string
 	Tag       string
 	Basis     string
-	Priority  int32
+	Priority  Priority
 	Tasks     []TaskDefinition
 }
 
@@ -76,6 +76,23 @@ const (
 	// Job/Task finished unsuccessfully all compensating actions
 	// have been applied.
 	RolledBack
+)
+
+type Priority int
+
+const (
+	// Default, queue new runs until any resources are available
+	P0 Priority = iota
+
+	// Run asap, ahead of priority=0, consuming nodes as they become available if no nodes are free
+	P1
+
+	// Run asap with at least MinRunningNodesForGivenJob, killing youngest lower priority tasks if no nodes are free
+	P2
+
+	// Run immediately trying to acquire as many nodes as required, killing all tasks with lower priority that
+	// have been executing with duration < 10min or so
+	P3
 )
 
 // transforms a thrift Job into a scheduler Job
@@ -131,7 +148,7 @@ func makeDomainJobFromThriftJob(thriftJob *schedthrift.Job) *Job {
 	domainJobDef := JobDefinition{
 		JobType:   jobType,
 		Tasks:     domainTasks,
-		Priority:  priority,
+		Priority:  Priority(priority),
 		Tag:       tag,
 		Basis:     basis,
 		Requestor: requestor,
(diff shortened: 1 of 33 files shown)


sched/scheduler/stateful_scheduler.go
```diff
@@ -257,22 +308,91 @@ func (s *statefulScheduler) step() {
...
+		case checkJobMsg := <-s.checkJobCh:
+			if jobs, ok := s.requestorMap[checkJobMsg.jobDef.Requestor]; !ok {
+				if len(s.requestorMap) >= MaxRequestors {
+					err := fmt.Errorf("Exceeds max number of requestors: %s (%d)", checkJobMsg.jobDef.Requestor, MaxRequestors)
+					checkJobMsg.resultCh <- err
+				}
+			} else if len(jobs) >= MaxJobsPerRequestor {
+				err := fmt.Errorf("Exceeds max jobs per requestor: %s (%d)", checkJobMsg.jobDef.Requestor, MaxJobsPerRequestor)
+				checkJobMsg.resultCh <- err
+			} else if checkJobMsg.jobDef.Priority < 0 || checkJobMsg.jobDef.Priority > 3 {
```
dgassaway: 0, 3 defined as vars somewhere

sched/scheduler/task_scheduler.go
```diff
@@ -33,6 +36,87 @@ func getTaskAssignments(cs *clusterState, tasks []*taskState) ([]taskAssignment,
 		snapshotIds = append(snapshotIds, snapId)
 	}
 
+	// Sort jobs by priority.
```
JeanetteBruno: do we want to encapsulated this to make it easier to try out different approaches?

dgassaway:
Another general comment, this seems like a good time to start introducing scheduling metrics we might find useful related to aborts, queued tasks, etc.

jasonbs10:
This is ready for review.

TODO before merge:
- Fix recoverytest
- Test the new job fields: Priority, Requestor, and Tag.

sched/definitions.go
```diff
@@ -41,8 +41,12 @@ func DeserializeJob(input []byte) (*Job, error) {
 
 // JobDefinition is the definition the client sent us
 type JobDefinition struct {
-	JobType string
-	Tasks   map[string]TaskDefinition
+	JobType   string
+	Requestor string
+	Tag       string
+	Basis     string
+	Priority  Priority
```
jasonbs10: I added this because Drew correctly pointed out that I shouldn't be using 'magic' numbers all over the place, so now we have this enum.

sched/definitions.go
```diff
@@ -94,17 +115,42 @@ func makeDomainJobFromThriftJob(thriftJob *schedthrift.Job) *Job {
...
+			domainTasks = append(domainTasks, TaskDefinition{command})
 		}
 	}
 
 	jobType := ""
 	if thriftJobDef != nil {
 		jobType = thriftJobDef.GetJobType()
 	}
+
+	priority := int32(0)
```
jasonbs10: This is validated in stateful_scheduler.go. I can move it here though.

sched/scheduler/task_scheduler.go
jasonbs10: I'm going to punt on that since it's unlikely we'll have time for a different approach for now.

sched/scheduler/cluster_state.go
```diff
@@ -221,6 +230,7 @@ func (c *clusterState) update(updates []cluster.NodeUpdate) {
 					newNode.readyCh = nil
 					log.Infof("Added new node: %v (%#v), now have %d healthy nodes (%d suspended)",
 						update.Id, update.Node, len(c.nodes), len(c.suspendedNodes))
+					c.numIdle++
```
JeanetteBruno: do we need c.numIdle++ above - where we recover a suspended node?

jasonbs10: I believe so, if it's suspended it's offline so bringing it back puts it in the idle pool again. There are other issues with numIdle that I've already fixed but haven't pushed.

sched/scheduler/job_state.go
```diff
@@ -61,14 +64,24 @@ func newJobState(job *sched.Job, saga *saga.Saga) *jobState {
...
-			j.Tasks[taskId].Status = sched.Completed
+			j.getTask(taskId).Status = sched.Completed
 			j.TasksCompleted++
 		}
 	}
 
 	return j
 }
 
+// Helper, assumes that taskId is present given a consistent jobState.
```
JeanetteBruno: I'm a little worried about performance.... we could maintain the map and make sure the tasks are run in their original order if we generate task id's from the job id: <job_id>_1, <job_id>_2, ... , <job_id>_<numTasks -1> 
(just a thought)

PR #253, opened 2017-05-09
runner/execer/os: Memory usage and cleanup of processes

ryancouto:
- Kill lingering processes with same pgid as monitored command on both success and fail

- Monitor mem usage using process group id instead of parent process id. Gives us a more accurate measure of worker memory usage.

- Structured logging to make verbose ps calls easier to parse

### DIFF
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -206,7 +206,7 @@ func (p *osProcess) Abort() (result execer.ProcessStatus) {
 	result.ExitCode = -1
 	result.Error = "Aborted."
 
-	err = p.cmd.Process.Kill()
+	err := p.cmd.Process.Kill()
 	if err != nil {
 		result.Error = "Aborted. Couldn't kill process."
 	}
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -85,6 +85,10 @@ type osProcess struct {
 	mutex  sync.Mutex
 }
 
+// TODO(rcouto): More we can do here to make sure we're
+// cleaning up after ourselves completely / not leaving
+// orphaned processes behind
+//
 // Periodically check to make sure memory constraints are respected,
 // and clean up after ourselves when the process has completed
 func (e *osExecer) monitorMem(p *osProcess) {
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -62,6 +62,7 @@ func (e *osExecer) Exec(command execer.Command) (result execer.Process, err erro
 		cmd.Stderr = stderrW.WriterDelegate()
 	}
 
+	// Sets pgid of all child processes to cmd's pid
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 
 	err = cmd.Start()
@@ -89,11 +90,14 @@ func (p *osProcess) monitorMem(memCap execer.Memory, stat stats.StatsReceiver) {
 	reportThresholds := []float64{0, .25, .5, .75, .85, .9, .93, .95, .96, .97, .98, .99, 1}
 	memTicker := time.NewTicker(10 * time.Millisecond)
 	defer memTicker.Stop()
+	// Clean up after ourselves
+	defer cleanupProcs(pid)
 	log.Infof("Monitoring memory for pid=%d", pid)
 	for {
 		select {
 		case <-memTicker.C:
 			p.mutex.Lock()
+			// Process is complete
 			if p.result != nil {
 				p.mutex.Unlock()
 				log.Infof("Finished monitoring memory for pid=%d", pid)
@@ -101,9 +105,15 @@ func (p *osProcess) monitorMem(memCap execer.Memory, stat stats.StatsReceiver) {
 			}
 			mem, _ := memUsage(pid)
 			stat.Gauge("memory").Update(int64(mem))
+			// Aborting process, above memCap
 			if mem >= memCap {
-				msg := fmt.Sprintf("Cmd exceeded MemoryCap, aborting: %d > %d (%v)", mem, memCap, p.cmd.Args)
-				log.Info(msg)
+				msg := fmt.Sprintf("Cmd exceeded MemoryCap, aborting %d: %d > %d (%v)", pid, mem, memCap, p.cmd.Args)
+				log.WithFields(log.Fields{
+					"mem":    mem,
+					"memCap": memCap,
+					"args":   p.cmd.Args,
+					"pid":    pid,
+				}).Info(msg)
 				p.result = &execer.ProcessStatus{
 					State: execer.FAILED,
 					Error: msg,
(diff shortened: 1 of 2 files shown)


runner/execer/os/os_execer.go
```diff
@@ -89,21 +90,35 @@ func (p *osProcess) monitorMem(memCap execer.Memory, stat stats.StatsReceiver) {
...
 				log.Infof("Finished monitoring memory for pid=%d", pid)
 				return
 			}
 			mem, _ := memUsage(pid)
 			stat.Gauge("memory").Update(int64(mem))
+			// Aborting process, above memCap
 			if mem >= memCap {
-				msg := fmt.Sprintf("Cmd exceeded MemoryCap, aborting: %d > %d (%v)", mem, memCap, p.cmd.Args)
-				log.Info(msg)
+				msg := fmt.Sprintf("Cmd exceeded MemoryCap, aborting %d: %d > %d (%v)", pid, mem, memCap, p.cmd.Args)
```
jasonbs10: What does the output look like when doing both sprintf and structured logging?

runner/execer/os/os_execer.go
```diff
@@ -89,21 +90,35 @@ func (p *osProcess) monitorMem(memCap execer.Memory, stat stats.StatsReceiver) {
 	reportThresholds := []float64{0, .25, .5, .75, .85, .9, .93, .95, .96, .97, .98, .99, 1}
 	memTicker := time.NewTicker(10 * time.Millisecond)
 	defer memTicker.Stop()
+	// Clean up after ourselves
```
jasonbs10: This should also be included in monitorMem's description.

runner/execer/os/os_execer.go
```diff
@@ -89,21 +90,35 @@ func (p *osProcess) monitorMem(memCap execer.Memory, stat stats.StatsReceiver) {
...
+	pgid, err := syscall.Getpgid(pid)
```
jasonbs10: I suspect that this should be done at the top where we do "pid:=". Or we should confirm that if pid exits, all of the children processes that share its pgid have exited as well, but that seems unlikely.

runner/execer/os/os_execer.go
```diff
@@ -62,6 +62,7 @@ func (e *osExecer) Exec(command execer.Command) (result execer.Process, err erro
 		cmd.Stderr = stderrW.WriterDelegate()
 	}
 
+	// Sets pgid of all child processes to cmd's pid
```
jasonbs10: I believe this also sets a new sid but you should confirm.

runner/execer/os/os_execer.go
```diff
@@ -172,22 +202,14 @@ func (p *osProcess) Abort() (result execer.ProcessStatus) {
...
-	}
-	if err != nil {
-		result.Error = "Aborted. Parent only, couldn't kill by pgid."
-	}
-	err = p.cmd.Process.Kill()
+
+	err := p.cmd.Process.Kill()
 	if err != nil {
-		result.Error = "Aborted. Couldn't kill pgid or parent."
+		result.Error = "Aborted. Couldn't kill process."
```
jasonbs10: Indicate that we may still attempt cleanup.

runner/execer/os/os_execer.go
```diff
@@ -198,27 +220,26 @@ func (p *osProcess) Abort() (result execer.ProcessStatus) {
...
 func memUsage(pid int) (execer.Memory, error) {
-	// Query for all sets of (pid, ppid, rss) and create a graph. Given a pid, walk the graph accruing memory.
+	// Query for all sets of (pid, sid, rss). Given a pid, find its associated sid.
+	// From there, sum the memory of all processes with the same session id.
 	// Note: look at the revision history for previous approaches which seemed inaccurate at times? ex:
 	//   str := "export P=%d; echo $(ps -orss= -p$(echo -n $(pgrep -g $P | tr '\n' ',')$P) | tr '\n' '+') 0 | bc"
 	str := `
 PID=%d
-PSLIST=$(ps -e -o pid= -o ppid= -o rss= | tr '\n' ';' | sed 's,;$,,')
+PSLIST=$(ps -e -o pid= -o sess= -o rss= | tr '\n' ';' | sed 's,;$,,')
```
jasonbs10: We may want to turn execer.Memory into a struct that allows us to flag how to track memory, i.e. by pgid on OSX and sid in aurora/linux. That or just use an env variable.

Similarly, when running on linux, we can prefer sid over pgid when killing processes.

Currently, it looks like this iterates over all system+user processes on OSX? Hmm how does this pass tests if it excessively overcounts memory - did I only check for growing memory in the tests and not cap the expected memory usage? Can you check?

dgassaway: To add, can we have this function comment clarify if there are differences in accounting between architectures. Suggest removing note about looking in revision history and commented line, this should be working correctly when this is submitted.

runner/execer/os/os_execer.go
```diff
@@ -89,21 +90,35 @@ func (p *osProcess) monitorMem(memCap execer.Memory, stat stats.StatsReceiver) {
 	reportThresholds := []float64{0, .25, .5, .75, .85, .9, .93, .95, .96, .97, .98, .99, 1}
 	memTicker := time.NewTicker(10 * time.Millisecond)
```
dgassaway: This seems too frequent for running a ps and starting a python interpreter and could be a cpu distraction - do we think this wouldn't be as effective if it was 100ms or even 1s?

runner/execer/os/os_execer.go
ryancouto: Nope, just pgid. there's a Setsid: true option too, but I think it's better we don't touch that. That way we can more accurately measure session memory usage. This is from inside a worker in aurora for reference.
screen shot 2017-05-10 at 11 28 53 am

runner/execer/os/os_execer.go
ryancouto: Good call

runner/execer/os/os_execer.go
jasonbs10: This should be increased back to a reasonable value as you suggest once we get this working.

runner/execer/os/os_execer.go
ryancouto: In loglens:
screen shot 2017-05-10 at 11 36 16 am
In the worker:
screen shot 2017-05-10 at 11 38 47 am

runner/execer/os/os_execer.go
ryancouto: I don't know that we want to kill by sid on Linux, that will kill our entire worker. It will restart on a new host, but we still have to deal with the cost of a new startup.

Currently looking into the tests.

runner/execer/os/os_execer.go
ryancouto: Done.

runner/execer/os/os_execer.go
ryancouto: Done.

runner/execer/os/os_execer.go
ryancouto: But that's a good point about tracking by pgid on OSX. I'll make that change.

PR #254, opened 2017-05-11
Jbruno/dx2158job kill 1

JeanetteBruno:
added job kill api


sched/scheduler/stateful_scheduler.go
```diff
@@ -412,3 +434,53 @@ func (s *statefulScheduler) scheduleTasks() {
 			})
 	}
 }
+
+/**
+If the job is pending or in progress kill it.
+*/
+func (s *statefulScheduler) KillJob(jobId string) error {
```
jasonbs10: This allows inProgressJobs to be read and written from different threads, which isn't safe. This has to be done like checkJobs in my changeset, i.e. with a resultCh.

scootapi/client/cli.go
```diff
@@ -3,6 +3,7 @@ package client
 import (
 	"fmt"
 
+	log "github.com/Sirupsen/logrus"
```
jasonbs10: Good call.

scootapi/client/kill_job_cmd.go
```diff
@@ -0,0 +1,65 @@
...
+			return fmt.Errorf("Error getting status: %v", err.Error())
+		}
+	}
+
+	if c.printAsJson {
+		asJson, err := json.Marshal(status)
+		if err != nil {
+			return fmt.Errorf("Error converting status to JSON: %v", err.Error())
+		}
+		log.Infof("%s\n", asJson)
```
jasonbs10: I still don't believe we need both the fmt.Println() and log.Infof() lines but that's a nit. You should definitely change the comments here and in the other commands to remove Sickle OSS references, something like "print valid results to stdout"

scootapi/server/kill_job.go
```diff
@@ -0,0 +1,28 @@
+package server
+
+import (
+	"github.com/scootdev/scoot/sched/scheduler"
+	"github.com/scootdev/scoot/scootapi/gen-go/scoot"
+
+	"fmt"
```
jasonbs10: fmt -> log

runner/execer/execers/pausing_execer.go
```diff
@@ -0,0 +1,73 @@
+package execers
```
JeanetteBruno: removed

sched/definitions.go
```diff
@@ -63,6 +63,9 @@ const (
 	// Successfully Completed Job/Task
 	Completed
 
+	// Killed by request from Client
+	Killed
```
JeanetteBruno: done

sched/scheduler/stateful_scheduler.go
```diff
@@ -242,6 +244,7 @@ func (s *statefulScheduler) step() {
 	// nodes added or removed to cluster, new jobs scheduled,
 	// async functions completed & invoke callbacks
 	s.addJobs()
+	s.killJobs()
```
JeanetteBruno: done

sched/scheduler/stateful_scheduler.go
```diff
@@ -286,11 +289,17 @@ func (s *statefulScheduler) checkForCompletedJobs() {
 
 			// set up variables for async functions for async function & callbacks
 			j := jobState
+			endSagaFn := func() error {
+				return j.Saga.EndSaga()
```
JeanetteBruno: done

sched/scheduler/task_runner.go
```diff
@@ -130,6 +137,13 @@ func (r *taskRunner) runAndWait(taskId string, task sched.TaskDefinition) (runne
 	//TODO(jschiller): add a Nonce to Cmd so worker knows what to do if it sees a dup command?
 	log.Infof("Run() for job:%s taskId:%s", r.jobId, taskId)
 	for {
+		// was a job kill request recieved before we could start the run?
```
ryancouto: *received

sched/scheduler/task_runner.go
```diff
@@ -130,6 +137,13 @@ func (r *taskRunner) runAndWait(taskId string, task sched.TaskDefinition) (runne
...
+		if r.abortRequested() {
+			st = r.getAbortedRunStatus()
+			log.Infof("The run was aborted by scheduler before it was sent to worker: job:%s taskId:%s", r.jobId, taskId)
```
ryancouto: nit: jobId instead of job, and spaces after colons

sched/scheduler/task_runner.go
```diff
@@ -142,12 +156,19 @@ func (r *taskRunner) runAndWait(taskId string, task sched.TaskDefinition) (runne
...
 	}
 	id = st.RunID
+	r.runId = st.RunID
 
 	log.Infof("Query(running) for job:%s, taskId:%s", r.jobId, taskId)
 	// Wait for the process to start running, log it, then wait for it to finish.
 	elapsedRetryDuration = 0
 	includeRunning := true
 	for {
+		// was a job kill request recieved?
```
ryancouto: *received

scootapi/client/cli.go
```diff
@@ -56,6 +60,17 @@ func (c *simpleCLIClient) Init(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	if c.logLevel == "" {
```
ryancouto: not sure this is necessary, if the flag isn't used it should default to info

sched/scheduler/stateful_scheduler.go
JeanetteBruno: done

sched/scheduler/stateful_scheduler.go
```diff
@@ -412,3 +434,53 @@ func (s *statefulScheduler) scheduleTasks() {
...
+			s.killJob(jobId)
+		default:
+			return
+		}
+	}
+
+}
+
+// kill the job
+func (s *statefulScheduler) killJob(jobId string) {
```
JeanetteBruno: done

sched/scheduler/task_runner.go
```diff
@@ -69,7 +72,7 @@ func (r *taskRunner) run() error {
 	taskErr.st = st
 
 	// We got a good message back, but it indicates an error. Update taskErr accordingly.
-	if err == nil && st.State != runner.COMPLETE {
+	if err == nil && st.State != runner.COMPLETE && st.State != runner.ABORTED {
```
JeanetteBruno: done

sched/scheduler/task_runner.go
```diff
@@ -103,7 +106,11 @@ func (r *taskRunner) run() error {
 
 	err = r.logTaskStatus(&taskErr.st, saga.EndTask)
 	taskErr.sagaErr = err
-	if taskErr.sagaErr == nil && taskErr.runnerErr == nil && taskErr.resultErr == nil {
+	if taskErr.st.State == runner.ABORTED {
```
JeanetteBruno: done

sched/scheduler/task_runner.go
JeanetteBruno: done

sched/scheduler/task_runner.go
```diff
@@ -218,3 +240,31 @@ func (r *taskRunner) logTaskStatus(st *runner.RunStatus, msgType saga.SagaMessag
...
+	}
+}
+
+/**
+send an async Abort request to the remote runner.  Construct a RunStatus
+indicating that the run was aborted
+*/
+func (r *taskRunner) Abort() runner.RunStatus {
+	//TODO - use async Runner to not drop the error if Abort() returns an error?
+	go func(runId runner.RunID) {
```
JeanetteBruno: removed

sched/scheduler/task_runner.go
```diff
@@ -218,3 +240,31 @@ func (r *taskRunner) logTaskStatus(st *runner.RunStatus, msgType saga.SagaMessag
...
+		r.runner.Abort(runId)
+	}(r.runId)
+
+	return r.getAbortedRunStatus()
+}
+
+func (r *taskRunner) getAbortedRunStatus() runner.RunStatus {
```
JeanetteBruno: done

scootapi/client/kill_job_cmd.go
```diff
@@ -0,0 +1,65 @@
...
+		Use:   "kill_job",
+		Short: "KillJob",
+	}
+	r.Flags().BoolVar(&c.printAsJson, "json", false, "Print out job status as JSON")
+	return r
+}
+
+func (c *killJobCmd) run(cl *simpleCLIClient, cmd *cobra.Command, args []string) error {
+
+	log.Info("Checking Status for Scoot Job", args)
```
dgassaway: old comment from get job status

scootapi/client/kill_job_cmd.go
JeanetteBruno: done

sched/scheduler/task_runner.go
JeanetteBruno: done

scootapi/server/kill_job.go
JeanetteBruno: removed

scootapi/client/kill_job_cmd.go
dgassaway: Yes.. we should agree on what we want the API contract to be - is it formally returning status to the caller (requires print to stdout), or just logging it as a convenience? Sickle will probably use this in a fire-and-forget manner.

sched/scheduler/task_runner.go
JeanetteBruno: done (throughout the file)

sched/scheduler/task_runner.go
JeanetteBruno: done

scootapi/client/cli.go
JeanetteBruno: removed

scootapi/client/kill_job_cmd.go
JeanetteBruno: done

sched/scheduler/stateful_scheduler.go
```diff
@@ -145,6 +146,7 @@ func NewStatefulScheduler(
 		runnerFactory: rf,
 		asyncRunner:   async.NewRunner(),
 		addJobCh:      make(chan jobAddedMsg, 1),
+		killJobCh:     make(chan string, 1),
```
dgassaway: Dumb question because I'm probably missing something: if addJobCh and killJobCh have a buffer size of 1, and they only get drained once per scheduler loop, what happens if we get more than one request per loop?

jasonbs10: Not dumb, I fixed this in my PR and that should be replicated here.

sched/scheduler/task_runner.go
```diff
@@ -35,6 +35,8 @@ type taskRunner struct {
 	taskId string
 	task   sched.TaskDefinition
 	nodeId cluster.NodeId
+
+	abortCh chan interface{}
```
jasonbs10: Actually, let's make this a "chan bool" to support my use case of node preemption. The meaning will be endSaga=true/false, you'll want to set it to true, and I'll want to set it to false.

sched/scheduler/stateful_scheduler.go
```diff
@@ -355,39 +363,53 @@ func (s *statefulScheduler) scheduleTasks() {
...
+						//TODO -this may results in closed channel panic due to sending endSaga to sagalog (below) before endTask
 						go func() {
-							for err := errors.New(""); err != nil; err = run.logTaskStatus(&taskErr.st, saga.EndTask) {
+							for err := errors.New(""); err != nil; err = tRunner.logTaskStatus(&taskErr.st, saga.EndTask) {
 								time.Sleep(time.Second)
 							}
 							log.Info(msg, jobId, ", task:", taskId, " -> finished goroutine to handle failed saga.EndTask. ")
 						}()
 					}
+					if aborted {
```
jasonbs10: Why is this a separate block from the "if aborted" above? And can you just say "err == nil || aborted" below?

runner/runners/status_manager.go
```diff
@@ -130,6 +130,12 @@ func (s *StatusManager) Query(q runner.Query, wait runner.Wait) ([]runner.RunSta
 		return []runner.RunStatus{st}, s.svcStatus, nil
 	case <-timeout:
 		return nil, s.svcStatus, nil
+	case <-wait.AbortCh:
+		// we just consumed the abort request without really aborting the run,
+		// put a new request on the channel so taskRunner will abort the run.
+		wait.AbortCh <- 1
```
JeanetteBruno: done - and agree its klunky and dangerous- we have to make sure that all Query implementations put the abort back on the channel

sched/scheduler/stateful_scheduler.go
```diff
@@ -355,39 +363,53 @@ func (s *statefulScheduler) scheduleTasks() {
...
 		}
 
+		// mark the task as started in the jobState and record its taskRunner
+		jobState.taskStarted(taskId, tRunner)
+
 		s.asyncRunner.RunAsync(
-			run.run,
+			tRunner.run,
 			func(err error) {
+				log.Infof("In task callback for job: Error:%+v", err)
```
JeanetteBruno: removed

sched/scheduler/stateful_scheduler.go
```diff
@@ -355,39 +363,53 @@ func (s *statefulScheduler) scheduleTasks() {
...
 				if err != nil {
 					// Get the type of error. Currently we only care to distinguish runner (ex: thrift) errors to mark flaky nodes.
 					taskErr := err.(*taskError)
+					st := err.(*taskError).st
+					aborted := (st.State == runner.ABORTED)
 					flaky = (taskErr.runnerErr != nil)
 
 					msg := "Error running job (will be retried):"
-					if taskErr.resultErr != nil && taskErr.st.State == runner.COMPLETE {
-						msg = "Error running job (quitting, tasks that run to completion are not retried):"
```
JeanetteBruno: see task_runner.run() - we only set taskErr.resultErr when state != runner.COMPLETE.   This condition will never be true.

sched/scheduler/stateful_scheduler.go
JeanetteBruno: done

sched/scheduler/stateful_scheduler.go
```diff
@@ -412,3 +434,61 @@ func (s *statefulScheduler) scheduleTasks() {
...
+	var validKillRequests []jobKillRequest
+
+	// validate jobids and sending invalid ids back and building a list of valid ids
+	for haveKillRequest := true; haveKillRequest == true; {
+		select {
+		case req := <-s.killJobCh:
+			// can we find the job?
+			_, ok := s.inProgressJobs[req.jobId]
+			if !ok {
+				req.responeCh <- fmt.Errorf("Job Id %s, not found. "+
```
JeanetteBruno: done

sched/scheduler/stateful_scheduler.go
```diff
@@ -412,3 +434,61 @@ func (s *statefulScheduler) scheduleTasks() {
...
+	for _, req := range validKillRequests {
+		jobState := s.inProgressJobs[req.jobId]
+		for _, task := range jobState.Tasks {
+			if task.Status != sched.Completed {
+				if task.TaskRunner != nil {
+					task.TaskRunner.abortCh <- 1
+				}
+			}
+		}
+
```
JeanetteBruno: added jobKilled back in.

Have to talk with you about closing abortCh ... this is where we trigger the abort action, if we close it then how does taskRunner.abortRequested() ever become true?

sched/scheduler/task_runner.go
JeanetteBruno: done

sched/scheduler/task_runner.go
```diff
@@ -128,11 +130,30 @@ func (r *taskRunner) runAndWait(taskId string, task sched.TaskDefinition) (runne
...
+
+		// was a job kill request received while startingt the run?
+		if r.abortRequested() {
+			if err == nil { // we should have a status with runId, abort the run
+				r.runner.Abort(st.RunID)
+			}
+			st = runner.AbortStatus(id, runner.LogTags{JobID: r.jobId, TaskID: r.taskId})
+			log.Infof("Initial run attempts aborted by scheduler : jobId: %s taskId: %s", r.jobId, taskId)
+			return st, err
+		}
```
JeanetteBruno: done

sched/scheduler/stateful_scheduler.go
jasonbs10: I meant to delete the second part of this comment, ignore it, we actually want to send a value as I asked for in a subsequent comment. In any case, receives from a closed channel return immediately with a default value, and ok=false if you call it like that ex: "foo, ok :=  <-fooChan".

runner/runners/status_manager.go
jasonbs10: We could make two copies of this chan in taskRunner (one for itself and one for query()) and then it could spin up a goroutine that listens to the original chan and then forwards the value to both new ones. Up to you whether it's worth doing this as we won't be adding another query implementation any time soon.

sched/scheduler/stateful_scheduler.go
jasonbs10: It does look like you're right, but can you make 100% sure because I could've sworn we were retrying *all* failed tasks at some point before the introduction of sickle.

sched/scheduler/task_runner.go
```diff
@@ -128,11 +130,30 @@ func (r *taskRunner) runAndWait(taskId string, task sched.TaskDefinition) (runne
...
+		if r.abortRequested() {
+			st = runner.AbortStatus(id, runner.LogTags{JobID: r.jobId, TaskID: r.taskId})
+			log.Infof("The run was aborted by scheduler before it was sent to worker: jobId: %s taskId: %s", r.jobId, taskId)
+			return st, nil
+		}
+
+		// send the command to the worker
 		st, err = r.runner.Run(cmd)
+
+		// was a job kill request received while startingt the run?
```
dgassaway: nit: comment "startingt"

sched/scheduler/stateful_scheduler.go
```diff
@@ -145,6 +153,7 @@ func NewStatefulScheduler(
 		runnerFactory: rf,
 		asyncRunner:   async.NewRunner(),
 		addJobCh:      make(chan jobAddedMsg, 1),
+		killJobCh:     make(chan jobKillRequest, 1000), // TODO - what should this value be?
```
dgassaway: per @jasonbs10's prev comment - leave at 1; fixed in his PR? should probably double check that his fix would include this new ch and not just addJobCh, or merge the fix

sched/scheduler/stateful_scheduler.go
```diff
@@ -355,47 +364,56 @@ func (s *statefulScheduler) scheduleTasks() {
...
-						jobState.errorRunningTask(taskId, err)
+						if preventRetries {
+							msg = fmt.Sprintf("Error running task (quitting, hit max retries of %d):", s.maxRetriesPerTask)
+							err = nil
+						} else {
+							jobState.errorRunningTask(taskId, err)
+						}
 					}
-					log.Info(msg, jobId, ", task:", taskId, " err:", taskErr, " cmd:", taskDef.Argv)
+					log.Info(msg, ", jobId:", jobId, ", taskId:", taskId, " err:", taskErr, " cmd:", taskDef.Argv)
```
ryancouto: nit: Looks like a good candidate for structured logging

sched/scheduler/stateful_scheduler.go
```diff
@@ -355,47 +364,56 @@ func (s *statefulScheduler) scheduleTasks() {
...
 
 					// If the task completed succesfully but sagalog failed, start a goroutine to retry until it succeeds.
 					if taskErr.sagaErr != nil && taskErr.runnerErr == nil && taskErr.resultErr == nil {
-						log.Info(msg, jobId, ", task:", taskId, " -> starting goroutine to handle failed saga.EndTask. ")
+						log.Info(msg, ", jobId:", jobId, ", taskId:", taskId, " -> starting goroutine to handle failed saga.EndTask. ")
```
ryancouto: Same as above

scootapi/client/run_job_cmd.go
```diff
@@ -111,7 +111,7 @@ func (c *runJobCmd) run(cl *simpleCLIClient, cmd *cobra.Command, args []string)
 		}
 	}
 
-	fmt.Println(jobId.ID) // must go to std out so Sickle can pick up the results
+	fmt.Println(jobId.ID) // must go to std out so in case caller looking in stdout for the results
```
ryancouto: nit: remove "so"

scootapi/server/kill_job_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/golang/mock/gomock"
+
+	"github.com/scootdev/scoot/saga"
+	"github.com/scootdev/scoot/sched/scheduler"
+	"github.com/scootdev/scoot/scootapi/gen-go/scoot"
+)
+
+var mockCtrl *gomock.Controller
+
+/**
```
ryancouto: nit: I think the extra * is unnecessary in these block comments. See https://golang.org/doc/effective_go.html#commentary

sched/scheduler/task_runner.go
JeanetteBruno: done

sched/scheduler/stateful_scheduler.go
JeanetteBruno: A task is only marked as failed when the runner cannot run the command or get it's result.  A task whose command ran, but returned a non-zero exit code is marked as completed.

scootapi/client/run_job_cmd.go
JeanetteBruno: done

scootapi/server/kill_job_test.go
JeanetteBruno: done

PR #256, opened 2017-05-22
snapshot - add CLI create bundle command to support manual bundles

dgassaway:
This adds a command to scoot-snapshot-db for creating explicit bundles (where you define the basis and ref to be bundled).  The bundle is then Stored in the bundlestore setup by scoot-snapshot-db (i.e. it gets uploaded). This is built around the edges of the existing interfaces as the Snapshot mechanics do not handle non-single-rev snapshots, and existing snapshot bundle creation is embedded within that logic. This is required for creation and storage of source-funnel history bundles.

Quick summary of new command:
`scoot-snapshot-db create publish_git_bundle --basis="abd54d5dc85dcf1c78a0dd1e915213ac9475cf83" --ref=master --ttl="336h"`
Prints stored bundle location to stdout e.g.: "http:/localhost:9094/bundle/bs-abd54d5dc85dcf1c78a0dd1e915213ac9475cf83-master.bundle"


snapshot/cli/cli.go
```diff
@@ -136,6 +142,94 @@ func (c *ingestGitCommitCommand) run(db snapshot.DB, _ *cobra.Command, _ []strin
...
+	basis string
+	ref   string
+	ttld  time.Duration
+	keep  bool
+}
+
+func (c *createGitBundleCommand) register() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create_git_bundle",
+		Short: "Creates a git bundle file as specified by basis and reference for the cwd repo",
```
ryancouto: It creates the git bundle *and* writes it to the snapshot DB, right?

dgassaway: Yeah - not apparent from the name.

snapshot/cli/cli.go
```diff
@@ -136,6 +142,94 @@ func (c *ingestGitCommitCommand) run(db snapshot.DB, _ *cobra.Command, _ []strin
...
+	cmd.Flags().BoolVar(&c.keep, "keepFile", false, "Keep created bundles file")
+	return cmd
+}
+
+func (c *createGitBundleCommand) run(db snapshot.DB, _ *cobra.Command, _ []string) error {
+	if c.basis == "" || c.ref == "" {
+		return fmt.Errorf("Create bundle command requires a basis and reference")
+	}
+
+	wd, err := os.Getwd()
```
dgassaway: Yes - next block calls repo.NewRepository and errors out if it is not

snapshot/git/gitdb/bundlestore.go
```diff
@@ -273,3 +273,22 @@ func (s *bundlestoreSnapshot) downloadBundle(db *DB) (filename string, err error
 func makeBundleName(key string) string {
 	return fmt.Sprintf("bs-%s.bundle", key)
 }
+
+func (b *bundlestoreBackend) storeFile(filePath string, ttl *bundlestore.TTLValue) (string, error) {
```
jasonbs10: Can you make uploadLocalSnapshots call this? Also I think you can just name this uploadFile to keep naming consistent. Same for db.go below.

snapshot/cli/cli.go
jasonbs10: Might as well add that this uploads, and do the same for ingestGitCommitCommand.

snapshot/cli/cli.go
```diff
@@ -136,6 +142,94 @@ func (c *ingestGitCommitCommand) run(db snapshot.DB, _ *cobra.Command, _ []strin
...
+
+func (c *createGitBundleCommand) register() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create_git_bundle",
+		Short: "Creates a git bundle file as specified by basis and reference for the cwd repo",
+	}
+	// See https://git-scm.com/docs/git-bundle for basis, ref usage
+	cmd.Flags().StringVar(&c.file, "file", "", "Output bundle file (defaults to temp file)")
+	cmd.Flags().StringVar(&c.basis, "basis", "", "Basis for bundle")
+	cmd.Flags().StringVar(&c.ref, "ref", "", "Reference to be packaged")
```
jasonbs10: default to master?

snapshot/cli/cli.go
```diff
@@ -136,6 +142,94 @@ func (c *ingestGitCommitCommand) run(db snapshot.DB, _ *cobra.Command, _ []strin
...
+// Snapshot, ID, Stream schemas.
+type createGitBundleCommand struct {
+	file  string
+	basis string
+	ref   string
+	ttld  time.Duration
+	keep  bool
+}
+
+func (c *createGitBundleCommand) register() *cobra.Command {
```
jasonbs10: Copy your PR comment about how to call this (print it too if that's just as easy).

snapshot/cli/cli.go
```diff
@@ -136,6 +142,94 @@ func (c *ingestGitCommitCommand) run(db snapshot.DB, _ *cobra.Command, _ []strin
...
+	bundleFilename, tempDir := "", ""
+	if c.file != "" {
+		bundleFilename = c.file
+	} else {
+		td, err := temp.TempDirDefault()
+		if err != nil {
+			return fmt.Errorf("Couldn't create temp dir: %v", err)
+		}
+		tempDir = td.Dir
+		bundleFilename = path.Join(tempDir, fmt.Sprintf("bs-%v-%v.bundle", c.basis, c.ref))
```
jasonbs10: I'm not following why we want a separate bundle name format. It shouldn't ever conflict with any other bundles we can generate and now it's more work to ingest it (which remains to be done).

snapshot/cli/cli.go
dgassaway: sure

snapshot/cli/cli.go
dgassaway: yes, considering renaming to "publish_git_bundle" - wdyt?

snapshot/cli/cli.go
dgassaway: makes sense

snapshot/cli/cli.go
dgassaway: If someone ran scoot-snapshot-db on an old commit it would create bs-<rev>.bundle which the funnel could also create, and they would collide (and be fundamentally different, I think). Is it because we don't ever expect users to be creating snapshots on existing revs?

On the bundle contents, the existing bundles created with this tool are odd; they bundle a special scoot ref "reserved_scoot/bundlestore/__temp_for_writing" and appear to have a full history from the command line although they do not. Basically the new bundles are not easily cross-compatible with our existing ones if I understand them correctly (and I may not).

Agree though that the expanded name check is ugly. If we can determine there's no danger in using the old naming let's remove it.

jasonbs10: So <rev> in this case is a fingerprint for the bundle you just created, which should result in a unique bs-<rev> we can store. The handle for this bundle needs to adhere to the "bs-fs-bundleShaFingerprint-sm-checkoutSha" format though (I think that's it, I'm never 100% sure)

dgassaway: The long name is for a snapshot ID, which determines some backend settings that do not apply to this bundle (and the resulting file always gets just bs-\<rev\>, hence the existing check). Really the existing filename scheme doesn't accommodate anything new, unless we allow the 'long' names to be uploaded - even then I'm not sure it really fits. TBH these bundles don't really belong here.  If we don't want to extend the name we could generate a new non-commit sha for the bundles.

jasonbs10: I'm not quite following. We can do whatever, but based on my understanding, this PR is incomplete without the ability to ingest new snapshotID formats on the worker.

dgassaway:
Updated - bundle names created with the new command are now "bs-\<rev\>.bundle" where rev is a new sha1 created based on the --basis and --ref, which will not conflict with commit shas or require changing the name-checker (that change is reverted).
Added a cleanup for tmpdirs that were being created by each call to scoot-snapshot-db.

This is only the creation-side change for these bundles; ingestion by workers will be a separate PR.

snapshot/cli/cli.go
```diff
@@ -195,7 +197,11 @@ func (c *createGitBundleCommand) run(db snapshot.DB, _ *cobra.Command, _ []strin
 			return fmt.Errorf("Couldn't create temp dir: %v", err)
 		}
 		tempDir = td.Dir
-		bundleFilename = path.Join(tempDir, fmt.Sprintf("bs-%v-%v.bundle", c.basis, c.ref))
+
+		// Don't use commit sha in case of collision with other bundle,
+		// create simple sha derived from basis and ref
+		sha := sha1.Sum([]byte(revList))
```
jasonbs10: I see now, I was thinking about Stdout/Stderr snapshots which call write-tree which results in a unique sha every time. Obviously we can't do the same thing for repo diff snapshots since that would mess with our history. So yeah, this is a good workaround.

PR #257, opened 2017-05-22
Jbruno/dx2041 obs metrics

JeanetteBruno:
updating code to record stats as per https://docs.google.com/spreadsheets/d/15AMKnyV3xWkm4i4228sbCfD0ORH61AB6YO5filMZbFs/edit#gid=1404908761.  Also adding unit tests for the stats.

I would like to merge this PR.  There will be additional metrics added, but would like to get these in now and will add other metrics in smaller chunks.


tests/testhelpers/StatsVerify.go
```diff
@@ -0,0 +1,97 @@
+package testhelpers
```
ryancouto: statsVerify.go -> stats_verify.go

tests/testhelpers/StatsVerify.go
```diff
@@ -0,0 +1,97 @@
...
+type Rule struct {
+	Checker func(interface{}, interface{}) bool
+	Value interface{}
+}
+
+/*
+Verify that the stats registry object contains values for the keys in the contains map parameter and that
+each entry conforms to the rule (condition) associated with that key.
+ */
+func VerifyStats(statsRegistry stats.StatsRegistry, t *testing.T, contains map[string]Rule) {
```
ryancouto: if this was moved to common/stats/stats_test.go, I *think* we could keep finagleStatsRegistry private.

binaries/scheduler/config/config.go
```diff
@@ -1,5 +1,6 @@
 // Code generated by go-bindata.
 // sources:
+// config/.DS_Store
```
ryancouto: remove .DS_Store from dir and re-run

binaries/workerserver/config/config.go
```diff
@@ -1,5 +1,6 @@
 // Code generated by go-bindata.
 // sources:
+// config/.DS_Store
```
ryancouto: remove .DS_Store from dir and re-run

tests/testhelpers/StatsVerify.go
jasonbs10: The idea behind testhelpers was to create a neutral directory that can be reused from multiple locations/packages. I believe common/stats is a good place for this if using the same criteria, so I agree with Ryan's assessment.

sched/generators.go
```diff
@@ -84,7 +84,7 @@ func GenTask() TaskDefinition {
 func GenRandomTask(rng *rand.Rand) TaskDefinition {
 	snapshotId := fmt.Sprintf("snapShotId:%s", testhelpers.GenRandomAlphaNumericString(rng))
 	taskId := fmt.Sprintf("taskId:%s", testhelpers.GenRandomAlphaNumericString(rng))
-	numArgs := rng.Intn(5)
+	numArgs := rng.Intn(5) + 1 // must have at least 1 arg
```
jasonbs10: I might be wrong about this particular instance, but generally we want to test -all- combinations of inputs, including invalid ones.

snapshot/bundlestore/groupcache.go
```diff
@@ -73,8 +73,8 @@ func toPeers(nodes []cluster.Node, stat stats.StatsReceiver) []string {
 		peers = append(peers, "http://"+string(node.Id()))
 	}
 	log.Info("New groupcacheStore peers: ", peers)
-	stat.Counter("peerDiscoveryCounter").Inc(1)
-	stat.Gauge("peerCountGauge").Update(int64(len(peers)))
+	stat.Counter("peerDiscoveryCounter").Inc(1)		// TODO errata metric - remove if unused
+	stat.Gauge("peerCountGauge").Update(int64(len(peers)))	// TODO errata metric - remove if unused
```
jasonbs10: We're casting too wide a net for errata if we're not making super important stats like this first-class. I was thinking that errata would be the somewhat redundant or trivial stuff and that we'd have a separate section for standard metrics that were useful but weren't absolutely required as key indicators.

tests/testhelpers/StatsVerify.go
JeanetteBruno: goes away when move logic into stats_test.go

snapshot/bundlestore/groupcache.go
JeanetteBruno: k

tests/testhelpers/StatsVerify.go
JeanetteBruno: done

binaries/scheduler/config/config.go
JeanetteBruno: ok

binaries/workerserver/config/config.go
JeanetteBruno: ok

sched/generators.go
JeanetteBruno: done

common/stats/verify_stats.go
```diff
@@ -0,0 +1,108 @@
...
+	msg.WriteString(tag)
+	msg.WriteString(":stats registry error:\n")
+
+	if ok {
+		asJson := asFinagleRegistry.MarshalAll()
+
+		for key, rule := range contains {
+			gotValue, ok := asJson[key]
+			if !ok {
+				if !strings.Contains(runtime.FuncForPC(reflect.ValueOf(rule.Checker).Pointer()).Name(), "DoesNotExist") {
```
jasonbs10: It's generally better if we can avoid reflection - can you try assigning DoesNotExist and the other checkers to their own variables and then operate on pointers to those variables? That way you can do plain equality.

sched/scheduler/stateful_scheduler.go
```diff
@@ -321,6 +316,20 @@ func (s *statefulScheduler) step() {
 	s.checkForCompletedJobs()
 	s.killJobs()
 	s.scheduleTasks()
+
+	remaining := 0
+	waitingToStart := 0
+	for _, job := range s.inProgressJobs {
+		remaining += (len(job.Tasks) - job.TasksCompleted)
+		log.Infof("***jobid:%s, completed:%d, running:%d", job.Job.Id, job.TasksCompleted, job.TasksRunning)
```
jasonbs10: I suspect this is too verbose and I believe this information is already printed after doing scheduleTasks()

sched/scheduler/task_scheduler_test.go
```diff
@@ -288,7 +310,7 @@ func Test_TaskAssignments_PriorityStages(t *testing.T) {
 		nodes = append(nodes, fmt.Sprintf("node%d", i))
 	}
 	testCluster := makeTestCluster(nodes...)
-	cs := newClusterState(testCluster.nodes, testCluster.ch, nil)
+	cs := newClusterState(testCluster.nodes, testCluster.ch, nil, statsReceiver)
```
jasonbs10: You could have a utility function newFinagleStats() or similar, placed inline here and delete lines 275-276. This is a small amount of boilerplate that nonetheless appears a lot and introduces a lot of noise.

scootapi/server/server_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	_, err = handler.KillJob("testJobId")
+	if err != nil {
+		t.Errorf("GetStatus returned err:%s", err.Error())
+	}
+
+	stats.VerifyStats("", statsRegistry, t,
+		map[string]stats.Rule{
+			"runJobRpmCounter":        {Checker: stats.Int64EqTest, Value: 1},
+			"runJobLatency_ms.avg":    {Checker: stats.FloatGTTest, Value: 0.0},
```
jasonbs10: Is there an easy way to test latency as well?

scootapi/server/server_test.go
```diff
@@ -0,0 +1,86 @@
...
+			"jobStatusRpmCounter":     {Checker: stats.Int64EqTest, Value: 1},
+			"jobStatusLatency_ms.avg": {Checker: stats.FloatGTTest, Value: 0.0},
+			"jobKillRpmCounter":       {Checker: stats.Int64EqTest, Value: 1},
+			"jobKillLatency_ms.avg":   {Checker: stats.FloatGTTest, Value: 0.0},
+		})
+}
+
+// converts a scheduler Job into a Thrift Job
+func schedJobDefToScootJobDef(schedJobDef *sched.JobDefinition) (*scoot.JobDefinition, error) {
```
jasonbs10: This already appears in scootapi/client/run_job_cmd.go. Better to pull that out and share it via scootapi/api.go (similar to how workerapi does it). All scootapi domain/thrift  conversion should go in there really.

common/stats/verify_stats.go
JeanetteBruno: done

sched/scheduler/stateful_scheduler.go
JeanetteBruno: took out (if you see *** pattern,  that generally means its a line I used during debugging and forgot to take it out)

sched/scheduler/task_scheduler_test.go
JeanetteBruno: switched over to use NilStatsReceiver - was able to eliminate the added lines

scootapi/server/server_test.go
JeanetteBruno: not sure what you mean.... this is testing that the latency value was created and it got some value.  We can't be more precise on the value because of variable runtimes.  (we could ballpark it ... > 1000, but then the test may start failing if it ends up on a super fast machine or we modify the logic so the action runs faster).  
Note: this test is using apriori knowledge that the test stat structure create a bunch of field (avg, p50, p...) for each named latency measurement.  I'm using the avg field to validate that a value got into latency - didn't see the need to validate the histogram implementation within latency.

scootapi/server/server_test.go
JeanetteBruno: Looks like it, but it really is a different structure.  I've added a comment regarding this.

PR #258, opened 2017-05-25
Creating constants for the stats names. 

JeanetteBruno:
Creating one location for  finding the list of stats being collected.  Let me know if you want me to merge this PR and create one more PR with the rest of the changes, or merge the rest of the changes into this PR.

### DIFF
--- a/common/stats/stats_names.go
+++ b/common/stats/stats_names.go
@@ -1,0 +1,81 @@
+package stats
+
+/*
+This file defines all the metrics being collected.   As new metrics are added please follow this pattern.
+ */
+
+const (
+	/*** Node availability (ClusterManger) metrics ***/
+	/*
+	the number of worker nodes that are available or running tasks (not suspended)
+	 */
+	ClusterAvailableNodes = "availableNodes"
+
+	/*
+	the number of idle worker nodes (available, but not running a task)
+	 */
+	ClusterIdleNodes ="idleNodes"
+
+	/*
+	the number of lost worker nodes (not responding to status requests)
+	 */
+	ClusterLostNodes ="lostNodes"
+
+
+	/*** Scheduler Metrics ***/
+	/*
+	The number of jobs in the inProgress list at the end of each time through the
+	scheduler's job handling loop
+	 */
+
+	SchedAcceptedJobsGauge="schedAcceptedJobsGauge"
+
+	/*
+	the number of tasks that have finished (including those that have been killed)
+	 */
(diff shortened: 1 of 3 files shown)


common/stats/stats_names.go
```diff
@@ -0,0 +1,81 @@
...
+	The number of jobs in the inProgress list at the end of each time through the
+	scheduler's job handling loop
+	 */
+
+	SchedAcceptedJobsGauge="schedAcceptedJobsGauge"
+
+	/*
+	the number of tasks that have finished (including those that have been killed)
+	 */
+	SchedCompletedTaskCounter = "completedTaskCounter"
```
jasonbs10: Do we want to remove the 'sched' prefix from all of these strings? It should already have 'scheduler' and whatever reason I had to include 'sched' prefix may no longer apply.

common/stats/stats_names.go
```diff
@@ -0,0 +1,81 @@
...
+	*/
+	SchedInProgressTasksGauge="schedInProgressTasksGauge"
+
+	/*
+	the number of job requests that have been put on the addJobChannel
+	 */
+	SchedJobsCounter="schedJobsCounter"
+
+	/* TODO - remove? (discuss usefulness)
+	the amount of time it takes to verify a job definition, it to the add job channel and
```
jasonbs10: missing word

common/stats/stats_names.go
```diff
@@ -0,0 +1,81 @@
...
+	/*
+	the number of times the platform retried sending an end saga message
+	 */
+	SchedRetriedEndSagaCounter= "schedRetriedEndSagaCounter"
+
+	/*
+	The number of jobs waiting to start in the inProgress list at the end of each time through the
+	scheduler's job handling loop.  (No tasks in this job have been started.)
+	 */
+	SchedWaitingJobsGauge="schedWaitingJobsGauge"
```
dgassaway: Nit about inconsistent spacing in these var assignments, unless gofmt did that

common/stats/stats_names.go
JeanetteBruno: done

common/stats/stats_names.go
JeanetteBruno: I would like to leave the strings exactly as is right now.  Then as I update the graphs, I can update the strings.

common/stats/stats_names.go
JeanetteBruno: no - I hadn't run fmt yet - will do so with the bulk changes.

PR #260, opened 2017-05-26
Jbruno/dx2186 new metrics - worker download/upload

JeanetteBruno:
added metrics to track download/upload latency and counts from the worker - updated worker creation functions to allow passing in the stats collector.

updated the tests to use the constants instead of hard-coded names (forgot to make these changes in the last PR)

### DIFF
--- a/sched/worker/workers/makers.go
+++ b/sched/worker/workers/makers.go
@@ -19,7 +19,7 @@ func MakeInmemoryWorker(node cluster.Node, tmp *temp.TempDir) runner.Service {
 // Makes a worker suitable for using as an in-memory worker.
 func MakeDoneWorker(tmp *temp.TempDir) runner.Service {
 	ex := execers.NewDoneExecer()
-	r := runners.NewSingleRunner(ex, snapshots.MakeInvalidFiler(), nil, runners.NewNullOutputCreator(), tmp)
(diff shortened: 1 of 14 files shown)


sched/worker/workers/makers.go
```diff
@@ -19,7 +19,7 @@ func MakeInmemoryWorker(node cluster.Node, tmp *temp.TempDir) runner.Service {
 // Makes a worker suitable for using as an in-memory worker.
 func MakeDoneWorker(tmp *temp.TempDir) runner.Service {
 	ex := execers.NewDoneExecer()
-	r := runners.NewSingleRunner(ex, snapshots.MakeInvalidFiler(), nil, runners.NewNullOutputCreator(), tmp)
+	r := runners.NewSingleRunner(ex, snapshots.MakeInvalidFiler(), nil, runners.NewNullOutputCreator(), tmp, nil)
```
jasonbs10: just curious - is it better for clarity if we use nilStatsReceiver everywhere instead of nil, or is this usage fairly inconsequential?

JeanetteBruno: I think its inconsequential - we still need to check for nil in downstream uses in case we add/update code and nil sneaks back in.

PR #262, opened 2017-05-30
Add support for periodic repo updates

dgassaway:
This adds a mechanism for updating underlying repos.  Updates are synchronized between task runs, which are handled by the Controller.  The QueueController is extended to manage updates between runs.

Controllers are exposed to repos via Filers, so an Updater interface has been added to Filer, and wired down through gitdb as RepoUpdater (paired with existing RepoIniter).

In-practice implementation for closed source is added in https://phabricator.twitter.biz/D57159 (for now, golang impl only)

### DIFF
--- a/binaries/scoot-snapshot-db/main.go
+++ b/binaries/scoot-snapshot-db/main.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/scootdev/scoot/common/dialer"
 	"github.com/scootdev/scoot/common/log/hooks"
+	"github.com/scootdev/scoot/common/stats"
 	"github.com/scootdev/scoot/os/temp"
 	"github.com/scootdev/scoot/scootapi"
 	"github.com/scootdev/scoot/snapshot"
@@ -86,8 +87,12 @@ func (i *injector) Inject() (snapshot.DB, error) {
 	}
 
 	store := bundlestore.MakeHTTPStore(url)
-	return gitdb.MakeDBFromRepo(dataRepo, nil, tempDir, nil, nil, &gitdb.BundlestoreConfig{Store: store},
-		gitdb.AutoUploadBundlestore), nil
+	return gitdb.MakeDBFromRepo(
+			dataRepo, nil, tempDir, nil, nil,
+			&gitdb.BundlestoreConfig{Store: store},
+			gitdb.AutoUploadBundlestore,
+			stats.NilStatsReceiver()),
+		nil
 }
 
 func removeTemp() {
--- a/binaries/scoot-snapshot-db/main.go
+++ b/binaries/scoot-snapshot-db/main.go
@@ -86,7 +86,7 @@ func (i *injector) Inject() (snapshot.DB, error) {
 	}
 
 	store := bundlestore.MakeHTTPStore(url)
-	return gitdb.MakeDBFromRepo(dataRepo, tempDir, nil, nil, &gitdb.BundlestoreConfig{Store: store},
+	return gitdb.MakeDBFromRepo(dataRepo, nil, tempDir, nil, nil, &gitdb.BundlestoreConfig{Store: store},
 		gitdb.AutoUploadBundlestore), nil
 }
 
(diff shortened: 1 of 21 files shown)


snapshot/filer.go
```diff
@@ -46,6 +50,18 @@ type Ingester interface {
 	IngestMap(srcToDest map[string]string) (id string, err error)
 }
 
+const NoDuration time.Duration = time.Duration(0) * time.Second
```
jasonbs10: nit - don't think you need time.Second.

snapshot/git/gitdb/setup.go
```diff
@@ -57,3 +62,10 @@ func (i *NewRepoIniter) Init() (*repo.Repository, error) {
 
 	return repo.InitRepo(repoTmp.Dir)
 }
+
+// Noop Repo Updater
+type NewRepoUpdater struct{}
```
jasonbs10: nit - name it NoopRepoUpdater?

Also is there a way to make the distinction between ResourceUpdater and RepoUpdater clearer? I don't have any suggestions but they're easy to confuse as-is.

snapshot/git/gitdb/setup.go
```diff
@@ -21,6 +23,9 @@ func (m module) Install(b *ice.MagicBag) {
 		func(tmp *temp.TempDir) RepoIniter {
 			return &NewRepoIniter{tmp: tmp}
 		},
+		func() RepoUpdater {
+			return &NewRepoUpdater{}
```
jasonbs10: Can you add a github comment outlining the plan for the real impl? It'll be golang calling into a python script?

runner/runners/queue.go
```diff
@@ -60,13 +73,24 @@ func NewSingleRunner(
 // QueueController maintains a queue of commands to run (up to capacity).
```
JeanetteBruno: add comment about the new behavior - updating repo

runner/runners/queue.go
```diff
@@ -60,13 +73,24 @@ func NewSingleRunner(
...
+	filer         snapshot.Filer
 	statusManager *StatusManager
 	capacity      int
 
 	mu           sync.Mutex
 	queue        []cmdAndID
 	runningID    runner.RunID
 	runningAbort chan<- struct{}
+
+	startCh  chan cmdAndID
```
JeanetteBruno: add comments: 
startCh is used to signal/request starting a command
updateCh is used to signal/request updating the repo

snapshot/git/gitfiler/checkouter.go
```diff
@@ -82,9 +83,11 @@ func (c *Checkouter) CheckoutAt(id string, dir string) (co snapshot.Checkout, er
 	return &UnmanagedCheckout{id: id, dir: dir}, nil
 }
 
-// Implement noop ingest so this Checkouter can be passed around as a Filer.
+// Implement noop ingest/update so this Checkouter can be passed around as a Filer.
```
JeanetteBruno: Since we're editing this code, it would be nice to have a comment with the Checkouter type declaration explaining that Checkouter is a Filer that only implements checkout from a pool of repos (managed by RepoPool).

binaries/scoot-snapshot-db/main.go
```diff
@@ -86,7 +86,7 @@ func (i *injector) Inject() (snapshot.DB, error) {
 	}
 
 	store := bundlestore.MakeHTTPStore(url)
-	return gitdb.MakeDBFromRepo(dataRepo, tempDir, nil, nil, &gitdb.BundlestoreConfig{Store: store},
+	return gitdb.MakeDBFromRepo(dataRepo, nil, tempDir, nil, nil, &gitdb.BundlestoreConfig{Store: store},
```
dgassaway: Agree, it would be nice to have. Going to put a comment in gitdb and defer.

snapshot/filer.go
dgassaway: removed, thanks

snapshot/git/gitdb/setup.go
dgassaway: will add to summary

snapshot/git/gitdb/setup.go
dgassaway: updating this, should make more sense. Also removed ResourceUpdater in snapshot.DB and just reused Updater from snapshot/filer.

PR #263, opened 2017-06-02
Jschiller/tryout uncommitted

jasonbs10:
Updated scootapi to handle new scheduler strategy parameters. TODO: scheduler has a killjob() panic.

### DIFF
--- a/scootapi/scoot.thrift
+++ b/scootapi/scoot.thrift
@@ -49,11 +49,13 @@ struct TaskDefinition {
   1: required Command command,
   2: optional string snapshotId,
   3: optional string taskId,
+  4: optional i32 timeoutMs,
 }
 
 struct JobDefinition {
   1: required map<string, TaskDefinition> tasks,
   2: optional JobType jobType,
+  3: optional i32 defaultTaskTimeoutMs,
 }
 
 struct JobId {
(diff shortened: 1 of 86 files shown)


scootapi/scoot.thrift
```diff
@@ -55,7 +55,7 @@ struct TaskDefinition {
 
 struct JobDefinition {
   1: required list<TaskDefinition> tasks,
-  2: optional JobType jobType,
+  2: optional JobType DEPRECATED_jobType,
```
JeanetteBruno: will we be removing this? (I'm assuming we're leaving it in to limit the scope of this change)

jasonbs10: We can remove this after everything has been redeployed (but leave something to indicate that field '2:' is no longer available.)

scootapi/server/run_job.go
```diff
@@ -65,8 +65,17 @@ func thriftJobToScoot(def *scoot.JobDefinition) (result sched.JobDefinition, err
...
+	}
+	if def.Basis != nil {
+		result.Basis = *def.Basis
+	}
 	if def.JobType != nil {
-		result.JobType = def.JobType.String()
+		result.JobType = *def.JobType
+	}
+	if def.Priority != nil {
+		result.Priority = sched.Priority(*def.Priority)
```
dgassaway: Do we need one for requestor?

jasonbs10: Yep, thanks.

PR #264, opened 2017-06-02
Queue runner init, scoot-snapshot-db minor fixes

dgassaway:


runner/runners/queue.go
```diff
@@ -62,13 +62,11 @@ func NewQueueRunner(
 				statusManager.UpdateService(runner.ServiceStatus{Initialized: false, Error: err})
 			} else {
 				statusManager.UpdateService(runner.ServiceStatus{Initialized: true})
+				go controller.loop()
```
jasonbs10: I believe you'll need to revert this based on my latest merge.

snapshot/cli/cli.go
```diff
@@ -229,9 +231,17 @@ func (c *createGitBundleCommand) run(db snapshot.DB, _ *cobra.Command, _ []strin
 		}
 		tempDir = td.Dir
 
+		// Get the length of this revList for the sha we will compute for file name
+		// basis and ref alone do not guarantee uniqueness
+		revData, err := ingestRepo.Run("rev-list", revList)
+		if err != nil {
+			return fmt.Errorf("couldn't get rev-list for %s: %v", revList, err)
+		}
+		length := len(strings.Split(strings.TrimSpace(string(revData[:])), "\n"))
```
jasonbs10: nit - does this need the "[:]"?

snapshot/cli/cli.go
```diff
@@ -229,9 +231,17 @@ func (c *createGitBundleCommand) run(db snapshot.DB, _ *cobra.Command, _ []strin
 		}
 		tempDir = td.Dir
 
+		// Get the length of this revList for the sha we will compute for file name
```
jasonbs10: I don't quite follow why this is the chosen solution. Can we get a sha based on just the revList? Or can you explain a little more.

dgassaway: We need some extra 'entropy' because our given revList as <rev>..<ref> which is in practice "<rev>..master" is not unique - there could be different numbers of things inbetween <rev> and master.  In fact this was causing collisions with funnel bundles where a TIP bundle created with a recent <rev> eventually became the basis rev for the AUX bundle, and resulted in the same generated sha being used which caused some init failures.

snapshot/cli/cli.go
dgassaway: removed

jasonbs10:
Sure, but why not just use the revlist bytes directly? I'm still missing something if that wouldn't work and be clearer.

runner/runners/queue.go
dgassaway: I don't think this changes - assumption is we can't start the loop unless we were initialized successfully, which needs logic in 2 different places depending on whether we had an initDoneCh

dgassaway:
Gotcha - makes sense (we still need to run the rev-list cmd here, but don't need the Sprintf)

runner/runners/queue.go
jasonbs10: The latest merge has all requests being serviced within loop(). We need to run loop so we can reject run requests (even if they don't happen in practice due to new node init checking).

dgassaway: Ok - doing a workaround for updater

runner/runners/queue.go
```diff
@@ -109,14 +110,20 @@ type QueueController struct {
...
-func makeSimpleTicker(d time.Duration) <-chan time.Time {
-	if d != snapshot.NoDuration {
-		return time.NewTicker(d).C
+// Start a goroutine that forwards update ticks to updateCh, so that we can skip if e.g. init failed
+func startUpdateTicker(d time.Duration, u chan<- struct{}) {
+	if d == snapshot.NoDuration || u == nil {
+		return
 	}
-	return nil
+	ticker := time.NewTicker(d).C
```
jasonbs10: nit - typically want to have a mechanism to stop timers/goroutines, but we can leave this as-is.

dgassaway: agreed, I left it out because there's no close/stop implementation for queue controller/runner in general, but will add small comment

PR #266, opened 2017-06-07
Addded tracking initialization latency.  The metrics cover the elapse…

JeanetteBruno:
The metrics cover the elapsed time for an ongoing initialization (to allow identifying workers that are stuck in an 'initializing' state) and the elapsed time for completed initializations (to allow identifying if we need to optimize the initialization phase).  Implementation: a channel for the init done is passed to queue.  When init is done, the time is passed back on this channel.  Handler's stats() goroutine listens for the init done time.   Also made some minor code restructuring, comments and message changes.


runner/runners/queue.go
```diff
@@ -52,10 +52,24 @@ type cmdAndID struct {
...
+The init chan is optional and may be nil if the filer has no initializion step.
+@param: idtCh - channel that queue uses to report (to handler) the time the initialization finished
+@param: output
+@param: tmp
+@param: capacity - the maximum number of commands to support on the queue.  If 0 then the queue is unbounded.
+@param: stats - the stats receiver the queue will use when reporting its metrics
+*/
 func NewQueueRunner(
-	exec execer.Execer, filer snapshot.Filer, idc snapshot.InitDoneCh, output runner.OutputCreator, tmp *temp.TempDir, capacity int, stat stats.StatsReceiver) runner.Service {
+	exec execer.Execer, filer snapshot.Filer, idc snapshot.InitDoneCh, idtCh snapshot.InitDoneTimeCh, output runner.OutputCreator, tmp *temp.TempDir, capacity int, stat stats.StatsReceiver) runner.Service {
```
jasonbs10: I don't believe this extra channel is necessary, and the parameter list is already clunky. You should be able to query the service for a rough approximation of when init is done. You could do this in workerapi/server/server.go:stats().

workerapi/server/server.go
```diff
@@ -31,63 +32,83 @@ func MakeServer(
...
 //TODO: runner should eventually be extended to support stats, multiple runs, etc. (replacing loop here).
-func (h *handler) stats() {
-	var startTime time.Time
+func (h *handler) stats(idtCh snapshot.InitDoneTimeCh) {
+	var startTime time.Time = time.Now()
+	var initTime time.Duration
 	nilTime := time.Time{}
-	ticker := time.NewTicker(time.Millisecond * time.Duration(500))
+	initDoneTime := nilTime
+	ticker := time.NewTicker(time.Millisecond * time.Duration(h.statsCollectInterval))
```
jasonbs10: You should define statsCollectInterval as a time.Duration and convert from int->Duration when we define the value main.go.

ice/eval.go
```diff
@@ -56,9 +56,9 @@ type stack []frame
 func (s stack) String() string {
 	b := new(bytes.Buffer)
 	log.Info(b)
-	log.Info(b, "goice stacktrace:")
+	log.Info(b, "goice stacktrace (constructor chain):")
```
JeanetteBruno: It doesn't appear to be getting used.... I'm changing it to LogStack()

runner/runners/queue.go
JeanetteBruno: k

workerapi/server/server.go
JeanetteBruno: k

common/stats/stats_names.go
```diff
@@ -376,6 +376,16 @@ const (
...
 
 	/*
+		The amount of time it took a worker to init
+	*/
+	WorkerFinalInitLatency_ms = "workerFinishedInitLatency_ms"
+
+	/*
+		The number of workers who are currently exceeding the max init time
+	*/
+	WorkerOngoingInitLatency_ms = "workerActiveInitLatency_ms"
```
JeanetteBruno: changed to match names

PR #267, opened 2017-06-08
bundlestore: added http retries. Also added Controller.Release().

jasonbs10:
Should be a quick 'lgtm', as usual I'll add tests before landing.

### DIFF
--- a/runner/runners/queue.go
+++ b/runner/runners/queue.go
@@ -79,6 +79,7 @@ func NewQueueRunner(
 		capacity:      capacity,
 		reqCh:         make(chan interface{}),
 		updateCh:      make(chan struct{}),
+		cancelTimerCh: make(chan interface{}, 1),
 	}
 	run := &Service{controller, statusManager, statusManager}
 
@@ -93,12 +94,12 @@ func NewQueueRunner(
 				statusManager.UpdateService(runner.ServiceStatus{Initialized: false, Error: err})
 			} else {
 				statusManager.UpdateService(runner.ServiceStatus{Initialized: true})
-				startUpdateTicker(filer.UpdateInterval(), controller.updateCh)
+				startUpdateTicker(filer.UpdateInterval(), controller.updateCh, controller.cancelTimerCh)
 			}
 		}()
 	} else {
 		statusManager.UpdateService(runner.ServiceStatus{Initialized: true})
-		startUpdateTicker(filer.UpdateInterval(), controller.updateCh)
+		startUpdateTicker(filer.UpdateInterval(), controller.updateCh, controller.cancelTimerCh)
 	}
 
 	go controller.loop()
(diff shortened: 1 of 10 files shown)


snapshot/bundlestore/http_store.go
```diff
@@ -10,19 +10,25 @@ import (
...
 )
 
 func MakeHTTPStore(rootURI string) Store {
 	if !strings.HasSuffix(rootURI, "/") {
 		rootURI = rootURI + "/"
 	}
-	client := &http.Client{Timeout: 30 * time.Second}
+	client := pester.New()
+	client.Backoff = pester.ExponentialBackoff
+	client.MaxRetries = 8
```
ryancouto: I prefer constants to magic numbers, otherwise lgtm

jasonbs10: I think this is functionally the same given that it's an assignment with a clear name. If there were a larger number of options it might make more sense as a global def.

JeanetteBruno: I agree with Ryan.  Putting these sorts of constants (ones that impact performance) in a common section, (typically at the top of a file),  makes it easy to identify where we may want to tune the system.  We don't have to externalize them immediately.

runner/runners/queue.go
```diff
@@ -129,20 +130,27 @@ type QueueController struct {
...
-			u <- struct{}{}
+	ticker := time.NewTicker(d)
+	go func() {
+	Loop:
+		for {
+			select {
+			case <-ticker.C:
+				u <- struct{}{}
+			case <-cancelCh:
+				break Loop
```
JeanetteBruno: add ticker.Stop() above break Loop?

runner/runners/queue.go
```diff
@@ -205,6 +213,11 @@ func (c *QueueController) abort(run runner.RunID) (runner.RunStatus, error) {
 	return status, err
 }
 
+// Cancels goroutines and exits run loop.
```
JeanetteBruno: which goroutine(s) does it cancel?

JeanetteBruno:
I'm a bit confused on the Release() method... who is calling it when? and I see it being called from statefulScheduler, but don't see it in the Thrift interface.

snapshot/bundlestore/http_store.go
jasonbs10: Ok.

runner/runners/queue.go
jasonbs10: Yep, thanks.

runner/runners/queue.go
jasonbs10: All the goroutines created in this file.

jasonbs10:
Release is called by stateful_scheduler as you said. This is about releasing runner resources that are local to the process, and we leave remote alone, which is fine since we aren't actually promising an equivalence between Service and Worker. I'll clarify that these are process-local resources.

PR #272, opened 2017-06-15
bundlestore and scheduler reporting uptime

JeanetteBruno:
Will need to merge these changes with pr266

### DIFF
--- a/binaries/apiserver/main.go
+++ b/binaries/apiserver/main.go
@@ -60,6 +60,7 @@ func main() {
 	bag.InstallModule(endpoints.Module())
 	bag.PutMany(
 		func() endpoints.StatScope { return "apiserver" },
+		func() stats.UpTimeReportIntvl { return stats.UpTimeReportIntvl(500 * time.Millisecond)},
 		func() endpoints.Addr { return endpoints.Addr(*httpAddr) },
 		func(bs *bundlestore.Server, vs *snapshots.ViewServer, sh *StoreAndHandler) map[string]http.Handler {
 			return map[string]http.Handler{
@@ -71,7 +72,7 @@ func main() {
 				sh.endpoint: sh.handler,
 			}
 		},
-		func(fileStore *bundlestore.FileStore, stat stats.StatsReceiver, tmp *temp.TempDir) (*StoreAndHandler, error) {
+		func(fileStore *bundlestore.FileStore, stat stats.StatsReceiver, upTimeReportIntv stats.UpTimeReportIntvl, tmp *temp.TempDir) (*StoreAndHandler, error) {
 			cfg := &bundlestore.GroupcacheConfig{
 				Name:         "apiserver",
 				Memory_bytes: 2 * 1024 * 1024 * 1024, //2GB
@@ -79,7 +80,7 @@ func main() {
 				Endpoint:     "/groupcache",
 				Cluster:      createCluster(),
 			}
-			store, handler, err := bundlestore.MakeGroupcacheStore(fileStore, cfg, stat)
+			store, handler, err := bundlestore.MakeGroupcacheStore(fileStore, cfg, stat, upTimeReportIntv)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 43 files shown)


common/stats/stats.go
```diff
@@ -30,6 +30,8 @@ import (
 // For testing.
 var Time StatsTime = DefaultStatsTime()
 
+type UpTimeReportIntvl time.Duration
```
dgassaway: Seems odd to make a type for this when it looks like all uses are converting directly from or directly to a time.Duration

scootapi/server/server_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/scootdev/scoot/sched"
 	"github.com/scootdev/scoot/sched/scheduler"
 	"github.com/scootdev/scoot/scootapi/gen-go/scoot"
+	"time"
```
ryancouto: nit: move this up to same import block as testing

scootapi/server/server_test.go
```diff
@@ -50,6 +53,8 @@ func Test_RequestCounters(t *testing.T) {
 		t.Errorf("GetStatus returned err:%s", err.Error())
 	}
 
+	time.Sleep(120 * time.Millisecond)
```
ryancouto: why 120?

common/stats/stats.go
JeanetteBruno: making it its own type since it is injected by ice.  I'm thinking ice injected variables should be distinct types in case we need multiple variables of the same common type (though I'm not completely sure that this will prevent ice from confusing the multiple parameters)

scootapi/server/server_test.go
JeanetteBruno: done

binaries/apiserver/main.go
```diff
@@ -71,15 +72,15 @@ func main() {
...
+		func(fileStore *bundlestore.FileStore, stat stats.StatsReceiver, upTimeReportIntv stats.UpTimeReportIntvl, tmp *temp.TempDir) (*StoreAndHandler, error) {
 			cfg := &bundlestore.GroupcacheConfig{
 				Name:         "apiserver",
 				Memory_bytes: 2 * 1024 * 1024 * 1024, //2GB
 				AddrSelf:     *httpAddr,
 				Endpoint:     "/groupcache",
 				Cluster:      createCluster(),
 			}
-			store, handler, err := bundlestore.MakeGroupcacheStore(fileStore, cfg, stat)
+			store, handler, err := bundlestore.MakeGroupcacheStore(fileStore, cfg, stat, upTimeReportIntv)
```
jasonbs10: This would be cleaner if you could package the interval inside of the stat object. Also, consistent naming would be better wrt Intv vs Intvl

JeanetteBruno: this a a collection interval for a specific set of metrics.  Don't think this belongs in the stat object - different metrics will probably end up with different collection intervals. 

Will update names to Intvl

jasonbs10: And if a component needs to take multiple different intervals? I'd strongly prefer to keep stats confined as much as possible and given the *current* code I think it's ok to move it to stat and perhaps make the name more unique.

JeanetteBruno: k

common/stats/stats.go
```diff
@@ -507,3 +509,17 @@ func (r *finagleStatsRegistry) marshalHistogram(
 
 var defaultPercentiles = []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9999}
 var defaultPercentileLabels = []string{"p50", "p90", "p95", "p99", "p999", "p9999"}
+
+func StartUptimeReporting(stat StatsReceiver, statName string, upReportIntvl UpTimeReportIntvl) {
+	startTime := time.Now()
+	ticker := time.NewTicker(time.Duration(upReportIntvl))
+	for {
```
dgassaway: Can we just do a single loop here, "for _ := range ticker.C {..." ?

JeanetteBruno: good catch - didn't notice it!

PR #274, opened 2017-06-18
Added client timeout as sanity check. Fix for scheduler available node count

jasonbs10:


config/scootconfig/workers_modules.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/scootdev/scoot/workerapi/client"
 )
 
+type ClientTimeout time.Duration
+
```
ryancouto: would prefer to have a DefaultClientTimeout = time.Minute here to make it easily configurable across the board

PR #277, opened 2017-06-21
snapshot/cli - add refname to rev-list for publish git bundle

dgassaway:
See code comment for explanation, and dx-2245


JeanetteBruno:
test?

dgassaway:
Updated link ticket with more detail on the example seen.  Testing is manual right now, adding a comment that the business portion should be moved to its own package.  Externally this would benefit from an integration/smoketest that exercised the whole infrastructure

PR #287, opened 2017-07-27
Defensively abort worker task if it didn't return COMPLETE.

jasonbs10:
This fixes the issue where scheduler would exit the run step with in error (namely BADREQUEST) and then continue to send jobs to the same worker which was stuck running the previous command. This isn't comprehensive but Abort() is the only mechanism we currently have to clean up a worker.


dgassaway:
What case is this supposed to be fixing again?

PR #288, opened 2017-08-01
Jbruno/task latency

JeanetteBruno:
added metrics for 
- tracking task time (from worker) (to be able to create a alerts on many timeouts)
- tracking when a scheduler, worker and apiservices are starting (to create alerts/warnings when they are being restarted)
- explicitly tracking bundlestore upload/download errors (to be able to avoid having the bundlestore error dashboard go negative)


runner/runners/invoke.go
```diff
@@ -61,6 +61,7 @@ func (inv *Invoker) Run(cmd *runner.Command, id runner.RunID) (abortCh chan<- st
 func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struct{}, updateCh chan runner.RunStatus) (r runner.RunStatus) {
 	log.Infof("run. id: %v", id)
 	defer func() {
+		inv.stat.Latency(stats.WorkerTaskLatency_ms).Time().Stop()
```
dgassaway: Not sure what this is measuring - seems like we are immediately starting & stopping the timer?

workerapi/server/server.go
```diff
@@ -144,9 +144,9 @@ func (h *handler) QueryWorker() (*worker.WorkerStatus, error) {
 
 // Implements worker.thrift Worker.Run interface
 func (h *handler) Run(cmd *worker.RunCommand) (*worker.RunStatus, error) {
-	defer h.stat.Latency(stats.WorkerServerRunLatency_ms).Time().Stop() // TODO errata metric - remove if unused
+	defer h.stat.Latency(stats.WorkerServerStartRunLatency_ms).Time().Stop()
```
dgassaway: Same as above

runner/runners/invoke.go
JeanetteBruno: this should track how long it takes to do the checkout + run + record stdout/stderr actions.

workerapi/server/server.go
JeanetteBruno: This was already there - it basically reports how long it takes to get a run id (start the task).  I renamed it to include 'Start' to (try to) make that more clear.

runner/runners/invoke.go
dgassaway: I mean, I see other examples of timers where we call Time() first, and Stop() later - if we are immediately initializing the timer and calling stop in the defer func, aren't all of these measurements basically zero?

dgassaway: Hmm. I think the other examples work because they're the argument passed to defer, which get evaluated immediately before the defer function is executed - in this one the defer func has empty args and may not initialize the Time() until this function is already running?

common/stats/stats.go
```diff
@@ -521,3 +522,11 @@ func StartUptimeReporting(stat StatsReceiver, statName string) {
 		}
 	}
 }
+
+func ReportServerRestart(stat StatsReceiver, statName string) {
+	stat.Gauge(statName).Update(int64(1))
+	go func(stat StatsReceiver, statName string) {
+		time.Sleep(1 *time.Minute)
```
jasonbs10: Would it be safer to make this two minutes? Also this should be configurable since 1 minute doesn't mean much in the OSS world.

runner/runners/invoke.go
```diff
@@ -60,10 +60,12 @@ func (inv *Invoker) Run(cmd *runner.Command, id runner.RunID) (abortCh chan<- st
...
 func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struct{}, updateCh chan runner.RunStatus) (r runner.RunStatus) {
 	log.Infof("run. id: %v", id)
-	defer func() {
+	taskTimer := inv.stat.Latency(stats.WorkerTaskLatency_ms).Time()
+	defer func(taskTimer stats.Latency) {
+		taskTimer.Stop()
 		updateCh <- r
 		close(updateCh)
-	}()
+	}(taskTimer)
```
jasonbs10: Curious, I see this from time to time, what's the purpose of passing this in vs referring to the closure'd var directly?

runner/runners/single_test.go
```diff
@@ -130,33 +130,76 @@ func TestMemCap(t *testing.T) {
...
+		t.Fatal("stats check did not pass.")
+	}
+}
+
+func TestTimeout(t *testing.T) {
+	log.AddHook(hooks.NewContextHook())
+	logrusLevel, _ := log.ParseLevel("debug")
+	log.SetLevel(logrusLevel)
+
+	statsReg := stats.NewFinagleStatsRegistry()
```
jasonbs10: This is a lot of boilerplate, please refactor if you think it's worthwhile.

runner/runners/invoke.go
JeanetteBruno: just to make use of external parameters more visible.  Changed to use closure'd var directly.

runner/runners/single_test.go
JeanetteBruno: done

common/stats/stats.go
JeanetteBruno: good idea

PR #294, opened 2017-08-04
runner/runners: Write to stdout/err when a task times out or is aborted

ryancouto:


runner/runners/invoke.go
```diff
@@ -171,9 +171,13 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 	// Wait for process to complete (or cancel if we're told to)
 	select {
 	case <-abortCh:
+		stdout.Write([]byte(fmt.Sprintf("%s\n\nTask aborted: %v", marker, cmd.String())))
+		stderr.Write([]byte(fmt.Sprintf("$s\n\nTask aborted: %v", marker, cmd.String())))
```
jasonbs10: $s

Also, if it isn't a pain, can you paste what the output looks like / does it catch the attention enough?

ryancouto: yep - gotta do a dev deploy but will post in a bit

runner/runners/invoke.go
```diff
@@ -171,9 +171,13 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 	// Wait for process to complete (or cancel if we're told to)
 	select {
 	case <-abortCh:
+		stdout.Write([]byte(fmt.Sprintf("%s\n\nTask aborted: %v", marker, cmd.String())))
```
JeanetteBruno: This will not include the current contents of stdout/stderr from the running command will it?  If not, can we modify this (and timeout below) to include the current stdout/stderr?

ryancouto:
screen shot 2017-08-04 at 1 30 36 pm

ryancouto:
screen shot 2017-08-04 at 3 55 04 pm

dgassaway:
I'd throw in an "ERROR" or "FAILED" or some obvious loud keyword in there, gets kind of lost in the long command

PR #297, opened 2017-08-12
Stop streaming logs if the underlying resource has changed.

jasonbs10:
This was an easy change but I don't have the time to test it now, so this placeholder can be ignored atm.

From hipchat:
"I believe the issue from this morning is that we didn't print the timeout in the streaming log file and then eventually, after a worker restart, the runIds built back up to 222 and we started updating our old run output output output output with the new run"


runner/runners/local_output.go
```diff
@@ -101,14 +102,23 @@ func (s *localOutputCreator) ServeHTTP(w http.ResponseWriter, r *http.Request) {
     xhr.onreadystatechange = function () {
       var DONE=4, OK=200;
       if (xhr.readyState === DONE && xhr.status == OK) {
+        var id = xhr.getResponseHeader('X-Resource-Id')
+        var txt = xhr.responseText.substring(prevLength);
+        prevLength = xhr.responseText.length
+        if resourceId == ''
+          resourceId = id
+        if id != resourceId {
+          txt = Date() + ': Underlying resource changed! Quitting.'
```
JeanetteBruno: recommend give more info in this response and put marker in to highlight it.... 

*************************
Underlying resource changed - quitting streaming.
This is typically caused by the worker restarting.  
The file being streamed was: <path to original file>, use ssh to retrieve/view the file.
**************************

PR #298, opened 2017-08-20
Schedule long-running tasks first, and fix BADREQUEST handling

jasonbs10:
Scheduling tasks by historical run duration is a good-enough optimization until we have clients that can do efficient DAG-based scheduling.

Also fixes mishandling of BADREQUEST after a run request from scheduler -> worker.

### DIFF
--- a/runner/runners/invoke.go
+++ b/runner/runners/invoke.go
@@ -202,19 +202,33 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 		errPath := stderr.AsFile()
 		stdoutName := "STDOUT"
 		stderrName := "STDERR"
-		if writer, err := os.Create(filepath.Join(tmp.Dir, stdoutName)); err != nil {
-			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stdout: %v", err), runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
-		} else if reader, err := os.Open(outPath); err != nil {
-			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stdout: %v", err), runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
+		var writer *os.File
+		var reader *os.File
+		defer writer.Close()
+		defer reader.Close()
+
+		if writer, err = os.Create(filepath.Join(tmp.Dir, stdoutName)); err != nil {
+			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stdout: %v", err),
+				runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
+		} else if reader, err = os.Open(outPath); err != nil {
+			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stdout: %v", err),
+				runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
 		} else if _, err := io.Copy(writer, reader); err != nil {
-			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stdout: %v", err), runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
+			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stdout: %v", err),
+				runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
 		}
-		if writer, err := os.Create(filepath.Join(tmp.Dir, stderrName)); err != nil {
-			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stderr: %v", err), runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
-		} else if reader, err := os.Open(errPath); err != nil {
-			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stderr: %v", err), runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
+
+		writer.Close()
+		reader.Close()
+		if writer, err = os.Create(filepath.Join(tmp.Dir, stderrName)); err != nil {
+			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stderr: %v", err),
+				runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
+		} else if reader, err = os.Open(errPath); err != nil {
+			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stderr: %v", err),
+				runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
 		} else if _, err := io.Copy(writer, reader); err != nil {
-			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stderr: %v", err), runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
+			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stderr: %v", err),
+				runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
 		}
 
 		ingestCh := make(chan interface{})
--- a/runner/runners/invoke.go
+++ b/runner/runners/invoke.go
@@ -119,7 +119,7 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 			downloadTimer.Stop()
 		}
 		if err != nil {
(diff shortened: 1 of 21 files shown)


sched/scheduler/cluster_state.go
```diff
@@ -302,10 +303,14 @@ func (c *clusterState) update(updates []cluster.NodeUpdate) {
 			}
 		} else if ns.timeFlaky != nilTime && now.Sub(ns.timeFlaky) > c.maxFlakyDuration {
 			// This flaky node has been suspended long enough, try adding it back to the healthy node pool.
-			delete(c.suspendedNodes, ns.node.Id())
-			c.nodes[ns.node.Id()] = ns
+			// TimeFlaky should've been the only time* value set at this point, reset it.
```
dgassaway: So we don't have to delete from c.suspendedNodes any more? Or add back into c.nodes?  And no idea why startReadyLoop has to be added.

sched/scheduler/stateful_scheduler.go
```diff
@@ -414,9 +427,12 @@ addLoop:
 		select {
 		case newJobMsg := <-s.addJobCh:
 			receivedJob = true
-			js := newJobState(newJobMsg.job, newJobMsg.saga)
+			js := newJobState(newJobMsg.job, newJobMsg.saga, s.taskDurations)
 			s.inProgressJobs = append(s.inProgressJobs, js)
 
+			// TODO(jschiller): associate related tasks, i.e. task retries that decorate the taskId?
```
dgassaway: Do we have any way of measuring how long tasks are waiting to run or any indicator that will show improvements?

scootapi/locate.go
```diff
@@ -17,7 +17,8 @@ import (
 
 // Get the path of the file containing the address for scootapi to use
 func GetScootapiAddrPath() string {
-	return path.Join(os.Getenv("HOME"), ".cloudscootaddr")
+	optionalId := os.Getenv("SCOOT_ID") // Used to connect to a different set of scoot processes.
```
dgassaway: ?????

sched/scheduler/cluster_state.go
jasonbs10: I'll add to the comment. The intention is to reapply the existing logic to get a node from new -> healthy state. That means aborting anything that's running (happens in readyFn) and then adding the node back as if it's newly init'd.

sched/scheduler/stateful_scheduler.go
jasonbs10: I planned to rely on our existing metrics to see if sq latency improves and continue to manually review jobs that are taking longer than is typical. I'm open to other suggestions.

scootapi/locate.go
jasonbs10: This is a knob enabling devs to run concurrent tests against different cloudExec instances. I don't see a need to make this more prominent. Would it suffice to add a comment "Allows devs to concurrently test different scoot instances"?

dgassaway: Preferably add context to the explanation comment at the top of the file

runner/status.go
```diff
@@ -26,12 +26,12 @@ const (
 	ABORTED
 	// Operation timed out and was killed.
 	TIMEDOUT
-	// Invalid or error'd request. Original runner state not affected. Retry may work after mutation.
+	// Request rejected for reasons unrelated to the user.
```
JeanetteBruno: I'm not sure we have clear semantics on FAILED vs BADREQUEST...
if BADREQUEST is unrelated to user (task definition), then is FAILED related to user (task definition)?
FAILED = invalid command (os execer hits an error from Start(), invalid snapshot (checkout cannot find snapshot id)
BADREQUEST = the runner code hits an error that is other than listed for FAILED.....
The current code appears to be reporting FAILED from invoke.go for:
  . checkout returns error
  . creating stdout directory
  . creating stderr directory
  . execer.exec() returns an error
  . create tmp invoke dir
  . ingesting stdout/stderr
  . unexpected state from execer
(this doesn't align with the definition above)

We need to clarify the definitions (update the comments) and verify that the code matches.

runner/runners/invoke.go
```diff
@@ -248,9 +248,8 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 		}
 		return status
 	case execer.FAILED:
-		// We treat failure here as a bad request since it's more likely a faulty cmd than something we did wrong.
-		return runner.BadRequestStatus(id, fmt.Errorf("error execing: %v", st.Error), runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
+		return runner.ErrorStatus(id, fmt.Errorf("error execing: %v", st.Error), runner.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID})
```
JeanetteBruno: Can we change runner.ErrorStatus() to runner.FailedStatus() to match execer.FAILED

runner/status.go
jasonbs10: Great comment. I wasn't particularly satisfied with my description. I would say that we can break things down into two categories - transient failures due to resource issues, and failures due to inconsistent state / internal logic.

Do the current errors match those categories in your opinion? Are those useful error categories or do you think we can add more state or just boil it all down to FAILED?

sched/scheduler/cluster_state.go
```diff
@@ -84,6 +84,7 @@ func (ns *nodeState) ready() bool {
 
 // Stars a goroutine loop checking node readiness, waiting 'backoff' time between checks, and exiting if node is fully removed.
```
JeanetteBruno: typo: Stars -> Starts

sched/scheduler/cluster_state.go
```diff
@@ -84,6 +84,7 @@ func (ns *nodeState) ready() bool {
...
 func (ns *nodeState) startReadyLoop(rfn ReadyFn) {
+	ns.readyCh = make(chan interface{})
```
JeanetteBruno: the channel is created when ns is created in newNodeState() - remove from here or from newNodeState()?

sched/scheduler/job_state.go
```diff
@@ -45,13 +59,18 @@ func newJobState(job *sched.Job, saga *saga.Saga) *jobState {
 	}
 
 	for _, taskDef := range job.Def.Tasks {
+		duration := taskDurations[taskDef.TaskID].duration
```
JeanetteBruno: if taskDurations is optional (as per comment above),  do we risk null error?

scootapi/locate.go
JeanetteBruno: ditto ?????? - if the explanation is put in the dev docs, then this code may be easier to understand.

scootapi/scoot.thrift
```diff
@@ -21,10 +21,10 @@ enum RunStatusState {
   PENDING = 1      # Run scheduled but not yet started.
   RUNNING = 2      # Run is happening.
   COMPLETE = 3     # Succeeded or failed yielding an exit code. Only state with an exit code.
-  FAILED = 4       # Run mechanism failed and is no longer running. Retry may or may not work.
+  FAILED = 4       # Run mechanism failed and is no longer running.
```
JeanetteBruno: could use clearer definitions of FAILED vs BADREQUEST (see prior comment)

workerapi/worker.thrift
```diff
@@ -3,10 +3,10 @@ enum Status {
   PENDING = 1      # Run scheduled but not yet started.
   RUNNING = 2      # Run is happening.
   COMPLETE = 3     # Succeeded or failed yielding an exit code. Only state with an exit code.
-  FAILED = 4       # Run mechanism failed and is no longer running. Retry may or may not work.
+  FAILED = 4       # Run mechanism failed and is no longer running.
```
JeanetteBruno: see prior comments about definitions of FAILED vs BADREQUEST

sched/scheduler/stateful_scheduler.go
JeanetteBruno: There is a lot going on in sq latency... I too was thinking we should measure how long tasks are waiting to start.

runner/status.go
dgassaway: I think we should remove "un/related to the user," as that just causes confusion.  BADREQUEST is a specific type of failure that prevented receipt of a request (it looks like we use this if we can't convert from thrift to domain object, which makes sense).  I don't have a problem with the categories as-is but the comment was misleading.

JeanetteBruno: I had always thought BADREQUEST was 'bad task description: invalid os command, invalid snapshot id, etc' and the FAILED was 'an error other than BADREQUEST: typically due to missing support service, bug in the platform code, etc.  example: error creating directories used by the runner, error communicating with bundlestore, etc.'  (Note: so my understanding/description of FAILED vs BADREQUEST appears to be opposite of what was implemented... I can adjust my understanding.)

PR #299, opened 2017-08-28
Increase min number of nodes for higher priority jobs.

jasonbs10:
Increase min number of nodes for higher priority jobs.
Allow new related jobs to jump the queue.
Clearer scheduler config defaults.

### DIFF
--- a/sched/scheduler/stateful_scheduler.go
+++ b/sched/scheduler/stateful_scheduler.go
@@ -44,19 +44,22 @@ const DefaultDefaultTaskTimeout = 30 * time.Minute
 const DefaultTaskTimeoutOverhead = 15 * time.Second
 
 // Number of different requestors that can run jobs at any given time.
-const DefaultMaxRequestors = 100
+const DefaultMaxRequestors = 10
 
-// Number of jobs any single requestor can have.
+// Number of jobs any single requestor can have (to prevent spamming, not for scheduler fairness).
 const DefaultMaxJobsPerRequestor = 100
 
 // Expected number of nodes. Should roughly correspond with the actual number of healthy nodes.
-const DefaultNumConfiguredNodes = 100
+const DefaultNumConfiguredNodes = 20
+
+// The maximum number of nodes required to run a 'large' job in an acceptable amount of time.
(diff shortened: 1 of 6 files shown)


sched/scheduler/stateful_scheduler.go
```diff
@@ -44,19 +44,22 @@ const DefaultDefaultTaskTimeout = 30 * time.Minute
...
 
 // Number of different requestors that can run jobs at any given time.
-const DefaultMaxRequestors = 100
+const DefaultMaxRequestors = 10
 
-// Number of jobs any single requestor can have.
+// Number of jobs any single requestor can have (to prevent spamming, not for scheduler fairness).
 const DefaultMaxJobsPerRequestor = 100
 
 // Expected number of nodes. Should roughly correspond with the actual number of healthy nodes.
```
JeanetteBruno: I don't understand this comment - there are thousands of nodes (hopefully most are healthy)... is this number of healthy nodes for a job?

sched/scheduler/stateful_scheduler.go
```diff
@@ -44,19 +44,22 @@ const DefaultDefaultTaskTimeout = 30 * time.Minute
 const DefaultTaskTimeoutOverhead = 15 * time.Second
 
 // Number of different requestors that can run jobs at any given time.
-const DefaultMaxRequestors = 100
+const DefaultMaxRequestors = 10
```
dgassaway: Reasoning behind this?

jasonbs10: This lines up better with CloudExec size during dev testing. I'll add a comment indicating that these are example numbers for a relatively small cluster.

sched/scheduler/stateful_scheduler.go
jasonbs10: I don't like having too many knobs and this one is here primarily to address cluster size during worker redeploys. Scheduling will be suboptimal during this period but it's likely fine to get rid of this and use the actual cluster size directly.

ryancouto:
+1 on updating high level readme. recommend using README.md over internal docs.

PR #300, opened 2017-08-28
scoot/runners - make StatusManager and Queue messages Info

dgassaway:
This is a "feeler" diff for an adjustment to worker logs. The motivations are:
1. make more useful information about tasks available via loglens by promoting them to 'info' level.
2. Get rid of noise in worker by switching over to info level (this will be a closed-source diff).
By promoting these logs to info, the main thing we would ditch from debug level on the worker are the ps output messages from os/os_execer.go's logs, that make it very hard to follow the worker log.

On a worker that's been up ~4 days, the logs promoted here amounted to 500 or so per day average, which I don't think will be an issue for log lens.

Note: if we do this, planning on leaving devel & staging at debug level.


dgassaway:
Amounts to 1 or 2 extra log lines per task run, which shouldn't break the kps rating (if adding a line or two puts us over the edge that'd be a good thing to fix anyway)

PR #301, opened 2017-08-30
Disabled priority>1 since it's not tested.

jasonbs10:
We were seeing a panic in task_scheduler kill logic. This disables untested priorities >1 but keeps the job insertion order logic, albeit in a hacky way that should eventually be replaced by a protocol change.

### DIFF
--- a/sched/scheduler/job_state.go
+++ b/sched/scheduler/job_state.go
@@ -11,13 +11,14 @@ import (
 // Contains all the information for a job in progress
 // Note: Only Job, Saga, and Tasks are provided during scheduler recovery. Anything else must be initialized separately.
 type jobState struct {
-	Job            *sched.Job
-	Saga           *saga.Saga   //saga associated with this job
-	Tasks          []*taskState //ordered list of taskState
-	EndingSaga     bool         //denotes whether an EndSagaMsg is in progress or not
-	TasksCompleted int          //number of tasks that've been marked completed so far.
-	TasksRunning   int          //number of tasks that've been scheduled or started.
-	JobKilled      bool         //indicates the job was killed
+	Job               *sched.Job
+	Saga              *saga.Saga     //saga associated with this job
+	Tasks             []*taskState   //ordered list of taskState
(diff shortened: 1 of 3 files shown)


sched/scheduler/job_state.go
```diff
@@ -49,15 +50,19 @@ func (s taskStatesByDuration) Less(i, j int) bool {
...
+		Tasks:             make([]*taskState, 0),
+		EndingSaga:        false,
+		TasksCompleted:    0,
+		TasksRunning:      0,
+		JobKilled:         false,
+		InsertionPriority: job.Def.Priority, // Shoehorn in priority to determine insertion behavior in task_scheduler.go
 	}
 
+	// Job Priorities higher than 1 are currently disabled as per stateful_scheduler.go (excepting InsertionPriority).
+	job.Def.Priority = sched.Priority(min(int(sched.P1), int(job.Def.Priority)))
```
JeanetteBruno: this is introducing a side effect on the job structure, recommend we override job.Def.Priority in thriftJobToScoot() in run_job.go

jasonbs10: I can't do that until we introduce a new field for Job InsertionOrder, which I'd prefer to defer for now.

sched/scheduler/stateful_scheduler.go
```diff
@@ -55,7 +55,7 @@ const DefaultMaxJobsPerRequestor = 100
 const DefaultSoftMaxSchedulableTasks = 1000
 
 // Increase the NodeScaleFactor by a percentage defined by 1 + (Priority * NodeScaleAdjustment)
-var NodeScaleAdjustment = .5
+var NodeScaleAdjustment = 1.0
```
dgassaway: i still don't grok why this gets changed around

jasonbs10: We've been running with P1 for long enough to be confident about it imo (at least a week I believe). This adjustment is elaborated on a bit more in GetNodeScaleFactor().

In practice this change means that a P1 job will get at least 2x the nodes of a P0 jobs, rather than 1.5x.

PR #302, opened 2017-08-31
Removed DeadLetterExitCode in preference for an err msg.

jasonbs10:
DeadLetterExitCode was breaking the contract specified in runner/status.go

### DIFF
--- a/sched/scheduler/task_runner.go
+++ b/sched/scheduler/task_runner.go
@@ -13,7 +13,7 @@ import (
 	"github.com/scootdev/scoot/workerapi"
 )
 
-const DeadLetterExitCode = -200
+const DeadLetterTrailer = " -> Canceling task, exceeded max tries."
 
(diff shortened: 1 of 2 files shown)


sched/scheduler/task_runner.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/scootdev/scoot/workerapi"
 )
 
-const DeadLetterExitCode = -200
+const DeadLetterTrailer = " -> Canceling task, exceeded max tries."
```
JeanetteBruno: 'Canceling task, exceeding max tries.'  will be confusing with the Sickle log statements about 'Retries'....
stateful_scheduler.go also has 3 lines (610, 612, and 616) that put something about retrying a task in the error statement.  

I do not have an alternative recommendation... except that since max.retriesPerTask is currently set to 0, don't put anything in the messages about retrying: 'Error encountered, cancelling task.'

PR #303, opened 2017-09-07
DX-2086: Allow requestor to track a specific job through the logs

ryancouto:
Also modified a lot of the logging to use structured fields as described in ticket


ryancouto:
Great suggestions, thx @JeanetteBruno ! I'll pick up the doc writing for our next sprint.

I think it's worth adding requestorTag to execer.Command to allow for further tracing. Will push that out in a bit.

runner/execer/os/os_execer.go
```diff
@@ -70,7 +71,7 @@ func (e *osExecer) Exec(command execer.Command) (result execer.Process, err erro
 		return nil, err
 	}
 
-	proc := &osProcess{cmd: cmd}
+	proc := &osProcess{cmd: cmd, LogTags: tags.LogTags{RequestorTag: command.RequestorTag}}
```
jasonbs10: If you used a LogTags field instead of embedding it in Command, you could just do "LogTags: command.LogTags". (I'd try to differentiate the naming a bit more between the LogTags type and the LogTags fields).

Either way, we should make sure to keep the LogTags entries together where possible rather than split it out like here.

sched/definitions.go
```diff
@@ -181,18 +178,20 @@ func makeThriftJobFromDomainJob(domainJob *Job) (*schedthrift.Job, error) {
...
-		Tag:       &(domainJob).Def.Tag,
-		Basis:     &(domainJob).Def.Basis,
-		Requestor: &(domainJob).Def.Requestor,
+		JobType:      &(*domainJob).Def.JobType,
+		Tasks:        thriftTasks,
+		Priority:     &prio,
+		Tag:          &(domainJob).Def.Tag,
+		Basis:        &(domainJob).Def.Basis,
+		Requestor:    &(domainJob).Def.Requestor,
+		RequestorTag: &(domainJob).Def.RequestorTag,
```
jasonbs10: Tag, Requestor, and RequestorTag?

See scootapi/scoot.thrift for tag and requestor descriptions - I don't think we're supposed to have a third RequestorTag entry.

scootapi/client/run_job_cmd.go
```diff
@@ -42,17 +45,18 @@ type CLIJobDef struct {
...
-	JobID      string
-	TaskID     string
 	TimeoutMs  int32
+	tags.LogTags
 }
 
 func (c *runJobCmd) run(cl *simpleCLIClient, cmd *cobra.Command, args []string) error {
 	log.Info("Running on scoot, args:", args)
 	jobDef := scoot.NewJobDefinition()
+	jobDef.RequestorTag = &c.RequestorTag
```
jasonbs10: What's happening with the rest of the log tags here?

runner/runners/invoke.go
```diff
@@ -59,7 +60,13 @@ func (inv *Invoker) Run(cmd *runner.Command, id runner.RunID) (abortCh chan<- st
 // Run will not return until the process is not running.
 // NOTE: kind of gnarly since our filer implementation (gitdb) currently mutates the same worktree on every op.
 func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struct{}, updateCh chan runner.RunStatus) (r runner.RunStatus) {
-	log.Infof("run. id: %v", id)
+	log.WithFields(
+		log.Fields{
+			"runID":        id,
+			"requestorTag": cmd.RequestorTag,
```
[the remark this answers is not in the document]
ryancouto: It sorts for us

sched/definitions.go
ryancouto: My understanding was that RequestorTag would be provided by the requestor specifically for the purpose of tracing a job through its lifecycle. Requestor and Tag seem to have orthogonal purposes wrt this goal. Were you thinking of repurposing Tag?

runner/runner.go
```diff
@@ -39,14 +41,21 @@ type Command struct {
 	// Note: nil and empty maps are different!, nil means don't filter, empty means filter everything.
 	// SnapshotPlan map[string]string
 
-	// Runner is given JobID and TaskID to help trace tasks throughout their lifecycle
-	JobID  string
-	TaskID string
+	// Runner is given JobID, TaskID, and RequestorTag to help trace tasks throughout their lifecycle
+	// JobID        string
```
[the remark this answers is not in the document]
ryancouto: Yep

scootapi/client/run_job_cmd.go
ryancouto: To add to that - we don't have the jobID until cl.scootClient.RunJob(jobDef) returns

sched/definitions.go
jasonbs10: That was the original purpose of Tag, it's kind of implied that a tag would be from one requestor. I don't really care about the naming, I just don't think we need both Tag and RequestorTag unless you have another distinction in mind.

scootapi/client/run_job_cmd.go
jasonbs10: In that case, should we be using LogTags at all in TaskDef? What we do on the client side doesn't have to match the structs used by scheduler and worker which have additional context and can supplement the request with jobId etc.

sched/definitions.go
ryancouto: Got it - replacing with Tag

scootapi/client/run_job_cmd.go
ryancouto: Fair enough - guess not!

workerapi/api.go
```diff
@@ -78,6 +96,8 @@ func DomainRunCommandToThrift(domain *runner.Command) *worker.RunCommand {
 	thrift.JobId = &jobID
 	taskID := domain.TaskID
 	thrift.TaskId = &taskID
+	tag := domain.Tag
+	thrift.Tag = &tag
```
jasonbs10: You can use copyString(domain.Tag) here right?

scootapi/server/run_job.go
```diff
@@ -80,6 +80,9 @@ func thriftJobToScoot(def *scoot.JobDefinition) (result sched.JobDefinition, err
 	if def.Priority != nil {
 		result.Priority = sched.Priority(*def.Priority)
 	}
+	if def.Tag != nil {
```
jasonbs10: This is already present above.

scootapi/scoot.thrift
```diff
@@ -51,6 +52,7 @@ struct TaskDefinition {
   # TaskId should generally be unique, otherwise previous tasks with the same Requestor and Tag will be stomped.
   3: optional string taskId,
   4: optional i32 timeoutMs,
+  5: optional string tag
```
jasonbs10: Can we get rid of this?

workerapi/api.go
ryancouto: nope - that fails workerapi/api_test.go

we expect the pointer value even for empty strings

scootapi/server/run_job.go
ryancouto: thanks

scootapi/scoot.thrift
ryancouto: yep

PR #304, opened 2017-09-14
workerapi/server: Log values instead of pointers

ryancouto:
DX-2418


workerapi/server/server.go
```diff
@@ -146,15 +147,20 @@ func (h *handler) QueryWorker() (*worker.WorkerStatus, error) {
...
 	log.WithFields(
 		log.Fields{
 			"argv":       cmd.Argv,
 			"env":        cmd.Env,
-			"snapshotId": cmd.SnapshotId,
-			"timeoutMs":  cmd.TimeoutMs,
-			"jobID":      cmd.JobId,
-			"taskID":     cmd.TaskId,
-			"tag":        cmd.Tag,
+			"snapshotId": snapID,
```
jasonbs10: I'd put these defs inline but this is fine.

PR #312, opened 2017-09-28
runner/runners: Check that the running command isn't nil

ryancouto:


jasonbs10 (commented):
Are we sure this is fixing the underlying issue? If runningID is set shouldn't runningCmd be as well (this code is a little confusing to come back to)?

runner/runners/queue.go
```diff
@@ -222,13 +222,20 @@ func (c *QueueController) Abort(run runner.RunID) (runner.RunStatus, error) {
 func (c *QueueController) abort(run runner.RunID) (runner.RunStatus, error) {
 	if run == c.runningID {
 		if c.runningAbort != nil {
-			log.WithFields(
-				log.Fields{
+			var fields log.Fields
+			if c.runningCmd != nil {
```
dgdonut: Can we init log.Fields and then add to it? i.e. add c.runningID, then use a single 'if' to add runningCmd fields that drops the need for 'else'. (nit) Same for l287

ryancouto:
@jasonbs10  It was the other block (with the "err" field) that panic'd, but the "currentRun" one was sufficiently similar it seemed worthwhile to just double check

jasonbs10:
I'm going to push back on this a little. It looks like runningCmd is always set at the same time as runningID, which you should confirm. When we start adding checks just in case, it confuses the code contracts unless explicitly accompanied by a comment saying we don't know if we need it but just being sure (not my preferred approach).

ryancouto:
@jasonbs10 that's fair. What about the block that did panic?

jasonbs10 (approved):
Approving since we shouldn't block on somewhat trivial stuff like this. Changes are suggested but not required.

PR #313, opened 2017-09-29
RemoteExec - Stub Execute API support over gRPC in scheduler

dgdonut:
This is the stub for the remote exec bazel API. It serves over gRPC by default from the scheduler. This stub is a simple roundtrip request/response, and has no material impact on the scheduler at the moment (you can't send work to Scoot this way).

The client in binaries/execapi makes a "blank" hardcoded request via gRPC. This is a rough stub, but is an initial stepping stone towards actual remote execution support via the bazel API.


JeanetteBruno (approved):
looks good for a prototype.  
- would like to see instructions for trying it out (how to start local scheduler, what is the command to run the cli?) 
- assuming automated test is out of scope.


regarding final production ....

- is binaries/execapi/main.go the cli for remote executing sending a 'bazel' request to the scheduler?  if so, can we just add it to scootapi?  do we need a new dir?
- remexec/README.md 
  . recommend we put high level design docs in confluence
  . the doc linked is the bazel design doc... will need a scoot design - assuming we are translating bazel structures to scoot structures... need a doc for that translation
- the grpc transport should go as a grpc utility in the common/dialer area?
- would put the scheduler's remexec/server stuff right in with the scheduler's thrift server (scootapi/server) (perhaps in a bazel dir: scootapi/server/bazel)

deps.sh
```diff
@@ -4,7 +4,7 @@
 echo "Note: caller must validate/curate/commit changes to .gitmodules and 'vendor/'"
 echo "FIXME: may not work for gopkg.in on OSX: 'Unknown SSL protocol error in connection to gopkg.in:-9838'"
 echo "       see https://github.com/niemeyer/gopkg/issues/49 for context"
-echo '       also try temporarily doing:  'git config --global url."http://".insteadOf https://'
+echo "       also try temporarily doing:  'git config --global url.\"http://\"' instead of https://"
```
jasonbs10, line 7: This doesn't work, the full string is an actual command. You may need to do some escaping though.

jasonbs10 (commented):
Can you elaborate on why this should happen on the server vs as an adapter around our thrift api? I was imagining the latter but haven't thought through pros/cons.

This should be more heavily commented so we don't have to go into the weeds reviewing this.

dgdonut:
* Will add specific run instructions to the sub readme
* Needs tests moving forward, but out of scope for now
*binaries/execapi is only a development tool (thanks for pointing this out, will update the readme). We aren't planning for any production client functionality here, but we still need something for testing. I would rather not mix the thrift client API in scootapi with the grpc API.
* The "scoot translation" design will go in the closed source doc. Linking the Bazel doc here as a ref since our intention is to eventually support that spec.
* Some stuff will end up in libs/common dirs, but waiting for the next phase when further design/direction is more clear
* Not sure where this api should end up. It's not really part of scootapi,although we put all the serving functionality in there.

* Server vs adapter around thrift api - not sure how that would work, what is serving grpc requests in the latter case if not the scheduler? This server would end up going through a translation layer before becoming a 'normal' scoot job. The CAS functionality makes sense to be more centered around apiserver, but the execute service here is pretty close to run_job. This all needs to be formalized in the design before it goes much further.
* Will add more comments

deps.sh
dgdonut: Will update, as is I was getting syntax error running this script (it had an early closing single quote)

PR #315, opened 2017-10-03
runner: Use *int for exitCode in runner.RunStatus

ryancouto:
We got rid of DeadLetterExitCode with e395c1af3aca2f7dd298545533baccc0ff26afd1. The intent was that for tasks that were unable to run, we wouldn't return an exit code. RunStatus's exit code is of type int, which meant when there was no exit code provided, it initialized to a zero value. Changing to an *int allows for nil exit codes.


dgdonut:
What do we get for supporting nil in these codes? API clients like Sickle should just be ignoring ExitCode unless status is COMPLETE, I thought.

JeanetteBruno (approved):
Assuming this change is to be able to differentiate between nil exit code and 0?

Would be nice to provide the description for the worker's QueryWorker() api method explain the semantics of the fields in the WorkerStatus structure.  (Note: I took a look at the public go docs - they were not useful)

jasonbs10 (commented):
I'd table this PR unless it's fixing a specific issue. My preference is to revisit the API holistically when Drew makes changes to support remoting.

ryancouto:
It is fixing a specific issue (DX-2512) - though one that could be solved differently (as @dgassaway  described) without any modifications to OSS

PR #317, opened 2017-10-11
Only Priority3 tasks should preempt lower priority tasks, Priority0,1,2 do not preempt.

jasonbs10:
We may eventually want a more configurable set of priorities but in the meantime it's more useful to have an additional non-preempting priority and only allow P3 to preempt.

### DIFF
--- a/sched/scheduler/stateful_scheduler.go
+++ b/sched/scheduler/stateful_scheduler.go
@@ -56,6 +56,7 @@ const DefaultMaxJobsPerRequestor = 100
 const DefaultSoftMaxSchedulableTasks = 1000
 
 // Increase the NodeScaleFactor by a percentage defined by 1 + (Priority * NodeScaleAdjustment)
+// Note: the use case here is to hit an SLA for each job priority, and this is a coarse way to do so.
 var NodeScaleAdjustment = .75
 
 // Scheduler Config variables read at initialization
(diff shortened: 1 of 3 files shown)


sched/scheduler/stateful_scheduler.go
```diff
@@ -56,7 +56,7 @@ const DefaultMaxJobsPerRequestor = 100
 const DefaultSoftMaxSchedulableTasks = 1000
 
 // Increase the NodeScaleFactor by a percentage defined by 1 + (Priority * NodeScaleAdjustment)
-var NodeScaleAdjustment = 1.0
+var NodeScaleAdjustment = .75
```
ryancouto: Curious - what are you taking into account when adjusting this value? Could you add a short comment?

sched/scheduler/task_scheduler.go
```diff
@@ -116,7 +116,7 @@ func getTaskAssignments(cs *clusterState, jobs []*jobState,
 	// An array indexed by priority. The value is the number of tasks that a job of the given priority can kill.
 	// Only priority=3 and priority=2 jobs can kill other tasks (note, killable tasks are double counted here).
```
ryancouto: Is this still true? (i.e. priority=2 jobs can kill other tasks)

dgdonut (approved):
lgtm less rcouto's comments

PR #320, opened 2017-10-19
Allow P2 jobs

ryancouto:
Respect max priority of P2


dgdonut:
Seems fine, but where are we going with this? Is there an internal ticket with approach we are taking?

ryancouto:
This was just an oversight when we deployed changes using P2 jobs last week

dgdonut (approved):

JeanetteBruno (approved):
Recommend we have a MaxPriority constant in stateful_scheduler.go, to avoid having to remember to change in multiple places in the future

dgdonut:
^ oh, nice

ryancouto:
@JeanetteBruno good call will do

PR #322, opened 2017-10-25
tests: Create integration test for simple CLI job

ryancouto:


tests/integration_test.go
```diff
@@ -0,0 +1,128 @@
+// +build integration
+
+package tests_test
```
dgdonut: just "tests" based on dir conventions? unless this to work around some travis or +build issue

tests/integration_test.go
```diff
@@ -0,0 +1,128 @@
+// +build integration
+
```
dgdonut, line 2: Can we add a package-level comment describing what this test is for? At this point it would also be nice to have a small comment on the readme w/ a one-liner summary of the various tests and what they are for (swarm, recovery, integration)

.gitignore
```diff
@@ -40,3 +40,4 @@ coverage.txt
 .\#*
 
 .scootdata/*
+*/.scootdata/*
```
dgdonut, line 43: does this do anything?

ryancouto: yeah - running this test creates a tests/.scootdata dir that we don't want in here

tests/integration_test.go
ryancouto: yep definitely

tests/integration_test.go
ryancouto: yeah just because of dir conventions

ryancouto: not sure I understand the question. Are you suggesting this should be in the "tests" pkg? Or asking if it's in the "tests_test" pkg b/c of dir conventions?

If the former, we've been using *_test pkg naming convention.
If the latter, yes :)

dgdonut: Oh, hm. I forgot this is another golang thing. Either is acceptable and depends on whether the test is white or black box. I think here it's definitely black box so _test is appropriate.

dgdonut (approved):

tests/integration_test.go
```diff
@@ -0,0 +1,129 @@
...
+func getStatus(jobID string) ([]byte, error) {
+	gopath, err := getGopath()
+	if err != nil {
+		return nil, err
+	}
+	return exec.Command(gopath+"/bin/scootapi", "get_job_status", jobID, "--json").Output()
+}
+
+func getGopath() (gopath string, err error) {
```
jasonbs10, line 118: This should split on ':' and take the first entry.

PR #325, opened 2017-10-26
Trying timeouts on os_execer debug commands

jasonbs10:
It appears that os.execer CombinedOutput() can fail to return...

### DIFF
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -195,6 +195,7 @@ func (e *osExecer) monitorMem(p *osProcess) {
 						"taskID":      p.TaskID,
 					}).Infof("Increased mem_cap utilization for pid %d to %d", pid, int(memUsagePct*100))
 
+				// Debug output with timeout since it seems CombinedOutput() sometimes fails to return.
 				ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
 				ps, err := exec.CommandContext(ctx, "ps", "-u", os.Getenv("USER"), "-opid,sess,ppid,pgid,rss,args").CombinedOutput()
 				log.WithFields(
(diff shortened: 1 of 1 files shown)


runner/execer/os/os_execer.go
```diff
@@ -193,7 +194,9 @@ func (e *osExecer) monitorMem(p *osProcess) {
 						"jobID":       p.JobID,
 						"taskID":      p.TaskID,
 					}).Infof("Increased mem_cap utilization for pid %d to %d", pid, int(memUsagePct*100))
-				ps, err := exec.Command("ps", "-u", os.Getenv("USER"), "-opid,sess,ppid,pgid,rss,args").CombinedOutput()
+
+				ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+				ps, err := exec.CommandContext(ctx, "ps", "-u", os.Getenv("USER"), "-opid,sess,ppid,pgid,rss,args").CombinedOutput()
```
ryancouto: Recommend logging ctx.Err() if it == context.DeadlineExceeded so we know this is happening

runner/execer/os/os_execer.go
```diff
@@ -193,16 +194,21 @@ func (e *osExecer) monitorMem(p *osProcess) {
 						"jobID":       p.JobID,
 						"taskID":      p.TaskID,
 					}).Infof("Increased mem_cap utilization for pid %d to %d", pid, int(memUsagePct*100))
-				ps, err := exec.Command("ps", "-u", os.Getenv("USER"), "-opid,sess,ppid,pgid,rss,args").CombinedOutput()
+
+				ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
```
ryancouto: Quick nit - could you add a comment providing context for the... context. It isn't immediately clear looking at this why exec.Cmd.CombinedOutput() would need a timeout

ryancouto:
I'm curious, should we be doing this for all of our exec.Command's? All Output()'s? Or just CombinedOutput()'s?

jasonbs10:
I'm just doing it for execer right now since these are debug and we don't care if they terminate early. Doing it for the git operations would be a little more involved.

ryancouto (approved):

dgdonut (approved):

PR #326, opened 2017-11-01
Removed SoftMaxSchedulableTasks config val and made it dynamic.

jasonbs10:
It's easier to set this based on actual task load than having to guess a config value.


sched/scheduler/task_scheduler.go
```diff
@@ -39,21 +39,23 @@ func getTaskAssignments(cs *clusterState, jobs []*jobState,
...
 	}
 	if totalUnschedTasks == 0 {
 		return nil, nil
 	}
 
+	// Udate SoftMaxSchedulableTasks based on number of healthy nodes and the total number of tasks.
+	if config == nil {
+		config = &SchedulerConfig{}
+	}
+	config.SoftMaxSchedulableTasks = max(config.SoftMaxSchedulableTasks, totalOutstandingTasks, len(cs.nodes))
```
JeanetteBruno: I've been trying to understand the scheduling algorithm - need much more time to digest it, perhaps we can review it in tech discussion and update the write up.... given that:
- config.SoftMaxSchedulableTasks will be ever increasing since config.SoftMaxSchedulableTasks is part of the max statement - this seem suspicious

do we have tests validating the scheduling?

dgdonut (approved):
Still a bit of a head scratcher, but I guess not any more than it already was

sched/scheduler/task_scheduler.go
```diff
@@ -39,21 +39,24 @@ func getTaskAssignments(cs *clusterState, jobs []*jobState,
...
 	if totalUnschedTasks == 0 {
 		return nil, nil
 	}
 
+	// Udate SoftMaxSchedulableTasks based on number of healthy nodes and the total number of tasks.
+	// Setting the max to num healthy nodes means that each job can be fully scheduled.
+	if config == nil {
+		config = &SchedulerConfig{}
+	}
+	config.SoftMaxSchedulableTasks = max(config.SoftMaxSchedulableTasks, totalOutstandingTasks, len(cs.nodes))
```
ryancouto: Still a little concerned about this line - seems weird that it never resets & is always increasing.

jasonbs10: Ok, I took that part out since it appears to be more confusing than useful.

PR #330, opened 2017-11-14
ContentAddressableStore in Apiserver

dgdonut:
This is an initial implementation for CAS in Apiserver/Bundlestore. This is large (sorry) so please read for some context before diving in.

* Changes to files outside of bazel/ are small or just rearrangements. In general, I separated snapshot Stores into snapshot/store/ to separate them from snapshot/bundlestore/, which functions as a server.

How does CAS fit into Scoot?
* Existing Apiserver acted as a Bundlestore, serving a custom HTTP API in front of an underlying Snapshot Store.
* Now, Bundlestores serve both our HTTP API and the GRPC CAS API, to the same underlying Store.

Small stuff:
* Makefile was changed because of binaries from MDE now being on the path


bazel/digest.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"fmt"
+
+	remoteexecution "google.golang.org/genproto/googleapis/devtools/remoteexecution/v1test"
+)
+
+// Validate Digest hash and size components assuming SHA256
+func IsValidDigest(hash string, size int64) bool {
+	return len(hash) == 64 && size >= 0
```
illicitonion: Would it be possible to drop the size >= 0 constraint, and remove size from the addressing information of blobs?

I have desires to fetch digests without knowing how big they are in advance, and it feels like passing size=-1 is a better way of doing that then specifying a valid-but-incorrect size.

(I will push this feedback upstream too)

bazel/cas/service.go
```diff
@@ -66,27 +98,173 @@ func (s *casServer) GetTree(
...
+	}
+
+	// Parse resource name per Bazel API specification
+	resource, err := ParseReadResource(req.GetResourceName())
+	if err != nil {
+		return status.Error(codes.InvalidArgument, fmt.Sprintf("%v", err))
+	}
+
+	// Input validation per API spec
+	if req.GetReadOffset() < 0 || req.GetReadOffset() >= resource.Digest.GetSizeBytes() {
```
illicitonion: In line with my other comment about size, can you drop the GetReadOffset() >= resource.Digest.GetSizeBytes() check (and make the check be against the actual underlying blob's real size)?

bazel/cas/service.go
```diff
@@ -66,27 +98,173 @@ func (s *casServer) GetTree(
...
+	res := &bytestream.ReadResponse{}
+	c := resource.Digest.GetSizeBytes() - req.GetReadOffset()
+
+	// If ReadOffset was specified, discard bytes prior to it
+	if req.GetReadOffset() > 0 {
+		_, err = io.CopyN(ioutil.Discard, r, req.GetReadOffset())
+		if err != nil {
+			return status.Error(codes.Internal, "Failed to read until offset")
+		}
+		c = c - req.GetReadOffset()
```
illicitonion: Why are you re-subtracting this after it was already subtracted on line 133?

bazel/cas/service.go
```diff
@@ -66,27 +98,173 @@ func (s *casServer) GetTree(
...
+			return status.Error(codes.InvalidArgument, fmt.Sprintf("ResourceName %s mismatch with previous %s", wr.GetResourceName(), resourceName))
+		}
+		if wr.GetWriteOffset() > 0 {
+			return status.Error(codes.Unimplemented, "Currently unsupported in Scoot - Writes are not resumable")
+		}
+
+		buffer.Write(wr.GetData())
+
+		// Per API, client indicates all data has been sent
+		if wr.GetFinishWrite() {
```
illicitonion, line 257: As we're not implementing resumable writes, it feels like we probably want to error if this isn't set to true, and always break.

common/endpoints/endpoints.go
```diff
@@ -41,7 +41,7 @@ func (s *TwitterServer) Serve() error {
 	for path, handler := range s.Handlers {
 		mux.Handle(path, handler)
 	}
-	log.Info("Serving http & stats on", s.Addr)
+	log.Info("Serving http & stats on: ", s.Addr)
```
illicitonion, line 44: You probably don't actually want a trailing space here; log.Info will add one

illicitonion (changes requested):
Looks good :) I haven't looked too closely at the tests.

Overall, mostly just one comment about using size in the blob name.

I guess at some point we may want to add a way of setting a TTL for objects added over grpc, and I guess we should talk with the API folks about adding that as a concept. It feels like something they'll care about too, sooner or later.

dgdonut:
Back from the dead. Thanks for reviewing!

On leaving size out - this is a big departure from the API.  It is pretty strongly worded around this. What kind of scenario would exist where you know the hash of what you want to retrieve but not the size?

> // The size is considered to be an integral part of the digest and cannot be
// separated. That is, even if the `hash` field is correctly specified but
// `size_bytes` is not, the server MUST reject the request.

Updating now with the minor things.

On TTL - API-providable would be nice. In the meantime, we essentially set up a default in the Apiserver config, which should get us going initially.

bazel/cas/service.go
dgdonut: I'd like to circle back on resumable writes soon - mainly taking a shortcut there. I think we still need this as even without resumables, we have to loop over the this and check each request that is part of the write stream.

common/endpoints/endpoints.go
dgdonut: Space is actually needed here

illicitonion:
Welcome back!

On size:

I have a post out to the buildfarm list: https://groups.google.com/forum/#!topic/bazel-buildfarm/yeCu-lw4MYM - my concrete use case is that I'd like to be able to explore the CAS without needing to have been explicitly given a reference to a Digest from an ExecuteRequest - I put together a little FUSE filesystem to do this (e.g. you can `cat /mnt/cas/file/[sha256]` or `ls /mnt/cas/directory/sha256`), and it would be a little annoying to have to add size into there, as it means that you can only use it to explore things which you've been told exist, rather than also exploring things that you don't know much about, but want to learn about.

An example of where I think this could be useful is debugging issues in Pants (or Scoot) where something messes up size computations (I've certainly messed up by doing something like truncating the last frame in a buffer by forgetting to flush a stream); being able to say "Oh, if I ask for digest foo, scoot knows about foo, but when I fetched it from scoot, it's actually one byte longer than we asked for, I understand where this bug may be" seems like a useful ability which has very little cost associated with it.

In fact, just a more informative error message from scoot ("You asked for digest=foo,size=10, but we think digest=foo has size 12") would be useful.

The only thing making these things hard with this implementation is that size is in the name of the underlying resource. Whether we make size optional in the API, or make the Read error message informative like above, are definitely discussions we can defer, but being able to have that discussion is predicated on scoot being able to find the digest for which the size is missing/wrong. Just dropping the size from the name seems like a cosmetic change, but enables all of the above.

I only really care about the read path here, where I'd like size to be optional. On read, you're currently only functionally using the user-provided size to size the capacity of a buffer (where if no size were provided, picking a sensible default like 1M doesn't seem crazy).

dgdonut:
I'm willing to make this change for now, although we lose some safety-related convenience on the server by omitting the length - our Store abstraction has no 'stat' method, so we have to assume the content fetched from the underlying file/http/cache resource ends up with the right size.

jasonbs10:
We don't have to store the exact bytes in blobstore do we? I'd think that we are free to additionally store the size when we do a write and use that to verify size-less reads. Right now we're doing git bundles but whatever new format we use can incorporate whatever metadata we want. Generally I'm in strong agreement with Daniel that the read shouldn't require size.

dgdonut:
We can't store the size because it becomes an addressing problem for size-less read requests, unless we put it in some kind of encapsulated structure or store metadata/mapping somewhere. I'm just going to drop the size for now, although I think the "in case I computed the size wrong" scenario seems convoluted.

dgdonut:
Please approve when you have a chance.

Makefile
```diff
@@ -76,14 +76,14 @@ swarmtest:
 	# Then run (with go run) scootapi run_smoke_test with 10 jobs, wait 1m
 	# We build the binaries becuase 'go run' won't consistently pass signals to our program.
 	go install ./binaries/...
-	setup-cloud-scoot --strategy local.local run scootapi run_smoke_test --num_jobs 10 --timeout 1m $(TRAVIS_FILTER)
+	$(GOPATH)/bin/setup-cloud-scoot --strategy local.local run scootapi run_smoke_test --num_jobs 10 --timeout 1m $(TRAVIS_FILTER)
```
jasonbs10: How does this work with parenthesis instead of brackets? Also GOPATH needs to be split on the first instance of ';' if any.

illicitonion (approved):
Looks good :) Thanks!

bazel/cas/resource.go
```diff
@@ -0,0 +1,109 @@
...
+	"strconv"
+	"strings"
+
+	uuid "github.com/nu7hatch/gouuid"
+	remoteexecution "google.golang.org/genproto/googleapis/devtools/remoteexecution/v1test"
+
+	"github.com/twitter/scoot/bazel"
+)
+
+// Keep track of a Resource, identified primarily by a Digest.
```
jasonbs10: This could use more comments, i.e. why does this need three identifiers, two of which are unique?

bazel/cas/service.go
jasonbs10: I think we shouldn't break when we can reasonably satisfy a request. That said, depending on how long we keep some manner of shortcut, if it should be an error to upload something that's inordinately large or taking too much time between each write.

bazel/constants.go
```diff
@@ -0,0 +1,9 @@
+package bazel
+
+const (
+	// Bazel Instance-related constants
+	DefaultInstanceName = ""
+
+	// Bazel Resource storage-related constants
```
jasonbs10, line 7: PrePost?

snapshot/bundlestore/README.md
```diff
@@ -10,13 +10,33 @@ There are a few different store types that handle bundles differently. They can
 in memory, bundles on disk, and bundles located behind an http server. Further, we can
 wrap these stores in a parent store to add additional business logic to our handling.
 
+## Bazel Content Addressable Store
+Bundlestore servers now also support the CAS API over gRPC, backed by the same underlying Store as the HTTP API.
+
 ## Bundle name conventions
-For now names look like 'bs-<sha>.bundle'.
+* HTTP Bundlestore server - For now names look like 'bs-<sha>.bundle'
+* GRPC Bazel CAS server - based on Digest - see Bazel remote execution/CAS documentation
```
jasonbs10: Link? And what does a grpc name look like?

snapshot/bundlestore/README.md
```diff
@@ -10,13 +10,33 @@ There are a few different store types that handle bundles differently. They can
...
+Note that the above means that although the Store backing a server can support HTTP and GRPC APIs,
+the HTTP API naming restriction prevents accesing GRPC artifacts.
```
jasonbs10: sp: accessing. I'm not sure what the implications of this are / why we can't check for 40 or 64 digits and act accordingly.

snapshot/bundlestore/server.go
```diff
@@ -2,144 +2,66 @@ package bundlestore
...
 
 	log "github.com/sirupsen/logrus"
 
+	"github.com/twitter/scoot/bazel"
+	"github.com/twitter/scoot/bazel/cas"
 	"github.com/twitter/scoot/common/stats"
+	"github.com/twitter/scoot/snapshot/store"
 )
 
+// Represents a Bundlestore Server that serves HTTP and gRPC Bazel CAS APIs
```
jasonbs10, line 16: The naming change from bundlestore -> store seems incomplete? When are we choosing to use one or the other?

jasonbs10 (approved):
Generally looks ok although this'll require a more comprehensive review at some later point.

Makefile
dgdonut: Updated to work with split gopaths. Parens because makefile.

bazel/cas/resource.go
dgdonut: added

bazel/cas/service.go
dgdonut: that's not what this is doing, the GetFinishWrite is the end of a normal successful write request. the write server could probably use some timeout or safeguard but this is fine for v1/now

bazel/constants.go
dgdonut: changed to just prefix

snapshot/bundlestore/README.md
dgdonut: this was mainly for by-hand testing convenience, e.g. write with grpc then read with http. the APIs are not really cross-compatible for other reasons, though, so I left out naming derestriction for http until there is a real reason to have it

snapshot/bundlestore/server.go
dgdonut: I'm not renaming bundlestore to store, I just separated Store into its own package so it is portable. If anything bundlestore should be apiserver or blobserver or storeserver or something now.

PR #332, opened 2017-12-05
snapshot/bazel: Snapshot implementation

ryancouto:


snapshot/bazel/filer.go
```diff
@@ -0,0 +1,41 @@
...
+	localStorePath string
+	// Not yet implemented:
+	// bypassLocalStore bool
+	// skipServer       bool
+	// serverAddress    string
+}
+
+func (bf *bzFiler) RunCmd(args []string) ([]byte, error) {
+	if bf.localStorePath != "" {
+		args = append(args, "--local_store_path", bf.localStorePath)
```
illicitonion: This ended up getting changed to --local-store-path because it turns out pants uses -s everywhere. Sorry about that!

snapshot/bazel/filer.go
```diff
@@ -0,0 +1,41 @@
...
+
+// Satisfies snapshot.Checkouter, snapshot.Ingester, and snapshot.Updater
+type bzFiler struct {
+	command        string
+	updater        snapshot.Updater
+	localStorePath string
+	// Not yet implemented:
+	// bypassLocalStore bool
+	// skipServer       bool
+	// serverAddress    string
```
illicitonion: FYI this is implemented in https://github.com/pantsbuild/pants/pull/5179 (but currently only used for read operations). Write operations coming soon!

illicitonion: As soon as we support write operations (probably next week), I'll start publishing compiled fs_util binaries somewhere useful :)

ryancouto: cool! looking forward to giving it a whirl

snapshot/bazel/filer.go
ryancouto: no prob 👍

snapshot/bazel/constants.go
```diff
@@ -0,0 +1,10 @@
+package bazel
+
+const (
+	invalidIdMsg       = "Expected ID to be of format bz-<sha256>-<sizeBytes>, was"
+	invalidFileTypeMsg = "Error getting fileType of path"
+	noSuchFileOrDirMsg = "no such file or directory"
+
+	directory = "directory"
```
dgdonut: Nit, but prefer these to be more specific for readability. Like "bzCmdDirectory" or something (or maybe it's just me). Also include "materialize" here.

snapshot/bazel/checkouter.go
```diff
@@ -0,0 +1,62 @@
...
+	if !bazel.IsValidDigest(sha, size) {
+		return nil, fmt.Errorf("Error: Invalid digest. SHA: %v, size: %d", sha, size)
+	}
+
+	output, err := bf.RunCmd([]string{directory, "materialize", sha, dir})
+	if err != nil {
+		log.WithFields(
+			log.Fields{
+				"error":  err,
+				"output": string(output),
```
dgdonut: I don't think it's safe to access output here, or at least it defies usual "output, err := " conventions

snapshot/bazel/checkouter.go
```diff
@@ -0,0 +1,62 @@
...
+				"sha":    sha,
+				"dir":    dir,
+			}).Error("Failed to materialize %s", directory)
```
dgdonut: want to log var id instead of directory?

snapshot/bazel/filer.go
```diff
@@ -0,0 +1,41 @@
+package bazel
+
+import (
+	"github.com/twitter/scoot/snapshot"
+	"github.com/twitter/scoot/snapshot/snapshots"
+	"os/exec"
```
dgdonut: nit - ordering

snapshot/bazel/checkout.go
```diff
@@ -0,0 +1,25 @@
...
+	dir string
+	remoteexecution.Digest
+}
+
+func (bc *bzCheckout) Path() string {
+	return bc.dir
+}
+
+func (bc *bzCheckout) ID() string {
+	return bc.GetHash()
```
dgdonut: Should return "bz-\<hash\>-\<size\>"?

snapshot/bazel/ingester.go
```diff
@@ -0,0 +1,49 @@
...
+			}).Errorf("Error ingesting %s %s", fileType, path)
+		return "", err
+	}
+
+	id := string(output)
+	return id, nil
+}
+
+func (bf *bzFiler) IngestMap(srcToDest map[string]string) (string, error) {
+	panic("not implemented")
```
dgdonut: Prefer to error log and no-op rather than panic

snapshot/bazel/ingester.go
```diff
@@ -0,0 +1,49 @@
...
+	fileType, err := bf.getFileType(path)
+	if err != nil {
+		return "", err
+	}
+	output, err := bf.RunCmd([]string{fileType, "save", path})
+	if err != nil {
+		log.WithFields(
+			log.Fields{
+				"err":    err,
+				"output": string(output),
```
dgdonut: same comment about logging output

snapshot/bazel/ingester.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		log.Errorf("%s %s: %v", invalidFileTypeMsg, path, err)
+		return fileType, err
+	}
+
+	switch mode := stat.Mode(); {
+	case mode.IsDir():
+		fileType = "directory"
+	case mode.IsRegular():
+		fileType = "file"
```
dgdonut: default case should probably return an err

snapshot/bazel/ingester.go
```diff
@@ -0,0 +1,49 @@
...
+		log.WithFields(
+			log.Fields{
+				"err":    err,
+				"output": string(output),
+				"path":   path,
+			}).Errorf("Error ingesting %s %s", fileType, path)
+		return "", err
+	}
+
+	id := string(output)
```
dgdonut: Need to parse and validate output, and return ID conforming to our expectation of bz-\<digest\>-\<size\> format. In our cases, fs_util returns "\<digest\> \<size\>".

We should probably have a couple util functions for all ID-related operations - generating from hash & size and validating ID strings (similar to bazel/digest.go).

snapshot/bazel/filer.go
```diff
@@ -0,0 +1,41 @@
...
+
+func MakeBzFilerWithLocalStore(localStorePath string) *bzFiler {
+	return &bzFiler{
+		command:        "fs_util",
+		updater:        snapshots.MakeNoopUpdater(),
+		localStorePath: localStorePath,
+	}
+}
+
+// Satisfies snapshot.Checkouter, snapshot.Ingester, and snapshot.Updater
```
dgdonut: plus comment on expected filer usage wrt "command"

snapshot/bazel/filer_test.go
```diff
@@ -0,0 +1,189 @@
...
+	tmpFile, err := tmp.TempFile("")
+	if err != nil {
+		t.Fatalf("Error creating temp file. %v", err)
+	}
+
+	_, err = noopBf.Ingest(tmpFile.Name())
+	if err != nil {
+		t.Fatalf("Error ingesting file %v. Err: %v", tmpFile.Name(), err)
+	}
+
```
dgdonut: would add a check against returned ID here to make sure it checks out - should be able to expect hardcoded hash and size 0 from empty temp file (see also later comment on ID returned form Ingest)

dgdonut (commented):
Also, add basic readme w/ simple summary of what this does, reliance on fs_util tool, and snapshot ID format

snapshot/bazel/ingester.go
ryancouto: For this one I just copied the implementation from dbaAdapter - I prefer an error log no-op too and will change both.

snapshot/bazel/utils.go
```diff
@@ -0,0 +1,81 @@
...
+	var fileType string
+	stat, err := os.Stat(path)
+	if err != nil {
+		log.Errorf("%s %s: %v", invalidFileTypeMsg, path, err)
+		return fileType, err
+	}
+
+	switch mode := stat.Mode(); {
+	case mode.IsDir():
+		fileType = "directory"
```
dgdonut: use the constants here

snapshot/bazel/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+### Snapshot ID format
+Bazel Remote Execution snapshots use Snapshot IDs of format bz-<sha256>-<sizeBytes> and map to a unique Bazel digest
+
+### Components:
+* bzFiler satisfies the snapshot.Snapshot interface
+  * Checks out a Bazel Digest as a snapshot & sets up local filesystem
+  * Ingests directory or file and stores it as a snapshot
+  * noop Updates
+
```
dgdonut: add bzCheckout

snapshot/bazel/ingester.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	return id, err
+}
+
+func (bf *bzFiler) Ingest(path string) (string, error) {
+	fileType, err := getFileType(path)
+	if err != nil {
+		return "", err
+	}
+	output, err := bf.RunCmd([]string{fileType, "save", path})
```
dgdonut: might as well make "save" a const as well

snapshot/bazel/checkouter.go
```diff
@@ -0,0 +1,61 @@
...
+		log.WithFields(
+			log.Fields{
+				"error": err,
+				"sha":   sha,
+				"dir":   dir,
+			}).Error("Failed to materialize %s", id)
+		return nil, err
+	}
+
+	log.Info(string(output))
```
dgdonut: can ditch this, no expected output

dgdonut (commented):
My suggestion for the validator situation:

Turn command into an interface, and rename it (uhh... no idea... BazelRunner? TreeRunner? BazelFS?). The interface can have a save and materialize function that support the Ingester and Checkouter use cases.

bzFiler can use an implementation that does what RunCmd is doing currently (and should do the parsing of fs_util output into the real ID). Ingest can call the save function and then should do id fmt validation, and IngestAndValidate can go away.

For the tests, we can just make a fake implementation that can be really simple for now.  Can do a real Ingest + Checkout integration test once we're actually using fs_util.

snapshot/bazel/utils.go
```diff
@@ -46,6 +47,11 @@ func validateID(id string) error {
 	return nil
 }
 
+// Checks that bytes of ID is well formed
```
dgdonut: unused?

snapshot/bazel/bz_cmd.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	output, err := bc.runCmd([]string{fileType, fsUtilCmdSave, path})
+	if err != nil {
+		return nil, err
+	}
+
+	err = validateID(string(output))
```
dgdonut: this impl. function will have to do its own conversion from fs_util output into a bazel snapshot id before we can validate - fs_util returns space-separated hash/size, and lacks the bz- prefix

snapshot/bazel/bz_cmd.go
```diff
@@ -0,0 +1,58 @@
+package bazel
+
+import (
+	"os/exec"
+)
+
+type bzRunner interface {
+	save(path string) ([]byte, error)                   // Called by bzFiler.Ingest
```
dgdonut: save interface might want to return string (for snapshot ID). materialize, I don't think needs any output beyond error. WDYT?

dgdonut (approved):

dgdonut:
lgtm

PR #334, opened 2017-12-07
Pass env vars from scootapi through to worker.

jasonbs10:
Also updated snapshot/git/gitdb/stream.go to bypass refspec.

### DIFF
--- a/runner/execer/execer.go
+++ b/runner/execer/execer.go
@@ -15,12 +15,12 @@ import (
 type Memory uint64
 
 type Command struct {
-	Argv   []string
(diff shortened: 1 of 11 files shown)


dgdonut (approved):
What was the usecase for the refspec change?

jasonbs10:
The refspec change is to support the functionality discussed with aurora-team. They won't have the src repo with which to generate a real snapshotid, so they'll construct a standalone stream id with their desired sha, and the optional suffix if that sha is on an untracked branch (only master is tracked by default on worker startup).

PR #337, opened 2017-12-13
snapshot/bazel: integrate fs_util tool

ryancouto:
* Adds a make target to fetch a versioned, architecture appropriate fs_util
* Integration tests for usage of bzFiler & bzCommand with fs_util


get_fs_util.sh
```diff
@@ -0,0 +1,31 @@
...
+			local url="https://binaries.pantsbuild.org/bin/fs_util/linux/x86_64/$pants_release.dev$fs_util_release/fs_util"
+			;;
+
+		*)
+			echo "Expected Darwin or Linux output from uname, received $arch"
+			;;
+
+	esac
+
+	wget "$url" -P "$GOPATH/bin/"
```
dgdonut: GOPATH supports ':' separated lists, so we should do what Makefile is doing with FIRSTGOPATH

snapshot/bazel/bz_cmd.go
```diff
@@ -24,8 +28,16 @@ func (bc bzCommand) save(path string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	args := []string{fileType, fsUtilCmdSave}
+
+	// directory save requires root path
+	if fileType == fsUtilCmdDirectory {
+		args = append(args, fsUtilCmdRoot, path, fsUtilCmdGlobWildCard)
```
dgdonut: Why is this needed? Assuming a setup such as: /var/bazel/sandbox/ is our root and we want to ingest /var/bazel/sandbox/foo (foo being a dir), does `fs_util directory save --root /var/bazel/sandbox foo` not work?

I'm guessing this is because of something in the way fs_util works, but I think intuitively the above should work

snapshot/bazel/bz_cmd.go
```diff
@@ -58,11 +70,20 @@ func (bc bzCommand) materialize(sha string, dir string) error {
...
 // Runs fsUtilCmd as an os/exec.Cmd with appropriate flags
 func (bc bzCommand) runCmd(args []string) ([]byte, error) {
+	if bc.serverAddr != "" {
+		args = append([]string{fsUtilCmdServerAddr, bc.serverAddr}, args...)
+	}
 	if bc.localStorePath != "" {
-		args = append(args, "--local-store-path", bc.localStorePath)
+		args = append([]string{fsUtilCmdLocalStore, bc.localStorePath}, args...)
+	}
+	// We expect fs_util binary to be located at $GOPATH/bin, due to get_fs_util.sh
```
dgdonut, line 103: same thing about first go path. may want to be a function that lives in common/

snapshot/bazel/checkouter.go
```diff
@@ -13,10 +15,14 @@ func (bf *bzFiler) Checkout(id string) (snapshot.Checkout, error) {
 	if err != nil {
 		return nil, err
 	}
-	return bf.CheckoutAt(id, tempDir.Dir)
+	return bf.CheckoutAt(id, path.Join(tempDir.Dir, fsUtilCmdMaterialize))
```
dgdonut: I am wondering what we should use as our filer root directories - we may want to initialize bzFilers to a certain root location (or a temp by default) - so that these can be easier to inspect.

This is more of an optimization, but it might be nice to see e.g. aurora-sandbox/workspace/bazel/ full of bazel exec checkout dirs that can be inspected. This would also mean having an option to not blow away the checkout dir on Release, but I suspect both of these things might be useful during initial debugging.

We could do some of, or part of, or none of this now and record in a ticket.

get_fs_util.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# Pulls in fs_util tool for use in remote exec ingestion, checkout, and CAS ops
+
+pants_release="1.4.0"
```
illicitonion: Technically the pants release here is 1.4.0dev23, and fs_util is just a part of the pants release.

get_fs_util.sh
```diff
@@ -0,0 +1,31 @@
...
+			local url="https://binaries.pantsbuild.org/bin/fs_util/mac/10.13/$pants_release.dev$fs_util_release/fs_util"
+			;;
+
+		Linux)
+			echo 'Linux'
+			local url="https://binaries.pantsbuild.org/bin/fs_util/linux/x86_64/$pants_release.dev$fs_util_release/fs_util"
+			;;
+
+		*)
+			echo "Expected Darwin or Linux output from uname, received $arch"
```
illicitonion: $arch isn't set anywhere

snapshot/bazel/fs_util_test.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	bf := makeTestingFiler()
+	id, err := bf.Ingest(root.Dir)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if id == emptyID {
+		t.Fatalf("Expected id to not be %s, was %s", emptyID, id)
+	}
+	size, err := getSize(id)
```
illicitonion: You should be able to assert on the actual expected size and sha here, rather than just that they're non-empty?

Handy hint for manually doing this kind of thing, to generate a binary serialized proto you can run something like:
```
echo "files: {
  name: 'cheese'
  digest: {
    hash: '727d05cc3cb7b48eadc75bbbb8d740a9321691745f136d58270b93ba4f88c218'
    size_bytes: 19
  }
}
" | protoc --encode=google.devtools.remoteexecution.v1test.Directory google/devtools/remoteexecution/v1test/remote_execution.proto --protopath=/path/to/googleapis
```

which you can then pipe to shasum or wc to get shas and sizes.

snapshot/bazel/fs_util_test.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	bf := makeTestingFiler()
+	id, err := bf.Ingest(f.Name())
+	if err != nil {
+		t.Fatal(err)
+	}
+	if id == emptyID {
+		t.Fatalf("Expected id to not be %s, was %s", id, emptyID)
+	}
+	size, err := getSize(id)
```
illicitonion: As above

snapshot/bazel/fs_util_test.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	newTemp, err := temp.TempDirDefault()
+	if err != nil {
+		t.Fatal(newTemp)
+	}
+	_, err = bf.Checkout(id)
+	if err != nil {
```
illicitonion: Look for the right files (and maybe their contents)?

Just running `diff` on the two directories should do the job pretty easily.

illicitonion (commented):
Looks good, may want to firm up the tests a little.

snapshot/bazel/filer.go
```diff
@@ -5,20 +5,26 @@ import (
...
-		command: bzCommand{
-			localStorePath: localStorePath,
-		},
-		updater: snapshots.MakeNoopUpdater(),
+// Options is a variadic list of functions that take a *bzCommand as an arg
+// and modify its fields, e.g.
+// localStorePath := func(bc *bzCommand) {
+//     bc.localStorePath = "/path/to/local/store"
+// }
+func MakeBzFilerWithOptions(options ...func(*bzCommand)) *bzFiler {
```
dgdonut, line 20: Have to say I am not a fan of this pattern, strongly prefer explicit parameter list that constructs bzcommand as required. If this was an external API, maybe, but this is a specific interface implementation and not likely to cause breaking changes

get_fs_util.sh
ryancouto: whoops, good catch. thanks

snapshot/bazel/fs_util_test.go
ryancouto: cool, thanks for the tip

snapshot/bazel/checkouter.go
ryancouto: agreed - think that would be very useful. I'll throw in the option to not blow away on Release in here, and we can think more about how we'd want these root dirs to look like and implement later.

snapshot/bazel/filer.go
ryancouto: My hope was to avoid having MakeBzFiler, MakeBzFilerWithLocalStore, MakeBzFilerWithServerAddr, MakeBzFilerWithLocalStoreAndServerAddr, & all the other possible options when skipServer and skipLocalStore are enabled

dgdonut: We wouldn't need more than one constructor - MakeBzFiler("", "", false, false) would create the filer with no local store, no server addr, skip options disabled for example.  I will leave this one up to you to change or leave as-is

snapshot/bazel/bz_cmd.go
ryancouto: you have to specify the globs to save - your example would save foo but not any files contained in foo, for that you would need foo/*

dgdonut (approved):
lgtm - might add a separate test later verifying ingest output vs non-empty sha sums and such

PR #339, opened 2017-12-15
bazel/execution - making an execution request starts a hardcoded scoot job

dgdonut:
- execute request starts a job
- made sched validateJob public
- readme fix from #338

### DIFF
--- a/bazel/execution/service.go
+++ b/bazel/execution/service.go
@@ -5,30 +5,44 @@ package execution
 import (
 	"fmt"
 	"net"
+	"time"
 
 	"github.com/golang/protobuf/ptypes"
-	uuid "github.com/nu7hatch/gouuid"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	remoteexecution "google.golang.org/genproto/googleapis/devtools/remoteexecution/v1test"
 	"google.golang.org/genproto/googleapis/longrunning"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/twitter/scoot/common/grpchelpers"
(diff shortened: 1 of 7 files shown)


bazel/execution/service.go
```diff
@@ -77,10 +116,28 @@ func (s *executionServer) Execute(
...
+	if req == nil {
+		return result, fmt.Errorf("Nil execute request")
+	}
+
+	// This creates a hardcoded job for now
+	result.Tasks = []sched.TaskDefinition{}
+	var task sched.TaskDefinition
+	task.TaskID = fmt.Sprintf("ExecuteRequest_%d", time.Now().Unix())
+	task.Command.Argv = []string{"sleep", "10"}
+	task.Command.EnvVars = make(map[string]string)
```
ryancouto, line 142: task.Command.SnapshotID is required too, right?

bazel/execution/service.go
```diff
@@ -77,10 +116,28 @@ func (s *executionServer) Execute(
...
 	op.Result = &longrunning.Operation_Response{Response: resAsPBAny}
 	return &op, nil
 }
+
+func execReqToScoot(req *remoteexecution.ExecuteRequest) (result sched.JobDefinition, err error) {
+	if req == nil {
+		return result, fmt.Errorf("Nil execute request")
+	}
+
+	// This creates a hardcoded job for now
```
ryancouto, line 134: Obvious, but could you add something slightly more explicit wrt what's left TODO

bazel/execution/service.go
dgdonut: Adding later in another ticket - putting snapshot in now will basically just make these tasks fail every time since they aren't likely to be available. Will add a todo

ryancouto (approved):

PR #340, opened 2017-12-20
Most of the gitdb requests should defer thread safety to the git binary.

jasonbs10:
We observed congestion in apiserver while trying to hit the /view endpoint. At a minimum we shouldn't be blocking concurrent-capable requests.

### DIFF
--- a/sched/scheduler/stateful_scheduler.go
+++ b/sched/scheduler/stateful_scheduler.go
@@ -480,6 +480,7 @@ func (s *statefulScheduler) updateStats() {
 		}
 
 		if time.Now().Sub(job.TimeMarker) > LongJobDuration {
+			job.TimeMarker = time.Now()
 			log.WithFields(
 				log.Fields{
 					"requestor":      job.Job.Def.Requestor,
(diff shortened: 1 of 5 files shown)


ryancouto (approved):
couple of nits: think those goroutines could be refactored into a helper func to help with readability.
also if you add a //+ build integration tag to fs_util_test.go we shouldn't need to make fs_util on test-unit and test-unit-property. otherwise lgtm

jasonbs10:
I added the build tag and some additional logging for long-running jobs. I'm going to skip the refactor though.

dgdonut (approved):
seems reasonable

sched/scheduler/stateful_scheduler.go
```diff
@@ -476,6 +479,19 @@ func (s *statefulScheduler) updateStats() {
 			s.requestorsCounts[requestor][numWaitingTasksKey] += len(job.Tasks) - job.TasksCompleted - job.TasksRunning
 		}
 
+		if time.Now().Sub(job.TimeMarker) > LongJobDuration {
```
ryancouto, line 482: Looks like TimeMarker never gets updated. Is that the intent? If so, why not just use TimeCreated?

jasonbs10:
Awesome, good catch on the missing assignment, I'll fix & merge.

ryancouto (approved):

PR #341, opened 2018-01-03
GetOperation stub for executionServer

ryancouto:


bazel/execution/service.go
```diff
@@ -144,3 +144,53 @@ func execReqToScoot(req *remoteexecution.ExecuteRequest) (result sched.JobDefini
...
+
+	// Marshal ExecuteActionMetadata to protobuf.Any format
+	eomAsPBAny, err := ptypes.MarshalAny(&eom)
+	if err != nil {
+		log.Errorf("Failed to marshal ExecuteOperationMetadata: %s", err)
+		return nil, fmt.Errorf("Failed to marshal ExecuteOperationMetadata as ptypes/any.Any: %s", err)
+	}
+	op.Metadata = eomAsPBAny
+
+	res := getOperationResponse{}
```
dgdonut: Would add one more TODO about moving some of this boilerplate response struct generation into a function later

bazel/execution/service.go
```diff
@@ -144,3 +144,53 @@ func execReqToScoot(req *remoteexecution.ExecuteRequest) (result sched.JobDefini
...
+		return nil, fmt.Errorf("Failed to marshal GetOperationResponse as ptypes/any.Any: %s", err)
+	}
+
+	log.Info("GetOperationRequest completed successfully")
+	// Include the response message in the longrunning operation message
+	op.Result = &longrunning.Operation_Response{Response: resAsPBAny}
+	return &op, nil
+}
+
+type getOperationResponse struct {
```
dgdonut: what's this for?

dgdonut (approved):

bazel/execution/service.go
ryancouto: just to explicitly differentiate responses dependent on request, it isn't necessary at all.

PR #344, opened 2018-01-17
Runner fetches Command data from CAS for tasks from Bazel executerequests

dgdonut:
To support this, these structural changes were made:
* Runners support multiple RunTypes (Scoot and Bazel). Runners now keep track of a map to Filers of supported RunTypes. These are set up in main.go's/ICE. It is up to the runner to decide how a runner.Command is of a certain RunType - for now we look at the SnapshotID.
* The filerMap maps to Filers and InitDoneChs (concept taken from closed source workerserver/main.go), and the QueueController can handle Init and Update of the multiple filers (even if in practice we don't expect to use this for bazel)

BzFiler (snapshot.Filer implementation for bazel requests) changed to:
* expose ServerAddr, which identifies the CAS, at top level. required for bazel requests to work.

Invoker detects bazel-type requests and does required preprocessing:
* Fetch the remote exec Command data (where the real Argv & Env live) for the runner.cmd prior to exec'ing it. Tasks started from bazel requests include a digest of the Command data, and the invoker makes a CAS Read request to fetch it. This is done with a CAS read library function as fs_util is designed mainly to handle file operations (and Scoot may eventually want its own CAS functions).

Note:
* Because of a recent /vendor update, you will have to regenerate your dependencies


bazel/cas/client.go
```diff
@@ -0,0 +1,68 @@
...
+// Read data as bytes from a CAS. Takes address in "host:port" format and a bazel Digest to read.
+// Scoot only supports CAS API over a network grpc interface (i.e. an apiserver)
+func ByteStreamRead(serverAddr string, d *remoteexecution.Digest) ([]byte, error) {
+	cc, err := grpc.DialContext(context.Background(), serverAddr)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to dial server %s: %s", serverAddr, err)
+	}
+	defer cc.Close()
+
+	rname, err := GetReadResourceName("", d.GetHash(), d.GetSizeBytes(), "")
```
ryancouto: Would be nice to have a GetReadResourceNameDefault that doesn't take an instance or fname

bazel/cas/client.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return nil, err
+	}
+	req := &bytestream.ReadRequest{
+		ResourceName: rname,
+		ReadOffset:   0,
```
ryancouto, line 32: Don't need to specify, 0 will be the default value. Think a constant would be good here though

bazel/cas/client.go
```diff
@@ -0,0 +1,68 @@
...
+	if req.ReadOffset != 0 || req.ReadLimit <= 0 {
+		return nil, fmt.Errorf("Unsupported ReadRequest - Offset must be 0, Limit must be known > 0")
+	}
+
+	rc, err := bsc.Read(context.Background(), req)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to make Read request: %s", err)
+	}
+
+	// Recv from seerver until Limit reached
```
ryancouto: *server

bazel/cas/client.go
```diff
@@ -0,0 +1,68 @@
...
+
+// CAS Client APIs
+// These are more straightforward than the server APIs - bytestream provides
+// the majority of the CAS Client implementation. We provide wrappers as
+// higher-level operations.
+
+// Read data as bytes from a CAS. Takes address in "host:port" format and a bazel Digest to read.
+// Scoot only supports CAS API over a network grpc interface (i.e. an apiserver)
+func ByteStreamRead(serverAddr string, d *remoteexecution.Digest) ([]byte, error) {
+	cc, err := grpc.DialContext(context.Background(), serverAddr)
```
ryancouto: It looks like we aren't doing anything with context.Background(), why not use grpc.Dial(serverAddr)?

bazel/cas/client_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"google.golang.org/genproto/googleapis/bytestream"
+	remoteexecution "google.golang.org/genproto/googleapis/devtools/remoteexecution/v1test"
+
+	"github.com/twitter/scoot/bazel"
+	"github.com/twitter/scoot/bazel/cas/mock_bytestream"
+	"github.com/twitter/scoot/snapshot/store"
+)
+
+func TestClientRead(t *testing.T) {
+	f := &store.FakeStore{}
```
jasonbs10: It's not immediately clear what this is used for.

dgdonut: this was useless; removed

bazel/cas/client.go
dgdonut: yep, done

bazel/cas/client.go
dgdonut: done

dgdonut:
Other updates:
* got rid of the store.StoreAddr and split logic in workerserver/main.go - I was trying to reuse the http_addr, but needed separate functionality to get the grpc_addr
* added a tool for posting data to CAS - will include in readme an example of how this is used in end-to-end testing
* generally tested working to the limitations of fs_util. will update some error messages/logs

dgdonut:
This should be considered 'done', requesting further review or approval

snapshot/bazel/setup.go
```diff
@@ -13,7 +14,7 @@ func Module() ice.Module {
 func (m module) Install(b *ice.MagicBag) {
 	b.PutMany(
 		func() *BzFiler {
-			return MakeBzFilerWithOptions()
+			return MakeBzFiler(dialer.NewConstantResolver(""))
```
jasonbs10: Doesn't seem like a useful default.

bazel/README.md
```diff
@@ -79,3 +79,26 @@ response {
...
+
+```sh
+$GOPATH/bin/setup-cloud-scoot --strategy local.local
+
+bzutil -cas_addr="localhost:12100" -args="sleep 10" -env="scootdir=pwd"
+INFO[0000] Using argv: ["sleep" "10"] env: map[scootdir:pwd]  file:line="bzutil/main.go:44"
+INFO[0000] Wrote to CAS successfully                     file:line="bzutil/main.go:73"
+00c1d37dc9582a81e9b4ee2ab0f90e51f28044cd2401507927a0d4ef36310989 28
+
+grpc_cli call localhost:12100 google.bytestream.ByteStream.Write "resource_name: 'uploads/123e4567-e89b-12d3-a456-426655440000/blobs/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855/0', write_offset: 0, finish_write: true, data: ''"
```
jasonbs10: Does this actually accomplish #3 above?

runner/runners/bazel.go
```diff
@@ -0,0 +1,50 @@
...
+// Get the Bazel Command from the embedded ExecuteRequest digest,
+// and populate the runner.Command's Argv and Env fields
+func fetchBazelCommand(f snapshot.Filer, cmd *runner.Command) error {
+	if cmd.ExecuteRequest == nil {
+		return fmt.Errorf("Command has no ExecuteRequest data")
+	}
+	digest := cmd.ExecuteRequest.Request.GetAction().GetCommandDigest()
+
+	// Read command data from CAS refernced by filer
+	bzFiler, ok := f.(*bzsnapshot.BzFiler)
```
jasonbs10: nit - should this happen in the caller instead?

runner/runners/bazel.go
```diff
@@ -0,0 +1,50 @@
...
+		log.Errorf("Error reading command data from CAS at %s: %s", bzFiler.ServerAddr, err)
+		return err
+	}
+	bzCommand := &remoteexecution.Command{}
+	if err = proto.Unmarshal(bzCommandBytes, bzCommand); err != nil {
+		return fmt.Errorf("Failed to unmarshal bytes as remoteexecution.Command: %s", err)
+	}
+
+	// Update cmd's Argv and EnvVars (overwrite) from Command
+	cmd.Argv = bzCommand.GetArguments()
```
jasonbs10, line 42: How does the rest of cmd get filled, ex: priority?

bazel/cas/client.go
```diff
@@ -0,0 +1,120 @@
...
+	"google.golang.org/genproto/googleapis/bytestream"
+	remoteexecution "google.golang.org/genproto/googleapis/devtools/remoteexecution/v1test"
+	"google.golang.org/grpc"
+)
+
+// CAS Client APIs
+// These are more straightforward than the server APIs - bytestream provides
+// the majority of the CAS Client implementation. We provide wrappers as
+// higher-level operations.
+// Scoot only supports CAS API over a network grpc interface (i.e. an apiserver)
```
jasonbs10: Is there another method that users would expect?

runner/runners/invoke.go
```diff
@@ -78,6 +77,27 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 	var co snapshot.Checkout
 	checkoutCh := make(chan error)
 
+	// Determine RunType from Command SnapshotID
+	// This invoker supports RunTypeScoot and RunTypeBazel
+	var runType runner.RunType
+	if err := bazel.ValidateID(cmd.SnapshotID); err == nil {
+		runType = runner.RunTypeBazel
+	} else {
+		runType = runner.RunTypeScoot
```
ryancouto, line 87: I wonder if it would make sense to also validate the ID in the case of a standard Scoot snapshot

runner/runners/queue.go
```diff
@@ -85,36 +86,58 @@ func NewQueueRunner(
...
+				if err != nil {
+					initErr = err
+					log.Errorf("Init channel for filer %v returned error: %s", rt, err)
+				}
+				wg.Done()
+			}(t, f.IDC)
+		}
+	}
+
+	if wait {
```
ryancouto, line 126: I think you can just call wg.Wait(), and it will immediately pass if it was never added to (i.e. if wait was never set to true)

runner/runners/queue.go
```diff
@@ -85,36 +86,58 @@ func NewQueueRunner(
...
+		// Wait for initialization to complete in a new goroutine to let this constructor return
 		go func() {
-			err = <-idc
-			if err != nil {
+			wg.Wait()
+			if initErr != nil {
```
ryancouto, line 130: Will this work if we have multiple filers with IDC's? E.g. the first x to return error, and the last to return is successful, then I think initErr will be nil.

dgdonut: initErr only gets set if the IDC returned a non-nil error, so although it can get overwritten, it can't get overwritten with nil (and we log all the individual failures)

runner/runners/queue.go
dgdonut: This has to be broken up because of the way some of our tests and implementations are - they expect if no IDC was given the controller will be available immediately after the constructor returns (this will fail race detector tests). We could make those implementations and/or tests cleaner to avoid this

snapshot/bazel/fs_util_test.go
```diff
@@ -10,29 +10,26 @@ import (
 
 	"github.com/twitter/scoot/bazel"
 	"github.com/twitter/scoot/common/log/hooks"
-	"github.com/twitter/scoot/os/temp"
 )
 
+// uses test vars and setup/teardown defined un filer_test.go
```
ryancouto: *in

ryancouto (approved):

runner/runners/queue.go
ryancouto: ah yep you're right - missed that

snapshot/bazel/filer.go
```diff
@@ -1,50 +1,48 @@
...
 		CASResolver:   r,
+		updater:       snapshots.MakeNoopUpdater(),
 	}
+	return bf, nil
 }
 
-// Satisfies snapshot.Checkouter, snapshot.Ingester, and snapshot.Updater
-// Default command is fs_util, a tool provided by github.com/pantsbuild/pants which
+// Implements Snapshot interface (snapshot.Checkouter, snapshot.Ingester, and snapshot.Updater)
+// Default runner is fs_util, a tool provided by github.com/pantsbuild/pants which
```
jasonbs10: Truncated

bazel/cas/resource.go
```diff
@@ -27,6 +27,26 @@ func (r *Resource) String() string {
...
 }
 
+// Return a valid read resource string based on individual components. Errors on invalid inputs
+func GetReadResourceName(instance, hash string, size int64, fname string) (string, error) {
+	rname := ""
+	if instance != "" {
+		rname += fmt.Sprintf("%s/", instance)
+	}
+	rname += fmt.Sprintf("%s/%s/%d", ResourceNameType, hash, size)
+	if fname != "" {
```
jasonbs10, line 38: Remind me, what is fname for and why is it optional?

binaries/bzutil/main.go
```diff
@@ -0,0 +1,75 @@
+package main
+
+// Utility functions related to Bazel operations in Scoot.
```
jasonbs10: This doesn't seem like an accurate description.

snapshot/bazel/filer.go
```diff
@@ -1,44 +1,65 @@
...
 // handles underlying implementation of bazel snapshot functionality
-type bzFiler struct {
-	command       bzRunner
-	updater       snapshot.Updater
+type BzFiler struct {
+	runner bzRunner
```
jasonbs10: A little late for this comment, but why is this considered a runner?

runner/runners/queue.go
```diff
@@ -326,7 +379,9 @@ func (c *QueueController) loop() {
 			// Handle run and abort requests.
 			if !ok {
 				c.reqCh = nil
-				c.cancelTimerCh <- nil
+				for i := 0; i < len(c.filerMap); i++ {
+					c.cancelTimerCh <- nil
```
jasonbs10: Can't you just close the ch once here?

runner/runners/queue.go
```diff
@@ -123,16 +146,18 @@ func NewQueueRunner(
...
 
 // QueueController maintains a queue of commands to run (up to capacity).
 // Manages updates to underlying Filer via Filer's Update interface,
 // if a non-zero update interval is defined (updates and tasks cannot run concurrently)
 type QueueController struct {
 	inv           *Invoker
-	filer         snapshot.Filer
+	filerMap      runner.RunTypeMap
+	updateReq     map[runner.RunType]bool
+	updateLock    sync.Mutex
```
jasonbs10, line 160: All this locking stuff seems unwieldy, I can't easily tell if it's necessary. It seems like you can just use yet another channel and send it the run type every time the respective ticker updates.

jasonbs10 (approved):
Looks like mostly plumbing and docs, so it seems ok as long as you're satisfied with the testing.

bazel/README.md
dgdonut: yes, the hand of grpc/protobuf is merciful and kind. I might update this later to data: ' ' (nothing w/ length 1) as it still works, and certain underlying internal blob stores do not accept 0 length

bazel/cas/client.go
dgdonut: removing. I guess I had in mind some of the other scoot semantics for Stores that try to transparently support local alternatives

bazel/cas/resource.go
dgdonut: part of the Bazel API spec - described in remote_execution.proto in the description of the ContentAddressableStore service. there is a fair amount of subtlety described in the API this way  https://github.com/googleapis/googleapis/blob/master/google/devtools/remoteexecution/v1test/remote_execution.proto

binaries/bzutil/main.go
dgdonut: updating

runner/runners/bazel.go
dgdonut: The handler in bazel/execution/service.go:execReqToScoot does this. The only things unavailable at that level are the argv/env vars. Adding comments to execReqToScoot

runner/runners/bazel.go
dgdonut: makes sense, will move

bazel/cas/resource.go
jasonbs10: Ugh, I'm not a fan of this api. Could you elaborate a little in the github comments about our interpretation of: 

"the client MAY reuse the `uuid` for uploading
// different blobs. The `resource_name` may optionally have a trailing filename
// (or other metadata) for a client to use if it is storing URLs, as in
// `{instance}/uploads/{uuid}/blobs/{hash}/{size}/foo/bar/baz.cc`"

First of all wtf would you reuse a uuid for a different blob, and what does the server do with the URL fragment? Is it just there for logging purposes?

snapshot/bazel/filer.go
dgdonut: runs underlying bazel operations... I'm open to other suggestions here, this would be easy to change. BzTreeer? BzFs? etc

bazel/README.md
illicitonion: FWIW as of https://github.com/pantsbuild/pants/pull/5378

`fs_util --local-store-path=/Users/dwagnerhall/tmp/store --server-address=localhost:12100 directory save --root=/ ''`

should work. I can put together a pre-release binary if you want to try it out without waiting for the pants release on Friday :)

runner/runners/queue.go
dgdonut: We start N updateTickers in separate goroutines, not sure how if we recv on one the others will get the message?

runner/runners/queue.go
dgdonut: I'm using the lock only when tickers update the map of bool flags that indicate what filers have update requests, and then once in the loop to read & reset the flags.  To queue these up on a channel it would have to be buffered -- to what length? how do updateTickers know if they have a request sitting on the channel or not without draining it? didn't see a clear way of doing that

bazel/README.md
dgdonut: oh, by certain underlying stores I mean Blobstore :) gives us an HTTP 411 for doing that

bazel/cas/resource.go
dgdonut: Yeah, I don't totally understand the point of the fname fragments as there is no prescribed behavior on how to use it on the server - perhaps just for logging as you indicated. Will add a comment on uuid

runner/runners/queue.go
jasonbs10: Why does it have to be buffered? They end up being processed serially right? Seems like we can do a blocking send and ultimately it doesn't matter if it gets backed up or not, although I'd expect it shouldn't with reasonable ticker times.

runner/runners/queue.go
jasonbs10: All the goroutines will receive "_, notOK := <-ch" when a channel is closed afaik.

snapshot/bazel/filer.go
jasonbs10: Seems confusing to reuse the term runner, anything else would be better as long as it's unique or thematically consistent.

runner/runners/queue.go
dgdonut: Ah yes - thanks

runner/runners/queue.go
dgdonut: I think an updater could get starved out in this case, if the request channel has a request and before it's drained by the updater (some task is running and taking a while), several updaters hit their tick again and compete to make the next request. Possible with sane values

jasonbs10: True. In that case I'd use a single ticker with the gcd of filer intervals and then on each tick loop through the filers seeing who all deserves a turn.

jasonbs10: Up to  you though, whatever you think will be the most understandable.

PR #346, opened 2018-01-22
log updates

JeanetteBruno:
- added job id to long runnning job log entry

- removed : in file:line tag - can't search for file:line tag in loglens

- updated cluster_state node logging to use String() to avoid getting object address


sched/scheduler/cluster_state.go
```diff
@@ -225,15 +224,17 @@ func (c *clusterState) update(updates []cluster.NodeUpdate) {
 			if ns, ok := c.suspendedNodes[update.Id]; ok {
 				if !ns.ready() {
 					// Adding a node that's already suspended as non-ready, leave it in that state until ready.
-					log.Infof("Suspended node re-added but still awaiting readiness check %v (%#v)", update.Id, ns)
+					log.Infof("Suspended node re-added but still awaiting readiness check %v (%s)", update.Id,
+						ns.node.String())
```
ryancouto: I think you can just have ns.node here, the %s calls String() on it

dgdonut (approved):

sched/scheduler/cluster_state.go
```diff
@@ -225,15 +224,17 @@ func (c *clusterState) update(updates []cluster.NodeUpdate) {
...
 				} else if ns.timeLost != nilTime {
 					// This node was suspended as lost earlier, we can recover it now.
 					ns.timeLost = nilTime
 					c.nodes[update.Id] = ns
 					delete(c.suspendedNodes, update.Id)
-					log.Infof("Recovered suspended node %v (%#v), %s", update.Id, ns, c.status())
+					log.Infof("Recovered suspended node %v (%s), %s", update.Id, ns.node.String(), c.status())
```
jasonbs10: It's not clear to me - why did you change from logging NodeState to logging Node?  We're losing a lot of useful information by doing this.

jasonbs10 (changes requested):

sched/scheduler/cluster_state.go
```diff
@@ -311,10 +313,10 @@ func (c *clusterState) update(updates []cluster.NodeUpdate) {
 			ns.timeFlaky = nilTime
 			ns.snapshotId = ""
 			if c.readyFn != nil {
-				log.Infof("Reinstating flaky node momentarily: %v (%#v), %s", ns.node.Id(), ns, c.status())
+				log.Infof("Reinstating flaky node momentarily: %v (%s), %s", ns.node.Id(), ns, c.status())
 				ns.startReadyLoop(c.readyFn)
 			} else {
-				log.Infof("Reinstating flaky node now: %v (%#v), %s", ns.node.Id(), ns, c.status())
+				log.Infof("Reinstating flaky node now: %v (%s), %s", ns.node.Id(), ns, c.status())
```
ryancouto, line 318: can just have ns.node here, %s calls String() which calls Id()

sched/scheduler/cluster_state.go
```diff
@@ -311,10 +313,10 @@ func (c *clusterState) update(updates []cluster.NodeUpdate) {
 			ns.timeFlaky = nilTime
 			ns.snapshotId = ""
 			if c.readyFn != nil {
-				log.Infof("Reinstating flaky node momentarily: %v (%#v), %s", ns.node.Id(), ns, c.status())
+				log.Infof("Reinstating flaky node momentarily: %v (%s), %s", ns.node.Id(), ns, c.status())
```
ryancouto, line 315: can just have ns.node here, %s calls String() which calls Id()

sched/scheduler/cluster_state.go
```diff
@@ -290,13 +292,13 @@ func (c *clusterState) update(updates []cluster.NodeUpdate) {
...
 			delete(c.suspendedNodes, ns.node.Id())
-			log.Infof("Node now ready, adding to rotation: %v (%#v), %s", ns.node.Id(), ns, c.status())
+			log.Infof("Node now ready, adding to rotation: %v (%s), %s", ns.node.Id(), ns, c.status())
 		} else if ns.timeLost != nilTime && now.Sub(ns.timeLost) > c.maxLostDuration {
 			// This node has been missing too long, delete all references to it.
 			delete(c.suspendedNodes, ns.node.Id())
 			delete(c.nodeGroups[ns.snapshotId].idle, ns.node.Id())
 			delete(c.nodeGroups[ns.snapshotId].busy, ns.node.Id())
-			log.Infof("Deleting lost node: %v (%#v), %s", ns.node.Id(), ns, c.status())
+			log.Infof("Deleting lost node: %v (%s), %s", ns.node.Id(), ns, c.status())
```
ryancouto, line 300: can just have ns.node here, %s calls String() which calls Id()

sched/scheduler/cluster_state.go
```diff
@@ -290,13 +292,13 @@ func (c *clusterState) update(updates []cluster.NodeUpdate) {
 			// This node is initialized, remove it from suspended nodes and add it to the healthy node pool.
 			c.nodes[ns.node.Id()] = ns
 			delete(c.suspendedNodes, ns.node.Id())
-			log.Infof("Node now ready, adding to rotation: %v (%#v), %s", ns.node.Id(), ns, c.status())
+			log.Infof("Node now ready, adding to rotation: %v (%s), %s", ns.node.Id(), ns, c.status())
```
ryancouto, line 294: can just have ns.node here, %s calls String() which calls Id()

JeanetteBruno: fixed and reverted to printing ns (all of nodeState)

sched/scheduler/cluster_state.go
JeanetteBruno: fixed and reverted to printing ns (all of nodeState)

sched/scheduler/cluster_state.go
JeanetteBruno: fixed and reverted to printing ns (all of nodeState)

sched/scheduler/cluster_state.go
JeanetteBruno: fixed and reverted to printing ns (all of nodeState)

sched/scheduler/cluster_state.go
JeanetteBruno: fixed - was trying to get logging to print the wily node content instead of its address

sched/scheduler/cluster_state.go
JeanetteBruno: fixed and reverted to using ns

JeanetteBruno (commented):
logging now prints wily node content as well as nodeState info minus the addresses for readyCh and removedCh.  If we think we want these addresses (basically to know that they were not null let me know.
See https://loglens_ui-devel-loglens.service.smf1.twitter.biz/search?indexName=scoot_scheduler_devel&dateRange=14400&searchText=%22node%22&pageSize=50&dummyTs=56 - logging statements around 1/24 18:00-18:10 UTC for new log messages.

sched/scheduler/cluster_state.go
```diff
@@ -225,15 +226,15 @@ func (c *clusterState) update(updates []cluster.NodeUpdate) {
 			if ns, ok := c.suspendedNodes[update.Id]; ok {
 				if !ns.ready() {
 					// Adding a node that's already suspended as non-ready, leave it in that state until ready.
-					log.Infof("Suspended node re-added but still awaiting readiness check %v (%#v)", update.Id, ns)
+					log.Infof("Suspended node re-added but still awaiting readiness check %v (%s)", update.Id, ns)
```
jasonbs10, line 229: And what was the reason for going from #%v to %s? If I remember correctly, I changed this so it would print the field names rather than just bare values.

JeanetteBruno: using %s trigger using ns.String() which prints the host name, shard id as well as the IP address from update.Node.  ns.String() prints field name as well as values.  The big difference is that it does not print readyCh nor removedCh.
You can see the differences if you compare devel vs prod loglens entries (for 1/24 noon to 1pm'ish ET)

jasonbs10 (approved):

PR #347, opened 2018-01-29
Jschiller/combined out err

jasonbs10:
Combined stdout/err has been requested multiple times over the past few months and it was easy to do, so this makes 'stdlog' available without actually changing the protocol etc.


dgdonut (approved):

kwlzn (approved):
sa-weet. thanks!

PR #348, opened 2018-01-30
runner/bazel - support for ingestion of std* logs and Output data for Bazel commands

dgdonut:
* some minor reorgs.

Added some split post-process ingestion for the Invoker, based on RunType. For Commands that are invoked that are of RunTypeBazel, we ingest output Bazel-style (according to the API). This means:
* stdout and stderr uploaded to CAS directly
* OutputFiles/Dirs are specified in the ExecuteRequest data. We scan the checkout dir after Argv is run and upload these to CAS (via BzFiler.Ingest) if found.
* Bazel Digests (functioning as lookup keys for the data in the CAS) for stdout, stderr, and any Ingested outputs are returned in the runner.Status's ActionResult field

* Some end to end testing is pending based on fs_util's ability to upload to CAS, but this is non-blocking

### DIFF
--- a/runner/runners/bazel.go
+++ b/runner/runners/bazel.go
@@ -6,31 +6,48 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
 
 	"github.com/golang/protobuf/proto"
 	log "github.com/sirupsen/logrus"
 	remoteexecution "google.golang.org/genproto/googleapis/devtools/remoteexecution/v1test"
 
+	"github.com/twitter/scoot/bazel"
 	"github.com/twitter/scoot/bazel/cas"
 	"github.com/twitter/scoot/bazel/execution/bazelapi"
 	"github.com/twitter/scoot/runner"
+	"github.com/twitter/scoot/runner/execer"
+	"github.com/twitter/scoot/snapshot"
 	bzsnapshot "github.com/twitter/scoot/snapshot/bazel"
 )
+
+// Preliminary setup for handling Bazel commands - verify Filer and populate command
+func preProcessBazel(filer snapshot.Filer, cmd *runner.Command) error {
+	bzFiler, ok := filer.(*bzsnapshot.BzFiler)
+	if !ok {
+		return fmt.Errorf("Filer could not be asserted as type BzFiler")
+	}
+	if cmd.ExecuteRequest == nil {
+		return fmt.Errorf("Nil ExecuteRequest data in Command with RunType Bazel")
+	}
+	if err := fetchBazelCommand(bzFiler, cmd); err != nil {
+		return fmt.Errorf("Error retrieving Bazel command data: %v", err)
+	}
+	log.Infof("Worker running with updated command arguments: %q", cmd.Argv)
+	return nil
+}
 
 // Get the Bazel Command from the embedded ExecuteRequest digest,
 // and populate the runner.Command's Argv and Env fields
 func fetchBazelCommand(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command) error {
 	digest := cmd.ExecuteRequest.Request.GetAction().GetCommandDigest()
 
-	casAddr, err := bzFiler.CASResolver.Resolve()
+	log.Info("Fetching Bazel Command data from CAS server")
+	bzCommandBytes, err := cas.ByteStreamRead(bzFiler.CASResolver, digest)
 	if err != nil {
-		return fmt.Errorf("Filer could not resolve a CAS server: %s", err)
-	}
-	log.Infof("Fetching Bazel Command data from CAS server %s", casAddr)
-
-	bzCommandBytes, err := cas.ByteStreamRead(casAddr, digest)
-	if err != nil {
-		log.Errorf("Error reading command data from CAS server %s: %s", casAddr, err)
+		log.Errorf("Error reading command data from CAS server: %s", err)
 		return err
 	}
 	bzCommand := &remoteexecution.Command{}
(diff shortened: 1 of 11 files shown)


runner/runners/bazel.go
```diff
@@ -46,3 +63,155 @@ func fetchBazelCommand(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command) error {
...
+
+// Post-execer actions for Bazel tasks - upload outputs and std* logs, format result structure
+func postProcessBazel(filer snapshot.Filer,
+	cmd *runner.Command,
+	coDir string,
+	stdout, stderr runner.Output,
+	st execer.ProcessStatus) (*bazelapi.ActionResult, error) {
+	bzFiler, ok := filer.(*bzsnapshot.BzFiler)
+	if !ok {
+		return nil, fmt.Errorf("Filer could not be asserted as type BzFiler")
```
JeanetteBruno: Can we print what type the filer was? 
fmt.Errorf("postProcessBazel() expected filer to be a BzFiler, instead it was %s", reflect.TypeOf(filer))

runner/runners/bazel.go
```diff
@@ -46,3 +63,155 @@ func fetchBazelCommand(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command) error {
...
+	}
+	log.Info("Processing Bazel outputs to CAS")
+
+	stdoutDigest, err := writeFileToCAS(bzFiler, stdout.AsFile())
+	if err != nil {
```
JeanetteBruno, line 81: I've been thinking about our error handling.... should we include the stack trace on the initial error generation,  these sorts of 'if err != nil {}' blocks simply return the error and have the top level that finally resolves the error print the stack trace?  (just a thought)

runner/runners/bazel.go
```diff
@@ -46,3 +63,155 @@ func fetchBazelCommand(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command) error {
...
+}
+
+// Write a file's bytes to the BzFiler's CAS, returning the Digest of the uploaded file.
+// This is distinct from using a Filer to Ingest data into the CAS,
+// which allows for Filer-implementation-specific behavior that could alter the bytes and expected digest.
+// (Our typical BzFiler use case is uploading Files that include relative path and other protobuf data)
+func writeFileToCAS(bzFiler *bzsnapshot.BzFiler, path string) (*remoteexecution.Digest, error) {
+	bytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("Error reading file %s as bytes: %s", path, err)
```
JeanetteBruno, line 116: continuing error handling comment from above.....
this is where we would put the stack trace on the error (see github.com/pkg/errors - already in vendors)

runner/runners/bazel.go
```diff
@@ -46,3 +63,155 @@ func fetchBazelCommand(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command) error {
...
+// Files located are Ingested in to the CAS via the BzFiler
+func ingestOutputFiles(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command, coDir string) ([]*remoteexecution.OutputFile, error) {
+	outputFiles := []*remoteexecution.OutputFile{}
+	for _, relPath := range cmd.ExecuteRequest.Request.GetAction().GetOutputFiles() {
+		relPath = path.Join("/", relPath)
+		absPath := filepath.Join(coDir, relPath)
+		info, err := os.Stat(absPath)
+		if err != nil {
+			if os.IsNotExist(err) {
+				log.Infof("Output file %s not present", absPath)
```
JeanetteBruno: put 'skipping' on the error message.

runner/runners/bazel.go
```diff
@@ -46,3 +63,155 @@ func fetchBazelCommand(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command) error {
...
+			} else {
+				return nil, fmt.Errorf("Error Statting output file %s: %s", absPath, err)
+			}
+			continue
+		}
+		if !info.Mode().IsRegular() {
+			return nil, fmt.Errorf("Expected output file %s not a regular file", absPath)
```
JeanetteBruno: put 'is' in the message: "Expected output file %s is not a regular file"

runner/runners/bazel.go
```diff
@@ -46,3 +63,155 @@ func fetchBazelCommand(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command) error {
...
+// Directories located are Ingested in to the CAS via the BzFiler
+func ingestOutputDirs(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command, coDir string) ([]*remoteexecution.OutputDirectory, error) {
+	outputDirs := []*remoteexecution.OutputDirectory{}
+	for _, relPath := range cmd.ExecuteRequest.Request.GetAction().GetOutputDirectories() {
+		relPath = path.Join("/", relPath)
+		absPath := filepath.Join(coDir, relPath)
+		info, err := os.Stat(absPath)
+		if err != nil {
+			if os.IsNotExist(err) {
+				log.Infof("Output dir %s not present", absPath)
```
JeanetteBruno: add 'skipping'

runner/runners/bazel.go
```diff
@@ -46,3 +63,155 @@ func fetchBazelCommand(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command) error {
...
+			} else {
+				return nil, fmt.Errorf("Error Statting output dir %s: %s", absPath, err)
+			}
+			continue
+		}
+		if !info.Mode().IsDir() {
+			return nil, fmt.Errorf("Expected output dir %s not a directory", absPath)
```
JeanetteBruno: add 'is'

runner/runners/invoke.go
```diff
@@ -113,9 +106,7 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 	}
 
 	go func() {
-		//FIXME(jschiller): allow aborts/timeouts to cancel the checkout process.
 		if cmd.SnapshotID == "" {
-			//TODO: we don't want this logic to live here, these decisions should be made at a higher level.
```
JeanetteBruno, line 118: do we really want to remove this TODO- it seems to be a reasonable comment to me.

runner/runners/invoke.go
```diff
@@ -283,98 +273,105 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
...
-				tags.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID, Tag: cmd.Tag})
-		} else if reader, err = os.Open(outPath); err != nil {
-			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stdout: %v", err),
-				tags.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID, Tag: cmd.Tag})
-		} else if _, err := io.Copy(writer, reader); err != nil {
-			return runner.FailedStatus(id, fmt.Errorf("error staging ingestion for stdout: %v", err),
-				tags.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID, Tag: cmd.Tag})
-		}
+		if runType == runner.RunTypeScoot {
+			tmp, err := inv.tmp.TempDir("invoke")
```
JeanetteBruno, line 278: this is a lot of code in this if block - can it be pulled into a well-named/documented function?

runner/runners/invoke.go
```diff
@@ -283,98 +273,105 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
...
-			snapshotID, err := inv.filerMap[runType].Filer.Ingest(tmp.Dir)
-			if err != nil {
-				ingestCh <- err
-			} else {
-				ingestCh <- snapshotID
+			// Meaningful to support Abort after execer has completed?
+			var snapshotID string
+			select {
+			case <-abortCh:
+				return runner.AbortStatus(id, tags.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID, Tag: cmd.Tag})
```
JeanetteBruno, line 312: just curious.... this doesn't really abort the ingest does it?

JeanetteBruno (approved):
My comments are more what you'd call 'guidelines' than actual change requests.

runner/runners/invoke.go
dgdonut: I don't feel these TODOs about design level decisions add anything. No one is going to notice this comment and then implement its suggestion, and todo's are not an issue tracking system. IMO todos should not be checked into master branch, but I found this one to be particularly useless (and todos littered all over the code add to cognitive load of reading/understanding)

runner/runners/invoke.go
dgdonut: It lets the worker go and do something else, but the actual ingestion keeps going (same as existing behavior)

runner/runners/invoke.go
dgdonut: Had thought about that. As-is I left the "Scoot" type behavior inline and broke out the Bazel processing. Will break some of this out into a post-process piece

runner/runners/invoke.go
jasonbs10: I agree and disagree. TODO's are useful to document things that are unintuitive, behaviorally or semantically, yet appear to work, and therefore don't warrant polluting jira. In this case, it's unlikely that anyone will ever be impacted, but you *did* notice the TODO and have the option of fixing it or deleting it while you're here.

PR #349, opened 2018-02-06
bazel/execution: GetOperation requests have valid responses

ryancouto:
GetOperation requests return valid data populated by the correct internal Scoot and remote execution job definition fields, and error cases are handled per the Bazel API spec.


bazel/execution/service.go
```diff
@@ -181,25 +186,65 @@ func validateExecRequest(req *remoteexecution.ExecuteRequest) error {
...
+	_ context.Context,
 	req *longrunning.GetOperationRequest) (*longrunning.Operation, error) {
 	log.Infof("Received GetOperation request: %s", req)
 
 	if !s.IsInitialized() {
 		return nil, status.Error(codes.Internal, "Server not initialized")
 	}
 
-	// TODO get status of job identified by req.Name
+	// TODO(rcouto): This skips over the stats handler increment in github.com/twitter/scoot/scootapi/server/server.go
```
dgdonut: I think this is fine w/out the note - metrics is a bigger todo tracked internally

bazel/execution/service.go
```diff
@@ -181,25 +186,65 @@ func validateExecRequest(req *remoteexecution.ExecuteRequest) error {
...
+	// Do we want to increment that counter? A different counter? Both?
+	js, err := api.GetJobStatus(req.Name, s.sagaCoord)
+	if err != nil {
+		return nil, err
```
dgdonut: For all the errors in these top-level api-defined function, return errors only as  grpc/status.Error(<grpc/codes.Code>, "<user-digestable message>") since these are client-facing. see other return examples. there might be a few returns that were in here already that also need to be fixed

bazel/execution/service.go
```diff
@@ -181,25 +186,65 @@ func validateExecRequest(req *remoteexecution.ExecuteRequest) error {
...
+		br = scoot.RunStatus_BazelResult__DEFAULT
+	} else {
+		br = runStatus.GetBazelResult_()
+	}
+	actionResult := bazelapi.MakeActionResultDomainFromThrift(br)
+
 	eom := &remoteexecution.ExecuteOperationMetadata{
-		Stage: remoteexecution.ExecuteOperationMetadata_COMPLETED,
+		Stage: bazelapi.RunStatusToExecuteOperationMetadata_Stage(runStatus),
 	}
```
dgdonut: Add Digest of the action here - I think that's the only additional EOM data required. This might be a good candidate for adding a field to the bazelapi wrapper data struct (record once in the Execute function when we initially calculate the ActionDigest and don't recalculate on each GetOperation request), if this can be passed around easily

bazel/execution/service.go
```diff
@@ -181,25 +186,65 @@ func validateExecRequest(req *remoteexecution.ExecuteRequest) error {
...
+	}
+	log.Info("Received job status %s", js)
 
 	op := longrunning.Operation{}
-	op.Name = req.Name
+	op.Name = fmt.Sprintf("operations/%s", js.ID)
 	op.Done = true
 
+	// There should only be one task per job, so runStatus/status is the
+	// first & last one we encounter when looping through js.TaskData/TaskStatus
```
dgdonut: Might want to move this validation and retrieval of runStatus to small function

dgdonut (changes requested):
Because I can't add in-line comment on unchanged lines: on line 111 in the Execute() func, change the initial EOM Stage state to QUEUED.

We should also leave the longrunning.Operation:Response field unset/nil in the Execute request, and ensure the operation's Done field is set to false. In GetOperation, we should do something similar - only set Done based on the status of the job. When Done is false, we are not required to return an Operation error nor response, and we should leave this as nil (no ExecuteResponse contained whatsoever).

In the Done case we should also be setting the Status field in the ER per the API - grpc/codes.OK or some error (probably codes.INTERNAL in most scoot failed cases)

dgdonut:
Sorry, I'm tacking on some stuff here that wasn't exactly part of the ticket, but these changes should get us to "V1" completion.

bazel/execution/service.go
```diff
@@ -98,141 +98,120 @@ func (s *executionServer) Execute(
...
+}
+
+func (s *executionServer) CancelOperation(context.Context, *longrunning.CancelOperationRequest) (*empty.Empty, error) {
+	return nil, status.Error(codes.Internal, fmt.Sprint("Not implemented"))
+}
+
+func (s *executionServer) getRunStatusAndValidate(jobID string) (*runStatus, error) {
+	js, err := api.GetJobStatus(jobID, s.sagaCoord)
+	if err != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("Error getting job status: %s", err))
```
dgdonut: should be fine with 'normal' errors here since top-level is wrapping into a status.Error

dgdonut (approved):
🥇

PR #352, opened 2018-02-12
Include directory size_bytes in fs_util directory materialize command

illicitonion:
It is required as of recently

### DIFF
--- a/get_fs_util.sh
+++ b/get_fs_util.sh
@@ -2,13 +2,13 @@
 # Pulls in fs_util tool for use in remote exec ingestion, checkout, and CAS ops
 set -e
 
-pants_release="1.4.0.dev23"
+pants_release="1.5.0.dev3%2B569f14c2"
 
 get_fs_util() {
 	case "$(uname -s)" in
 
 		Darwin)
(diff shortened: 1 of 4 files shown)


dgdonut (approved):

get_fs_util.sh
```diff
@@ -2,13 +2,13 @@
 # Pulls in fs_util tool for use in remote exec ingestion, checkout, and CAS ops
 set -e
 
-pants_release="1.4.0.dev23"
+pants_release="1.5.0.dev3%2B569f14c2"
```
dgdonut, line 5: just checking - supposed to be a "%2B" here or "+"?
edit: think it needs to be escaped, good as-is

illicitonion: Technically it's a `+` in the version number, but it needs to be `%2B` in the curl command line. This seemed nicer than `pants_release_urlencoded="$(perl -MURI::Escape -e "print uri_escape('${pants_release}');")"` and putting that in the `curl`s, but I can do that if you'd prefer :)

ryancouto (approved):

PR #355, opened 2018-02-14
Potential solution for apiserver memory spikes.

jasonbs10:


snapshot/bundlestore/http_server.go
```diff
@@ -100,13 +100,12 @@ func (s *httpServer) HandleDownload(w http.ResponseWriter, req *http.Request) {
 	}
 	if _, err := io.Copy(w, r); err != nil {
 		log.Infof("Copy err: %v --> StatusInternalServerError (from %v)", err, req.RemoteAddr)
-		http.Error(w, fmt.Sprintf("Error copying Bundle: %s", err), http.StatusInternalServerError)
```
jasonbs10, line 103: This was resulting in double writes to the response header, which isn't allowed. (io.Copy() calls Write() which sets status=200=OK as the final header content).

JeanetteBruno (approved):

ryancouto (approved):

PR #360, opened 2018-02-22
Implemented killing of previous jobs based on basis string.

jasonbs10:


sched/scheduler/stateful_scheduler.go
```diff
@@ -578,26 +576,33 @@ addLoop:
...
-						log.WithFields(
-							log.Fields{
-								"requestor": m.job.Def.Requestor,
-								"jobType":   m.job.Def.JobType,
-								"tag":       m.job.Def.Tag,
-								"basis":     m.job.Def.Basis,
-								"priority":  m.job.Def.Priority,
-								"numTasks":  len(m.job.Def.Tasks),
-							}).Info("Overriding job priority to match previous requestor/tag priority")
+						log.WithFields(lf).Info("Overriding job priority to match previous requestor/tag priority")
```
dgdonut, line 634: was the 'break' a bug before?

dgdonut (approved):

sched/scheduler/stateful_scheduler.go
jasonbs10: Since new retry jobs have to match previous jobs, would could exit early before, but now we want to unconditionally iterate across all jobs with the same requestor.

sched/scheduler/stateful_scheduler.go
```diff
@@ -578,26 +576,33 @@ addLoop:
...
+				"basis":     newJobMsg.job.Def.Basis,
+				"priority":  newJobMsg.job.Def.Priority,
+				"numTasks":  len(newJobMsg.job.Def.Tasks),
+			}
 			if _, ok := s.requestorMap[newJobMsg.job.Def.Requestor]; ok {
-				// If we have an existing job with this requestor/tag combination, make sure we use its priority level.
-				// Not an error since we can consider priority to be a suggestion which we'll handle contextually.
 				for _, js := range s.requestorMap[newJobMsg.job.Def.Requestor] {
+					// Kill any prior jobs that share the same requestor and non-empty basis string,
+					// but tag should be different so we don't kill the original jobs when doing retries.
```
ryancouto: Could you expand on this point? Do retries get assigned a non-unique "retry" tag?

Issue #353, opened 2018-02-12
Bazel gRPC API includes scoot header in stdout/stderr


illicitonion:
We should not add these prefixes to the stdout/stderr we report through the gRPC API:

```
$ cat /var/folders/1j/jzw90nc51nn_358mmhnwc_fh0000gn/T/scoot-tmp-988923338/common-bundles/blob-cf722969a46641061fbf9809045a0e10b15d8a9ba573a4e62296c1d96a669749.blob
###########################################
###########################################


Date: 2018-02-12 22:13:38.899871 +0000 GMT
Out: http://localhost:10102/output//2-stdout?file=file://tw-mbp-dwagnerhall/var/folders/1j/jzw90nc51nn_358mmhnwc_fh0000gn/T/scoot-tmp-445635530/output/2-stdout109126150    Err: http://localhost:10102/output//2-stderr?file=file://tw-mbp-dwagnerhall/var/folders/1j/jzw90nc51nn_358mmhnwc_fh0000gn/T/scoot-tmp-445635530/output/2-stderr209918381    OutErr: http://localhost:10102/output//2-stdlog?file=file://tw-mbp-dwagnerhall/var/folders/1j/jzw90nc51nn_358mmhnwc_fh0000gn/T/scoot-tmp-445635530/output/2-stdlog637639464    Cmd:
runner.Command -- SnapshotID: bz-e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855-0 # Argv: ["/bin/bash" "-c" "exit 42"] # Timeout: 30m0s # JobID: 878bbb7e-8e3a-4eae-5923-d438072b7f67 # TaskID: Bazel_ExecuteRequest_40e8c5bcece27be634101baca92efc1d654921688a0c1528deed9abbf6fac94e_1518473618 # Tag: 878bbb7e-8e3a-4eae-5923-d438072b7f67  ExecuteRequest=action:<command_digest:<hash:"8a1eb3dc8694551ee161e6f616de205f7c2d5b52c8f5b93d2745d67c3b750e6a" size_bytes:24 > input_root_digest:<hash:"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855" > platform:<> timeout:<> > 
<nil>

###########################################
###########################################



SCOOT_CMD_LOG
```

jasonbs10:
Seems like should be enough to have a flag optionally disable them. We still want to print something here for debugging purposes internally.

[...]

Issue #361, opened 2018-02-28
Bazel API output directories have inconsistent prefixes


illicitonion:
I ran this operation which expects output directory `dir`:
```
DEBUG 2018-02-28T13:49:24Z: process_execution::remote: Executing remotely request: action {command_digest {hash: "a4124e7e29e8f625473652fca900c6b8a89960316784ff31b54f784860278904" size_bytes: 71} input_root_digest {hash: "6f9a33e23a007c58c1e83f29bcf52979c797aba200cec0058695ff72d181e6ed" size_bytes: 85} output_directories: "dir"} (command: arguments: "/bin/bash" arguments: "-c" arguments: "/usr/bin/mkdir -p dir ; /usr/bin/echo 'foo' > dir/file")
```

This returned me:
```
DEBUG 2018-02-28T13:49:29Z: process_execution::remote: Got (nested) execute response: result {output_directories {path: "/dir" tree_digest {hash: "ebbc315a499c5aa2dc8f972a535486e0097a871279b275ec6cf3326ff1cb9fcc" size_bytes: 77}} stdout_digest {hash: "644a7d0ab7451b1fecc58a742b45fe720d5d6a9ea92c29d0e9e63f80f591e0ee" size_bytes: 2219} stderr_digest {hash: "644a7d0ab7451b1fecc58a742b45fe720d5d6a9ea92c29d0e9e63f80f591e0ee" size_bytes: 2219}} status {}
```

the `output_directories` entry is a directory containing `dir`:
```
$ ../fs/fs_util/target/debug/fs_util --local-store-path=${HOME}/tmp/store --server-address=localhost:50001 directory cat-proto --output-format=text ebbc315a499c5aa2dc8f972a535486e0097a871279b275ec6cf3326ff1cb9fcc 77
directories {name: "dir" digest {hash: "a8eeda526245a7d32d88e2f6cdf790a96990c2a77b575c12009e11975c62e35a" size_bytes: 78}}
```

I believe that this is one level too much nesting. If `path=/dir` then the root hash should be a directory containing `file`, not containing `dir/file`.

dgdonut:
Updated #362 to handle this per https://docs.google.com/document/d/12c3oAPgedckLpue2yj0xGgJTEOyUm4mXWWBJ157J-8I/edit#heading=h.sg3la2izb6yt

Example run based on scenario outlined in doc:
```sh
[...]

Issue #358, opened 2018-02-21
Bazel gRPC API does not report missing files


illicitonion:
If the input directory is not found, the Operation response I get from the scheduler is:

```name: "312a8f54-91a1-4ec9-598f-f435b2cdfd0a" metadata {type_url: "type.googleapis.com/google.devtools.remoteexecution.v1test.ExecuteOperationMetadata" value: "\010\004\022\000"} done: true response {type_url: "type.googleapis.com/google.devtools.remoteexecution.v1test.ExecuteResponse" value: "\n\000\032\002\010\001"}```

which nests the ExecuteResponse: ```result {} status {code: 1}```

Instead, it should populate a `FailedPrecondition` status as per the suggestion in the proto file: https://github.com/googleapis/googleapis/blob/eb2730ad8d88eb8e70ac647d1e91a3b41e4ca81c/google/devtools/remoteexecution/v1test/remote_execution.proto#L99

Otherwise clients can't know what to do to rectify the situation.

Ideally the `command_digest` missing would have the same effect.

dgdonut:
Note that as-is, the fix that is in progress for this will only report one of two things: the Command is missing or the InputRoot is missing. If Command is missing, we short circuit. If InputRoot is missing, we will not tell you in any higher resolution (we can't because fs_util doesn't tell us). This seems acceptable for now based on existing limitations and overall responsibility of the client in Bazel.

illicitonion:
Sounds good :) It may be nice at some point in the future to give higher resolution, but we can treat that as a future feature request.

Issue #357, opened 2018-02-20
Hardcode support for empty/zero-length bazel directory


dgdonut:
Using an empty directory is handy for testing when the input environment does not matter, but this can break when using a Store backend that does not handle zero-length blobs. Either add hardcoded support for this case or give an error on attempts to upload zero-length blobs to the CAS.

PR #362, opened 2018-03-07
Bazel overhauls

dgdonut:
Will go through a pass of refinement, but wanted to get this viewable - assume no substantive changes unless requested.

Addresses #353 #357 #358 #361 

High level changes:
* Added some exec client code, more functionality to bzutil tool that can be used to play with remoting capability (CLI, self-explanatory with --help)
* Added plumbing to support population of more complex GoogleAPI RPC Status data down in the Runner, to be able to provide clients with required "your input dir or command was not uploaded to the CAS" type messages
** Required making bazel thrift data all optional. Do not @ me here, Thrift is stupid and this is the only way to make this work
* Some general tightening of response values to API spec. Ticket is opened to address testing here.
* Hardcoded empty sha support in Filer. I may redo this to avoid having the Filer aware of this, and have hardcoded "blobs" that get handled by the CAS API, since clients could still try to read or write empty sha data directly.
* No scoot headers for bazel tasks


illicitonion:
Verified that this fixes #353 #357 #358 from the pants side :) Thanks!

bazel/cas/service.go
```diff
@@ -89,19 +95,19 @@ func (s *casServer) BatchUpdateBlobs(
 	return nil, status.Error(codes.Unimplemented, "Currently unsupported in Scoot - update blobs independently")
 }
 
-// DEPRECATED - Included for protobuf generated code compatability/compilation
+// GetTree not supported in Scoot for V1
```
ryancouto, line 102: Is this confirmed as being part of the API spec?

binaries/bzutil/main.go
```diff
@@ -5,38 +5,101 @@ package main
...
+
 func main() {
 	log.AddHook(hooks.NewContextHook())
 
-	// TODO include optional directory that gets BzFiler.Ingest'ed, so we have a tool that easily satisfies prereqs
+	// TODO additional commands
+	// - upload Directory (tbd)
+	// - download outputs, stdout/err
+
+	// Subcommands
```
ryancouto: nit: this main() is a little long, consider splitting into addSubcommands, addSubcommandFlags, etc.

dgdonut:
this needs a review or approval @ryancouto @JeanetteBruno @jasonbs10

runner/runners/bazel.go
```diff
@@ -34,7 +38,8 @@ func preProcessBazel(filer snapshot.Filer, cmd *runner.Command) error {
 		return fmt.Errorf("Nil ExecuteRequest data in Command with RunType Bazel")
 	}
 	if err := fetchBazelCommand(bzFiler, cmd); err != nil {
-		return fmt.Errorf("Error retrieving Bazel command data: %v", err)
```
ryancouto, line 37: maybe log.Errorf(Error retrieving Bazel command data: %v")

runner/runners/bazel.go
```diff
@@ -34,7 +38,8 @@ func preProcessBazel(filer snapshot.Filer, cmd *runner.Command) error {
...
+		// NB: Important to return this error as-is
```
ryancouto: why?

runner/runners/bazel.go
```diff
@@ -79,20 +85,28 @@ func postProcessBazel(filer snapshot.Filer,
 
 	stdoutDigest, err := writeFileToCAS(bzFiler, stdout.AsFile())
 	if err != nil {
-		return nil, fmt.Errorf("Error ingesting stdout to CAS: %s", err)
+		errstr := fmt.Sprintf("Error ingesting stdout to CAS: %s", err)
+		log.Error(errstr)
+		return nil, fmt.Errorf("%s", errstr)
```
ryancouto: nit: for all of these, can just fmt.Errorf(errstr)

runner/runners/bazel.go
```diff
@@ -103,6 +117,7 @@ func postProcessBazel(filer snapshot.Filer,
 			StdoutDigest:      stdoutDigest,
 			StderrDigest:      stderrDigest,
 		},
+		ActionDigest: cmd.ExecuteRequest.ActionDigest,
```
ryancouto: Do we need to check if cmd.ExecuteRequest is nil to avoid a nil pointer dereference?

testCoverage.sh
```diff
@@ -2,11 +2,15 @@
 
 set -e
 echo "" > coverage.txt
+echo "" > dirlist.out
 
-for d in $(find . -name \*.go | xargs -n 1 dirname | uniq | grep -v /vendor/ | grep -v /fuse); do
+find . -name \*.go | xargs -n 1 dirname | uniq | grep -v /vendor/ | grep -v /fuse > dirlist.out
+
+for d in $(sort dirlist.out | uniq); do
```
ryancouto, line 9: curious - why this change?

runner/runners/invoke.go
```diff
@@ -93,8 +96,23 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
...
+			// the failure run status that indicates missing data to client
+			if _, ok := err.(*cas.NotFoundError); ok {
+				log.Info("NotFound error for Command during Bazel preprocess - Setting grpc Status error")
+				errStatus, err := getFailedPreconditionStatus([]*remoteexecution.Digest{
+					cmd.ExecuteRequest.Request.GetAction().GetCommandDigest()})
+				if err != nil {
+					log.Errorf("Error generating Failed Precondition status: %s", err)
+				} else {
+					failedStatus.ActionResult = &bazelapi.ActionResult{GRPCStatus: errStatus}
+					return failedStatus
```
ryancouto: don't need this, right?

runner/runners/invoke.go
```diff
@@ -158,8 +176,24 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
...
+			// Status error indicating that the InputRoot data could not be found.
+			if runType == runner.RunTypeBazel {
+				if _, ok := err.(*bzsnapshot.CheckoutNotExistError); ok {
+					log.Info("Checkout for Bazel command returned CheckoutNotExistError - Setting grpc Status error")
+					errStatus, err := getCheckoutMissingStatus(cmd.SnapshotID)
+					if err != nil {
+						log.Errorf("Error generating Failed Precondition status: %s", err)
+					} else {
+						failedStatus.ActionResult = &bazelapi.ActionResult{GRPCStatus: errStatus}
+						return failedStatus
```
ryancouto: ditto

bazel/cas/service.go
dgdonut: Yes - had this clarified in the Bazel API Revisions  gdoc

binaries/bzutil/main.go
dgdonut: added a few comments and grouped flags w/ subcommands to make a bit clearer

runner/runners/bazel.go
dgdonut: added

runner/runners/bazel.go
dgdonut: done, thanks

runner/runners/bazel.go
dgdonut: added comments here and in bazel/cas/service about returning NotFoundError. kind of an open question what the standard for errors should be - bubbling up vs wrapping in a new err, when to log, etc.

runner/runners/bazel.go
dgdonut: We should never hit that case, the first thing we do in the bazel workflow is check that ExecReq is non nil

testCoverage.sh
dgdonut: We have always been testing packages during the coverage step multiple times because we are not uniq'ing between for loop iterations

runner/runners/invoke.go
dgdonut: removed both, thanks

testCoverage.sh
ryancouto: gotcha - nice catch

runner/runners/bazel.go
ryancouto: Gotcha. Think it'd still be worth using the helper method cmd.ExecuteRequest.GetActionDigest() from https://github.com/twitter/scoot/pull/362/files#diff-9776fc4ea2a6e05f737bf7c9ef6106bfR58?

ryancouto (approved):

runner/runners/bazel.go
dgdonut: ohh, yes, thanks

PR #364, opened 2018-03-19
Offlining/Reinstatement of worker nodes

ryancouto:
In order to investigate worker nodes after they exhibit problematic behaviors, we need a manner by which to offline them and preserve their current state. This PR adds 2 thrift methods to the scheduler which take OfflineWorker and ReinstateWorker requests to accomplish that goal.


sched/scheduler/stateful_scheduler.go
```diff
@@ -1016,6 +1017,30 @@ func (s *statefulScheduler) GetSagaCoord() saga.SagaCoordinator {
...
+	s.clusterState.updateCh <- []cluster.NodeUpdate{cluster.NewUserInitiatedRemove(n)}
+	return nil
+}
+
+func (s *statefulScheduler) ReinstateWorker(req sched.ReinstateWorkerReq) error {
+	if !stringInSlice(req.Requestor, s.config.Admins) {
+		return fmt.Errorf("Requestor %s unauthorized to reinstate worker", req.Requestor)
+	}
+	n := cluster.NodeId(req.ID)
+	if ns, ok := s.clusterState.offlinedNodes[n]; !ok {
```
dgdonut: Worth checking here? No check in Offline request, and if for some reason the state changes between now and when the request is processed, the requestor won't know what happened anyway.

dgdonut (approved):
+ cluster_state_test ?
Admins - support for "*" or default permit all setting?

sched/scheduler/stateful_scheduler.go
ryancouto: True - will remove.

ryancouto: Ah - remembered we need this check, ns is used for the add. Can put a similar check in OfflineWorker for consistency.

PR #365, opened 2018-03-20
Check priority and killable tasks when preempting node for task assignment

dgdonut:
Crude check to address or redirect earlier seen panic.
* Pass priority of tasks to task assignment function via wrapper
* If trying to preempt on non-P3, skip and log. Alternative would be maintaining separate killableTasks lists, but this is simpler
* Check that killableTasks is populated (error not seen but was theoretically possible)

Basically a punt.

### DIFF
--- a/sched/scheduler/task_scheduler.go
+++ b/sched/scheduler/task_scheduler.go
@@ -16,6 +16,11 @@ type taskAssignment struct {
 	running *taskState
 }
 
+type taskToAssign struct {
+	task     *taskState
+	priority sched.Priority
+}
+
 type KillableTasks []*taskState
 
 // More recent tasks should come first in ascending sorts.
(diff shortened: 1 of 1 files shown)


jasonbs10 (commented):
Seems like it'd be simpler to add Priority to the TaskState structure? Or if it looks doable you could rip out all the kill logic...

dgdonut:
I thought about that, but didn't want there to be an assumption that the Priority field in TaskState would be actually valid and populated everywhere.
Yeah, hoping to rip some of this out in Q2.

jasonbs10 (commented):
On second though I think this change is too much of a punt. It avoids a panic in assign() but doesn't address the inconsistent state that was passed in or some of the other oddities we were seeing.

Given how rare the panic is, I'd rather have us wait until we can properly clean this up, or actually assert on our assumptions so we know immediately where it's broken.

I don't want to be the gatekeeper on such minor changes so I'll accept if you think this is more useful than I'm giving it credit for.

dgdonut:
I agree, it should prevent the panic but not the root cause. It could be interesting to see if, without this panic, something else breaks that could be more revealing. Or, it could let what is clearly undesirable behavior persist even longer. I'm fine leaving this out until we get to something more substantial.

ryancouto:
I'm also fine with leaving it out as long as we have a properly prioritized tracking ticket for asserting our assumptions throughout the scheduler loop. That seems like a task that could easily get perpetually punted.

dgdonut:
closing, will update internal ticket.

PR #366, opened 2018-03-22
bazel - use ActionCache when running bazel tasks

dgdonut:
* Add UpdateActionResult implementation
* Add client functions for ActionCache
* Add Cached field to domain ActionResult to be able to show client if result was cached
* When running bazel task, check cache for result
* When completing bazel task, upload result to cache
* Respect optional cache/skipcache settings

### DIFF
--- a/runner/runners/bazel.go
+++ b/runner/runners/bazel.go
@@ -28,24 +28,44 @@ import (
 	bzsnapshot "github.com/twitter/scoot/snapshot/bazel"
 )
 
-// Preliminary setup for handling Bazel commands - verify Filer and populate command
-func preProcessBazel(filer snapshot.Filer, cmd *runner.Command) error {
+// Preliminary setup for handling Bazel commands - verify Filer and populate command.
+// Can return a bazelapi.ActionResult if a result was found already in the ActionCache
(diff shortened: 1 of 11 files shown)


runner/runners/bazel.go
```diff
@@ -28,24 +28,45 @@ import (
...
+
+	// Check for existing result in ActionCache
+	if !cmd.ExecuteRequest.GetRequest().GetSkipCacheLookup() {
+		log.Info("Checking for existing results for command in ActionCache")
+		ar, err := cas.GetCacheResult(bzFiler.CASResolver, cmd.ExecuteRequest.GetActionDigest())
+		if err != nil {
+			// Only treat as an error if we didn't get NotFoundError
+			if _, ok := err.(*cas.NotFoundError); !ok {
+				log.Errorf("Failed to check for cached result, will execute: %s", err)
+			}
```
ryancouto: should there be a `return nil, err` here? (for the case that `err` isn't a `*cas.NotFoundError`

dgdonut: Nah - no reason to fail the run - we just want to log that we weren't able to check the cache (metrics to come)

ryancouto: whoops, missed the ! in !ok

ryancouto (approved):

PR #367, opened 2018-03-28
workerapi/server: Stats for drive type

ryancouto:
Problem

Running in a distributed environment with many workers, we don't know what type of disk drives our machines have.

Solution

Check a file (/sys/block/sda/queue/rotational) for 0, corresponding to an SSD, or assume the drive is an HDD. Then include that metric in our exported stats.

Result

Our dashboards will be able to pull in stats for disk drive type.


workerapi/server/server.go
```diff
@@ -35,19 +35,25 @@ func MakeServer(
 // (avoid conflict with other injected integers)
 type StatsCollectInterval time.Duration
 
+// a func that inspects the machine this worker is running on and returns
+// an identifying int64 for its storage drive type. It is assumed that
+// 0 corresponds to an SSD & 1 to an HDD
+type DriveTypeFunc func() int64
```
dgdonut, line 41: Only suggestion is returning an iota int64 type with an entry for unknown that can be returned e.g. when the func is not compatible with the os

binaries/workerserver/main.go
```diff
@@ -91,6 +92,19 @@ func main() {
 		func() execer.Memory {
 			return execer.Memory(*memCapFlag)
 		},
+		func() server.DriveTypeFunc {
```
dgdonut, line 95: If this passes for internal use as well, could define once as a default in setup.go (or if not in default but we want to reuse, throw under common somewhere)

jasonbs10 (changes requested):
I don't see the value in passing this through the worker unless you're sending the info back to scheduler. Why not just do this from our internal workerserver main.go and populate a stats gauge in one shot from there?

ryancouto:
@jasonbs10 Seemed like a general enough concern that it was worth putting in OSS. Your suggestion is easy enough though, can post a diff for that if you have a strong preference.

jasonbs10:
It is a general concern but it's a very specific thing to be piping through our constructors, and if we really wanted to do the right thing for OSS we'd add the appropriate *_darwin *_unix etc platform files. My feeling is that we should avoid anything platform specific in OSS since we don't really want to support that (I think we've largely stuck to that?).

ryancouto:
Fair enough, think we have too (besides fs_util). Closing in favor of internal diff.

PR #368, opened 2018-04-11
scheduler - remove p3 along with task preemption and killable node scheduling

dgdonut:
scorch the earth, round 1

### DIFF
--- a/sched/scheduler/job_state.go
+++ b/sched/scheduler/job_state.go
@@ -141,12 +141,15 @@ func (j *jobState) taskCompleted(taskId string, running bool) {
 }
 
 // Update JobState to reflect that an error has occurred running this Task
-func (j *jobState) errorRunningTask(taskId string, err error) {
+func (j *jobState) errorRunningTask(taskId string, err error, preempted bool) {
 	taskState := j.getTask(taskId)
 	taskState.Status = sched.NotStarted
 	taskState.TimeStarted = nilTime
 	taskState.TaskRunner = nil
 	j.TasksRunning--
+	if preempted {
+		taskState.NumTimesTried--
+	}
 }
 
 // Returns the Current Job Status
--- a/sched/scheduler/job_state.go
+++ b/sched/scheduler/job_state.go
@@ -141,15 +141,12 @@ func (j *jobState) taskCompleted(taskId string, running bool) {
 }
 
 // Update JobState to reflect that an error has occurred running this Task
-func (j *jobState) errorRunningTask(taskId string, err error, preempted bool) {
+func (j *jobState) errorRunningTask(taskId string, err error) {
 	taskState := j.getTask(taskId)
 	taskState.Status = sched.NotStarted
 	taskState.TimeStarted = nilTime
 	taskState.TaskRunner = nil
 	j.TasksRunning--
-	if preempted {
-		taskState.NumTimesTried--
-	}
 }
 
 // Returns the Current Job Status
--- a/sched/scheduler/stateful_scheduler.go
+++ b/sched/scheduler/stateful_scheduler.go
@@ -566,8 +566,8 @@ checkLoop:
 				err = fmt.Errorf("Exceeds max number of requestors: %s (%d)", checkJobMsg.jobDef.Requestor, s.config.MaxRequestors)
 			} else if len(jobs) >= s.config.MaxJobsPerRequestor {
 				err = fmt.Errorf("Exceeds max jobs per requestor: %s (%d)", checkJobMsg.jobDef.Requestor, s.config.MaxJobsPerRequestor)
-			} else if checkJobMsg.jobDef.Priority < sched.P0 || checkJobMsg.jobDef.Priority > sched.P3 {
-				err = fmt.Errorf("Invalid priority %d, must be between 0-3 inclusive", checkJobMsg.jobDef.Priority)
+			} else if checkJobMsg.jobDef.Priority < sched.P0 || checkJobMsg.jobDef.Priority > sched.P2 {
+				err = fmt.Errorf("Invalid priority %d, must be between 0-2 inclusive", checkJobMsg.jobDef.Priority)
 			} else {
 				// Check for duplicate task names
 				seenTasks := map[string]bool{}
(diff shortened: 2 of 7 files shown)


sched/scheduler/task_scheduler_test.go
```diff
@@ -184,8 +184,8 @@ func Test_TaskAssignments_RequestorBatching(t *testing.T) {
 }
 
 /*
-Add Job1.P0, Job2.P1 Job3.P2, Job4.P3, Job5.P0
-With 4 nodes, expect: scheduled Job4, Job3, Job2, Job1
+Add Job1.P0, Job2.P1 Job3.P2, Job4.P0
+With 3 nodes, expect: scheduled Job3, Job2, Job1
```
ryancouto, line 188: Job 4 when one of those finishes?

sched/scheduler/task_scheduler_test.go
```diff
@@ -296,38 +286,15 @@ func Test_TaskAssignments_PriorityStages(t *testing.T) {
...
-			t.Fatalf("Expected all P3 tasks, got %v", spew.Sdump(assignment))
-		}
-		for _, j := range js {
-			if j.Job.Id == assignment.task.JobId {
-				j.taskCompleted(assignment.task.TaskId, true)
-			}
-		}
-	}
-
 	// Check for 5 P2, 3 P1, and 2 P0 tasks
```
ryancouto: Looks like this isn't accurate based on line 294

sched/scheduler/task_scheduler_test.go
```diff
@@ -229,28 +226,24 @@ func Test_TaskAssignments_PrioritySimple(t *testing.T) {
...
 	}
 }
 
 /*
 Set NodeScaleFactor=.2 (10 NumConfiguredNodes / 50 SoftMaxSchedulableTasks) to get the following scheduling.
-Add jobs: (10 P3 Tasks), (10 P2 Tasks), (10 P1 Tasks), (10 P0 Tasks)
-With 10 nodes: assign all 10 to the p3 tasks.
-After finishing p3 tasks: assign nodes for 5 P2, 3 P1, and 2 P0 tasks
+Add jobs: (10 P2 Tasks), (10 P1 Tasks), (10 P0 Tasks)
+With 10 nodes: assign nodes for 5 P2, 3 P1, and 2 P0 tasks
```
ryancouto: ditto per line 294

sched/scheduler/job_state.go
```diff
@@ -141,15 +141,12 @@ func (j *jobState) taskCompleted(taskId string, running bool) {
 }
 
 // Update JobState to reflect that an error has occurred running this Task
-func (j *jobState) errorRunningTask(taskId string, err error, preempted bool) {
+func (j *jobState) errorRunningTask(taskId string, err error) {
```
jasonbs10: I believe this was a recent addition of mine that is meant to handle the case where worker goes offline/online and we abort the old run. So I think we still need this but you can verify if that sounds suspect.

sched/scheduler/stateful_scheduler.go
```diff
@@ -74,7 +74,7 @@ const MaxPriority = sched.P2
 
 // Decrease the NodeScaleFactor by taking the percentage defined by NodeScaleAdjust[Priority].
 // Note: the use case here is to hit an SLA for each job priority, and this is a coarse way to do so.
-var NodeScaleAdjustment = []float32{.05, .2, .75, 1}
+var NodeScaleAdjustment = []float32{.05, .2, .75}
```
jasonbs10: Let's adjust these to be a little more fair, it's too easy for p0 tasks to starve when everything seems to be high priority. Maybe .15, .3, .55? I'd like this rewritten or at least moved to a config file but I don't think you want to do that in this PR.

sched/scheduler/stateful_scheduler.go
```diff
@@ -870,23 +851,9 @@ func (s *statefulScheduler) scheduleTasks() {
 							"jobType":     jobType,
 							"tag":         tag,
 						}).Info("Task *node* lost, cleaning up.")
-				} else if nodeSt.runningJob != jobID || nodeSt.runningTask != taskID {
```
jasonbs10, line 873: I'm not entirely sure if this is just for P3. Do you think the if-statement above is sufficient to handle the node lost case?

sched/scheduler/task_scheduler.go
```diff
@@ -339,26 +287,17 @@ func assign(
...
-					"tag":              task.Def.Tag,
-					"node":             nodeSt.node,
-					"wasRunningJobID":  wasRunning.JobId,
-					"wasRunningTaskID": wasRunning.TaskId,
-					"wasRunningTag":    wasRunning.TaskRunner.Tag,
-				}).Info("Preempting node")
+					"jobID":  task.JobId,
+					"taskID": task.TaskId,
+					"tag":    task.Def.Tag,
+				}).Info("Unable to assign, no free node for task")
```
jasonbs10: Should this be a warn or err?

jasonbs10 (changes requested):
Thanks! Glad to see this all scorched, not sure why I ever thought it'd be appropriate for someone to do P3... super urgent things bypass scoot at twitter anyway.

How bad was it modifying the tests?

sched/scheduler/stateful_scheduler.go
dgdonut: I was trusting the comment here stating that this section should only get hit in the P3 case, which would have already handled cleanup.

Assuming for a sec that this could be hit in the node lost case as well - would we want to return like this block was doing and skip cleanup? What would have done the cleanup?

sched/scheduler/stateful_scheduler.go
dgdonut: Will update, but yeah will skip a refactor here

sched/scheduler/job_state.go
dgdonut: This bool was only set in the "if ta.running != nil" block, and that taskAssignment.running field was only ever set in the assign() function when we went to steal a node from killableTasks. With the removal of killable tasks there's not longer an entry point for this

sched/scheduler/task_scheduler.go
dgdonut: will warn

dgdonut:
Tests were actually not bad at all, pretty straightforward. Worked after first edit - I did some negative testing to make sure they were doing stuff :)

sched/scheduler/stateful_scheduler.go
jasonbs10: On second thought, let's trust the comment at face value.

That said, I think what's bothering me is the handling of jobState.numTimesTried if a node hiccups and comes back. We may need to keep the 'preempted' arg and set it to true if the nodeId is the same but the actual node object has been replaced.

sched/scheduler/job_state.go
jasonbs10: See other comment, I may have missed an edge case in my last fix, and we may still need this.

dgdonut:
Updated - is this what we are going for? Not sure about testing

jasonbs10 (approved):
We should play it safe with this release. My suggestion would be to run autosandbox in smf1 and then repurpose atla for a shadow version of autosandbox (I'm not sure if you need to disable phab/email notifications in order to accomplish this). Do you have something easier in mind?

dgdonut:
Following up w/ internal PR to pull in and manage release

PR #369, opened 2018-04-17
integration test - move to standalone binary to avoid deadlocking unit tests

dgdonut:
Just moving tests/integration_test to an independent binary, a la recoverytest.

I could not consistently repro this, but the test running in-line caused me to have to manually kill a go test and zombied scoot binary procs at least half a dozen times.

Have to say I find myself not particularly fond of the Twitter OSS PR template.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,6 +88,11 @@ recoverytest:
 	go install ./binaries/...
 	$(FIRSTGOPATH)/bin/recoverytest &>/dev/null
 
+integrationtest:
+	# Integration test with some overlap with other standalone tests, but utilizes client binaries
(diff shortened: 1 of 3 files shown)


jasonbs10 (approved):
Curious, did you see log.Infof("Killing %d cmds", len(c.watching)) or anything related from CreateLocalTestCluster-> scoot/scootapi/setup/cmds.go ?

dgdonut:
No usable info - test would totally lock up to the point where I would have to kill the 'go test' process, then hunt down all the scoot procs. It survived interrupts and well past the go test timeout flag.

PR #371, opened 2018-04-29
When killing unstarted tasks, we need to correctly log SagaStartTask before EndTask.

jasonbs10:


sched/scheduler/stateful_scheduler.go
```diff
@@ -1069,22 +1076,22 @@ func (s *statefulScheduler) killJobs() {
 				if err != nil {
 					s.stat.Counter(stats.SchedFailedTaskSerializeCounter).Inc(1) // TODO errata metric - remove if unused
 				}
-				//TODO - is this the correct counter?
 				s.stat.Counter(stats.SchedCompletedTaskCounter).Inc(1)
-				jobState.Saga.EndTask(task.TaskId, statusAsBytes)
+				if err := jobState.Saga.StartTask(task.TaskId, nil); err != nil {
```
ryancouto, line 1080: Why do we need to log a start before an end?

jasonbs10: https://github.com/twitter/scoot/blob/master/saga/saga_state.go#L267

ryancouto: Ah - would be nice if that dir had a README just to clarify what saga updates are allowed and when.

ryancouto (approved):

PR #379, opened 2018-06-30
Move to Bazel V2 API

dgdonut:
This reuses "forked" API dir as the place where we pull in the official V2 API from https://github.com/bazelbuild/remote-apis (via a submodule), and generate .pb.go code (which was previously provided already generated from a google repo).

The major significant changes from V1 are:
* Output data moved from Action to Command, Command now referenced from Action
* Execute Requests only reference Action
* Execute returns a stream, but we don't honor this

Overview of what the above changes required:
* Data struct shuffling
* Service endpoints adjusted
* bzutil tool updated to support v2 workflow
* ActionCache service reworked


illicitonion (approved):
I've tested that this works with the pants v2 implementation - I've not looked at the code

PR #381, opened 2018-07-16
bzutil: json for upload_action & upload_command digests

ryancouto:
Problem

Clients who are upload command & action are passed unstructured data through an Info level log in stderr, which is then needed for later reference when utilizing the execution API

Solution

Add --json flags to upload_action & upload_command

Result

When --json is passed, marshaled digest structs will print to stdout

### DIFF
--- a/bazel/execution/client.go
+++ b/bazel/execution/client.go
@@ -12,6 +12,7 @@ import (
 	google_rpc_code "google.golang.org/genproto/googleapis/rpc/code"
 	"google.golang.org/grpc"
 
+	"github.com/twitter/scoot/bazel"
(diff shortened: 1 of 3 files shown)


bazel/execution/client.go
```diff
@@ -151,7 +151,7 @@ func ExecuteOperationToStr(op *longrunning.Operation) string {
 	return s
 }
 
-func digestToStr(d *remoteexecution.Digest) string {
+func DigestToStr(d *remoteexecution.Digest) string {
```
dgdonut: Suggest move to bazel/digest.go if making public

dgdonut (approved):

PR #382, opened 2018-07-18
bazel/cas - fix support for chunked Write requests

dgdonut:
WriteOffset was assumed to be zero, which was incorrect per the Bytestream API. Expect a WriteOffset that matches the amount of data committed so far during a Write stream request.

Note 1 that we still do not support resumable requests or write status queries, but this should allow chunked writes of arbitrary size.

Note 2 that GRPC has an inherent cap of ~4MB on transmission size.


illicitonion (approved):
Thanks!

ryancouto (approved):

PR #391, opened 2018-08-29
Bazel remote execution support for a symlink to $JAVA_HOME

ryancouto:


common/common.go
```diff
@@ -14,3 +14,18 @@ func GetFirstGopath() (string, error) {
 	s := strings.Split(gp, ":")
 	return s[0], nil
 }
+
```
dgdonut, line 17: Add a leading comment so 'k=v,' format is clear

runner/runners/bazel.go
```diff
@@ -135,6 +136,22 @@ func fetchBazelCommandData(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command, rts
...
+	// Adjust environment as specified by supported platform properties
+	for _, platProp := range bzCommand.GetPlatform().GetProperties() {
+		if platProp.GetName() == "JDK_SYMLINK" {
+			javaHome, ok := os.LookupEnv("JAVA_HOME")
+			if !ok {
+				msg := "Unsuccessful lookup of $JAVA_HOME, not defined."
+				log.Error(msg)
+				return nil, fmt.Errorf(msg)
+			}
+			if b, err := exec.Command("ln", "-s", javaHome, platProp.GetValue()).CombinedOutput(); err != nil {
```
dgdonut: Let's use https://golang.org/pkg/os/#Symlink instead of exec here

runner/runners/bazel.go
```diff
@@ -135,6 +136,22 @@ func fetchBazelCommandData(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command, rts
 	}
 	cmd.Timeout = d
 
+	// Adjust environment as specified by supported platform properties
+	for _, platProp := range bzCommand.GetPlatform().GetProperties() {
+		if platProp.GetName() == "JDK_SYMLINK" {
+			javaHome, ok := os.LookupEnv("JAVA_HOME")
+			if !ok {
+				msg := "Unsuccessful lookup of $JAVA_HOME, not defined."
```
dgdonut: Suggest more generic prefix for errors here - "Failed setting up platform property X: $reason"

runner/runners/bazel.go
```diff
@@ -135,6 +136,22 @@ func fetchBazelCommandData(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command, rts
 	}
 	cmd.Timeout = d
 
+	// Adjust environment as specified by supported platform properties
```
dgdonut: Does this work here? This is pre checkout in invoker - does the symlink end up in the checkout working directory for the task run? Was assuming this logic would likely live in a function here and be called immediately after the checkout channel recv<- in invoker.

ryancouto: Nope you're right, totally missed that. Good catch.

runner/runners/bazel.go
```diff
@@ -135,6 +135,18 @@ func fetchBazelCommandData(bzFiler *bzsnapshot.BzFiler, cmd *runner.Command, rts
...
+	go func() {
+		for _, platProp := range bzCommand.GetPlatform().GetProperties() {
+			if platProp.GetName() == "JDK_SYMLINK" {
+				log.Info("JDK_SYMLINK platform property identified. Creating %s symlink", platProp.GetValue())
+				bzFiler.JDKSymlinkCh <- platProp.GetValue()
+				break
+			}
+			bzFiler.JDKSymlinkCh <- nil
```
dgdonut: I think this will not work if we get multiple properties and the first one is not "JDK_SYMLINK" - the if will get skipped and we'll send nil on the ch, even if the next prop was "JDK_SYMLINK."

This looks pretty dangerous being async - if there is ever a call to checkout without a channel send we will deadlock. I was thinking we would just write the symlink dir ontop of the checkout path after it's done (using the original logic, no change to the filer required).

DPT-11882 is more about how we could do better support for this type of thing vs throwing logic into runners/ - maybe an additional CheckoutExtended function in the Checkouter interface would add flexibility.

ryancouto: Misunderstood your previous feedback. Submitting new diff...

dgdonut (approved):
Looks good! Sorry about the confusion from my first request.

PR #393, opened 2018-09-05
apiserver/groupcache - populate cache with correctly-sized/capped byte slice

dgdonut:
Prior fixes reduced the heap growth rate, but we still had an underlying issue that we would retain 2x the configured groupcache size. This was entirely due to ioutil.ReadAll resulting in a []byte with twice the capacity of the length of the data (standard behavior).

Because we can't see the length of the data from the Reader, we have to make an additional copy after the readall and pass that to cache population.


jasonbs10 (approved):
Good catch.

PR #395, opened 2018-09-18
cas,bundlestore - limit maximum connection access to underlying store resources

dgdonut:
Most basic limiter implementation. Don't limit incoming connections because they are cheap, but block on access to underlying store from incoming areas: bundlestore server HTTP API and CAS grpc API. The APIs can use  a separate max as it's unlikely a thundering herd will hit the APIs at the same time as the use cases are somewhat different.

Max of 100 is arbitrary but reasonable based on typical observed use cases: a 10MB object (which is well on the larger side) would result in a maximum 1GB memory usage spike.

In the future we will like to make this command line configurable.

### DIFF
--- a/bazel/cas/service.go
+++ b/bazel/cas/service.go
@@ -33,6 +33,7 @@ type casServer struct {
 	server      *grpc.Server
 	storeConfig *store.StoreConfig
 	stat        stats.StatsReceiver
+	concurrent  chan struct{}
 }
 
 // Creates a new GRPCServer (CASServer/ByteStreamServer/ActionCacheServer)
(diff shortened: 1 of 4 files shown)


bazel/cas/service.go
```diff
@@ -425,6 +439,10 @@ func (s *casServer) GetActionResult(ctx context.Context,
 	if !s.IsInitialized() {
 		return nil, status.Error(codes.Internal, "Server not initialized")
 	}
+	if s.concurrent != nil {
```
jasonbs10: Not sure why you'd limit action results. I thought those are supposed to be small but I could be mistaken?

dgdonut: It's more about controlling any access to the underlying resource, until we do something smarter like limit by bytes and use digests to count concurrent access.

jasonbs10: Fair enough, if you have other plans in mind.

jasonbs10 (approved):

dgdonut:
Eh, backtracked on this and took restriction on actioncache out.

PR #396, opened 2018-09-18
bazel,snapshot - shortcuts for empty digest operations on client read, write, and checkout

dgdonut:
Bypass network calls in the CAS client libraries for Reading and Writing empty data. Skip calling fs_util when materializing/Checkout'ing an empty input root.

### DIFF
--- a/bazel/digest.go
+++ b/bazel/digest.go
@@ -45,3 +45,7 @@ func DigestToStr(d *remoteexecution.Digest) string {
 	}
 	return fmt.Sprintf("%s/%d", d.GetHash(), d.GetSizeBytes())
 }
+
+func IsEmptyDigest(d *remoteexecution.Digest) bool {
+	return d.GetHash() == EmptySha && d.GetSizeBytes() == EmptySize
+}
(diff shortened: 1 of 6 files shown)


bazel/digest.go
```diff
@@ -45,3 +45,7 @@ func DigestToStr(d *remoteexecution.Digest) string {
 	}
 	return fmt.Sprintf("%s/%d", d.GetHash(), d.GetSizeBytes())
 }
+
+func IsEmptyDigest(d *remoteexecution.Digest) bool {
+	return d.GetHash() == EmptySha && d.GetSizeBytes() == EmptySize
```
jasonbs10, line 50: Is it a possibility/problem if hash is EmptySha but size is not empty?

jasonbs10 (approved):

bazel/digest.go
dgdonut: It's unlikely but possible that could get passed somehow, but it's essentially an invalid digest and a sign that something was not computed or passed in correctly. Checking it is on the pedantic side but won't hide any errors that should surface because of it.

PR #397, opened 2018-09-28
bazel/cas - connection, rate, stream limiting

dgdonut:
Prevent memory usage spikes in CAS apiserver that occur during high concurrent request traffic. The spikes are caused by additional allocations done in grpc internals that can accumulate rapidly during extreme request load. Limiting CAS API implementations doesn't stop these accumulations as they are held until the underlying rpc/stream/connection/? is closed.

This introduces 3 types of limits:
* A netutil.LimitListener that limits overall simultaneous connections at the listener level
* A rate.Limiter implemented in the grpc server InTap handler that makes incoming requests wait if over threshold
* A MaxConcurrentStreams grpc server option that caps the number of concurrent streams per client connected to the apiserver

This replaces the previous buffered-channel-based limiter in the CAS API implementations that did not prevent this problem.

These are configured via const values but could be extended to be configurable by command line arguments.

### DIFF
--- a/bazel/cas/service.go
+++ b/bazel/cas/service.go
@@ -64,7 +64,7 @@ type tapLimiter struct {
 
 // TODO experimental
 func newTap() *tapLimiter {
-	return &tapLimiter{limiter: rate.NewLimiter(200, 20)}
+	return &tapLimiter{limiter: rate.NewLimiter(100, 20)}
 }
 
 func (t *tapLimiter) Handler(ctx context.Context, info *tap.Info) (context.Context, error) {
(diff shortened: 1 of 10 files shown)


bazel/cas/service.go
```diff
@@ -33,25 +36,47 @@ type casServer struct {
...
+
+func newTap() *tapLimiter {
+	return &tapLimiter{limiter: rate.NewLimiter(MaxRequestsPerSecond, MaxRequestsBurst)}
+}
+
+// Wait until the Limiter allows the a request or the Context expires
+// Client sees non-nil err as an RPC error with code=Unavailable, desc includes RST_STREAM or REFUSED_STREAM
+func (t *tapLimiter) handler(ctx context.Context, info *tap.Info) (context.Context, error) {
+	if err := t.limiter.Wait(ctx); err != nil {
+		log.Debugf("Tap limiter dropped connection due to rate limit: %s. Incoming request: %s", err, info.FullMethodName)
```
ryancouto: Wonder if we'll want this to be Warn or Error level...

ryancouto (approved):

PR #401, opened 2018-11-20
Bazel CAS batch operations

dgdonut:
Enables BatchUpdateBlobs, BatchReadBlobs. FindMissingBlobs now parallelized.

Adds support for above APIs with parallelized access to underlying Store (to a const threshold).

Updated generated remote execution protobuf code to latest from bazelbuild/remote-apis, and updated generation instructions.


bazel/cas/service.go
```diff
@@ -98,37 +101,275 @@ func (s *casServer) FindMissingBlobs(
...
+	for _, digest := range req.GetBlobDigests() {
+		if err != nil {
+			break
 		}
+		sem <- struct{}{}
+		wg.Add(1)
+		go func(d *remoteexecution.Digest) {
+			// We hardcode support for empty data in snapshot/filer/checkouter.go, so never report it as missing
+			// Empty SHA can be used to represent working with a plain, empty directory, but can cause problems in Stores
+			if d.GetHash() == bazel.EmptySha {
```
ryancouto, line 131: Should we check that d.SizeBytes is 0 too?

bazel/cas/service.go
```diff
@@ -98,37 +101,275 @@ func (s *casServer) FindMissingBlobs(
...
+		if err != nil {
+			break
+		}
+		sem <- struct{}{}
+		wg.Add(1)
+		go func(r *remoteexecution.BatchUpdateBlobsRequest_Request) {
+			writeRes := &remoteexecution.BatchUpdateBlobsResponse_Response{Digest: r.GetDigest()}
+
+			// Return OK if we got a request to write the empty sha
+			if r.GetDigest().GetHash() == bazel.EmptySha {
```
ryancouto, line 209: ditto for SizeByes

bazel/cas/service.go
dgdonut: We don't check for size in most existing places we check for EmptySha currently. I'm leaning towards skipping the size check everywhere we still do it (that would be like 2 places). It's kind of an undefined behavior on the part of the client, but on the server makes no practical difference as that data can never really exist internally.

bazel/cas/constants.go
```diff
@@ -13,6 +13,10 @@ const (
 	// Default buffer sizes
 	DefaultReadCapacity = 1024 * 1024
 
+	// Batch parallelism for underlying store operations
+	// NOTE if service implementation is changed, retest with setting <= 5
+	BatchParallelism = 5
```
ryancouto, line 19: Why 5?

bazel/cas/service.go
ryancouto: Makes sense

ryancouto (approved):

bazel/cas/constants.go
dgdonut: Good call, that's an arbitrary low value that will hopefully see some speedup but not cause issues. Adding a comment about it.

PR #402, opened 2018-11-26
snapshot/git - verification of unbundle type assertion

dgdonut:
Fixes a panic where the error was not of the previously assumed type. AFAIK this was only seen once, so just fixing to use verified type assertion check.


ryancouto (approved):
What type of error created the panic?

dgdonut:
context.deadlineExceededError (see comment on dpt-11977)

PR #404, opened 2018-12-19
dialer/resolver - add ResolveAll, update fs_util

dgdonut:
bump fs_util version and add ResolveAll to dialer.Resolver. Pass all apiserver addrs to bz_tree operations via ResolveAll.

(should be considered a WIP for now)

### DIFF
--- a/common/dialer/resolver.go
+++ b/common/dialer/resolver.go
@@ -9,6 +9,8 @@ import (
 type Resolver interface {
 	// Resolve resolves a service, getting an address or URL (or an error)
 	Resolve() (string, error)
+	// ResolveAll resolves a slice of all known addresses or URLs (or an error)
+	ResolveAll() ([]string, error)
 }
 
 // ConstantResolver always returns the same value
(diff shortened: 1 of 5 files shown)


common/dialer/resolver.go
```diff
@@ -41,6 +52,16 @@ func (r *EnvResolver) Resolve() (string, error) {
 	return os.Getenv(r.key), nil
 }
 
+// ResolveAll retuns the env key in a slice
+func (r *EnvResolver) ResolveAll() ([]string, error) {
+	all := []string{}
+	s, err := r.Resolve()
+	if s != "" {
+		all = append(all, s)
+	}
```
ryancouto, line 63: should we return an err if s == ""?

common/dialer/resolver.go
```diff
@@ -26,6 +28,15 @@ func (r *ConstantResolver) Resolve() (string, error) {
 	return r.s, nil
 }
 
+// ResolveAll returns the constant in a slice
+func (r *ConstantResolver) ResolveAll() ([]string, error) {
+	all := []string{}
+	if r.s != "" {
+		all = append(all, r.s)
+	}
```
ryancouto, line 38: ditto re: s == ""

common/dialer/resolver.go
dgdonut: Not 100% clear from the interface, but I think it's ok to return an empty result, where empty is "" for Resolve and []string{} for ResolveAll, if the resolution logic was successful and there were just no results. For these simple resolvers that essentially means no errors, but gives room to error in more complicated resolvers e.g. service discovery lookups where an http call can fail.

ryancouto: Fair - makes sense.

ryancouto (approved):

PR #405, opened 2018-12-20
bazel/cas - client APIs get simple retries, remove pedantic checks in unexported functions

dgdonut:
Low-effort retries for bazel cas client libraries used in the workers. High-effort retries commented on in DPT-12747, but this at least buys us additional Resolve() calls if e.g. the first try Apiserver is down.

### DIFF
--- a/bazel/cas/client.go
+++ b/bazel/cas/client.go
@@ -45,12 +45,24 @@ func IsNotFoundError(err error) bool {
 // Read data as bytes from a CAS. Takes a Resolver for addressing and a bazel Digest to read.
 // Returns bytes read or an error. If the requested resource was not found,
 // returns a NotFoundError
-func ByteStreamRead(r dialer.Resolver, digest *remoteexecution.Digest) ([]byte, error) {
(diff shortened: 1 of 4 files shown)


bazel/cas/client.go
```diff
@@ -45,12 +45,24 @@ func IsNotFoundError(err error) bool {
 // Read data as bytes from a CAS. Takes a Resolver for addressing and a bazel Digest to read.
 // Returns bytes read or an error. If the requested resource was not found,
 // returns a NotFoundError
-func ByteStreamRead(r dialer.Resolver, digest *remoteexecution.Digest) ([]byte, error) {
+// If retries > 0, does simple retry attempts when encountering errors
```
ryancouto, line 48: What's the desired behavior if retries is negative? Unlikely edge case, but with this being an exported function think we should cover our bases.

dgdonut: Agreed, fixing by setting to 0

ryancouto (approved):

PR #406, opened 2018-12-28
store/groupcache - use updated twitter/groupcache library that supports proper cache population

dgdonut:
Refactor groupcache Write as a groupcache "PutterFunc", see https://github.com/twitter-forks/groupcache/pull/1 for context.

groupcache submodule branch will be updated to a real commit when above PR lands.

Keep travis go version on latest minor release.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -65,7 +65,7 @@
 [submodule "github.com/twitter/groupcache"]
 	path = vendor/github.com/twitter/groupcache
 	url = https://github.com/twitter/groupcache
-	branch = dgassaway/groupcache-put
+	branch = 2ace291295909af4da31eda5f01c5bc4077ef613
 [submodule "github.com/twitter/fuse"]
 	path = vendor/github.com/twitter/fuse
 	url = https://github.com/twitter/fuse
--- a/.gitmodules
+++ b/.gitmodules
@@ -65,7 +65,7 @@
 [submodule "github.com/twitter/groupcache"]
 	path = vendor/github.com/twitter/groupcache
 	url = https://github.com/twitter/groupcache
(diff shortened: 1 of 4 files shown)


.gitmodules
```diff
@@ -65,7 +65,7 @@
 [submodule "github.com/twitter/groupcache"]
 	path = vendor/github.com/twitter/groupcache
 	url = https://github.com/twitter/groupcache
-	branch = 85fde1c5c36988481c30fc26605cd83f21cc94d1
+	branch = dgassaway/groupcache-put
```
jasonbs10: Reminder to change.

snapshot/store/groupcache_store.go
```diff
@@ -57,8 +63,26 @@ func MakeGroupcacheStore(underlying Store, cfg *GroupcacheConfig, stat stats.Sta
...
-		},
-	))
+		}),
+		groupcache.PutterFunc(func(ctx groupcache.Context, bundleName string, data []byte) error {
+			log.Info("New bundle, write and populate cache: ", bundleName)
+			stat.Counter(stats.GroupcacheWriteUnderlyingCounter).Inc(1)
+			defer stat.Latency(stats.GroupcacheWriteUnderlyingLatency_ms).Time().Stop()
+
+			// Get TTL via context type assertion, or set to nil and let underlying use its default
+			ttlc, ok := ctx.(*ttlContext)
```
jasonbs10, line 73: Tests to make sure this is propagated correctly?

dgdonut: We could test this in scoot, but it doesn't really belong here. The functionality is tested in the groupcache repo. A test in scoot could also pass as a false positive if e.g. the consistent hash changed.

dgdonut:
merging per offline approval from @jasonbs10

PR #407, opened 2019-01-04
snapshot/store - passthrough ttl duration on groupcache put

dgdonut:
Requires https://github.com/twitter-forks/groupcache/pull/2 - will update vendor & gitmodules when landed.

Context doesn't work for passing arbitrary values through groupcache, so the above PR adds a generic time.Duration ttl value. Pass this through so we preserve ttl values correctly.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -58,14 +58,10 @@
 	path = vendor/github.com/bazelbuild/remote-apis
 	url = https://github.com/bazelbuild/remote-apis
 	branch = cbf6ada7f5b2a0ce14646bf983d03b49118f0ec8
-[submodule "vendor/github.com/golang/groupcache"]
-	path = vendor/github.com/golang/groupcache
-	url = https://github.com/golang/groupcache
-	branch = 72d04f9fcdec7d3821820cc4a6f150eae553639a
 [submodule "github.com/twitter/groupcache"]
 	path = vendor/github.com/twitter/groupcache
 	url = https://github.com/twitter/groupcache
-	branch = dgassaway/propagate-ttl
+	branch = 3f0c8041e29b3d32eade74079a62f8fd7d9d9268
 [submodule "github.com/twitter/fuse"]
 	path = vendor/github.com/twitter/fuse
 	url = https://github.com/twitter/fuse
--- a/.gitmodules
+++ b/.gitmodules
@@ -65,7 +65,7 @@
 [submodule "github.com/twitter/groupcache"]
 	path = vendor/github.com/twitter/groupcache
 	url = https://github.com/twitter/groupcache
-	branch = 2ace291295909af4da31eda5f01c5bc4077ef613
+	branch = dgassaway/propagate-ttl
 [submodule "github.com/twitter/fuse"]
 	path = vendor/github.com/twitter/fuse
 	url = https://github.com/twitter/fuse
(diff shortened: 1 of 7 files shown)


snapshot/store/store.go
```diff
@@ -31,6 +31,19 @@ func GetTTLValue(c *TTLConfig) *TTLValue {
 	return nil
 }
 
+// Return a time.Duration based on Now given a TTLValue specifying a fixed time.
+// Returns a zero duration if TTLValue is nil or if duration would be negative.
```
jasonbs10, line 35: Why not return 'ok' as well?

snapshot/store/groupcache_store.go
```diff
@@ -165,8 +156,10 @@ func (s *groupcacheStore) Write(name string, data io.Reader, ttl *TTLValue) erro
 	c := make([]byte, len(b))
 	copy(c, b)
 
-	ttlc := &ttlContext{ttl: ttl}
-	if err := s.cache.Put(ttlc, name, c); err != nil {
+	// NOTE we potentially lose an overridden TTLValue.Key by passing through Groupcache.
+	// This isn't currently used and we should consider removing support for per-Write TTL Key values.
```
jasonbs10: Let's consider it now, we're unlikely to come back to this code once it has the desired perf.

snapshot/store/store.go
dgdonut: There's no action to take if "not ok." Returning zero is effective enough.

snapshot/store/groupcache_store.go
dgdonut: This comment was actually not correct - removed it. We do have use cases for per-write values and can retain the information, except for ttl key via groupcache but that's a minor exception.

.gitmodules
```diff
@@ -65,7 +65,7 @@
 [submodule "github.com/twitter/groupcache"]
 	path = vendor/github.com/twitter/groupcache
 	url = https://github.com/twitter/groupcache
-	branch = 2ace291295909af4da31eda5f01c5bc4077ef613
+	branch = dgassaway/propagate-ttl
```
ryancouto: change when merged

binaries/bazel-integration/main.go
```diff
@@ -69,8 +69,8 @@ func main() {
 	commandDigest := &remoteexecution.Digest{}
 	json.Unmarshal(b, commandDigest)
 	expectedCommandDigest := remoteexecution.Digest{
-		Hash:      "87d505be361979c2aa929aeeb38a61f648cc54b348f29e49ec05e0a412ad941c",
-		SizeBytes: 33,
+		Hash:      "1b00e10d51c107c1a1f06ebdc09dea3c45e06fd257481d085d4e37566f6a6041",
```
ryancouto, line 72: weird... this is already in master. not sure why it's showing as a change.

runner/runners/bazel.go
```diff
@@ -150,6 +150,25 @@ func setupJDKSymlink(path, filename string) error {
 	return os.Symlink(jh, filepath.Join(path, filename))
 }
 
+// API indicates directories where output files and directories would be created exist before execution
+func createOutputPaths(cmd *runner.Command, coDir string) error {
```
ryancouto, line 154: this is already in master too... is something funky going on with this PR?

dgdonut: Hmm, weird - when I do git diff master on my branch it doesn't show changes. I did a rebase after the previous PR and it asked me to merge my branch with my branch...

dgdonut: Yeah, in the commit list (here and on my branch) the commits got doubled because of the merge. I think squash and merge should resolve that, though.

dgdonut:
Merging today unless there are any objections.

ryancouto (approved):
Sorry, thought I approved yesterday. LGTM

PR #412, opened 2019-02-19
Log to stdout/err if task is killed for memory usage & set exitCode to 1

ryancouto:
Problem

Tasks that exceeded the memory limit imposed by a bounded os.osExecer provided no useful indication to clients on the cause of their task failure.

Solution

Log to stdout/err if a task exceeds the memory limit, and set exit code to 1 to indicate failure.

Result

Stdout/err will have a trailing message stating that the task failed due to memory constraints.

### DIFF
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -102,7 +102,7 @@ func (e *osExecer) Exec(command execer.Command) (result execer.Process, err erro
 
 	proc := &osProcess{cmd: cmd, wg: &wg, LogTags: command.LogTags}
 	if e.memCap > 0 {
-		go e.monitorMem(proc)
+		go e.monitorMem(proc, command.MemCh)
 	}
 	return proc, nil
 }
@@ -121,7 +121,7 @@ type osProcess struct {
 //
 // Periodically check to make sure memory constraints are respected,
 // and clean up after ourselves when the process has completed
-func (e *osExecer) monitorMem(p *osProcess) {
+func (e *osExecer) monitorMem(p *osProcess, memCh chan struct{}) {
 	pid := p.cmd.Process.Pid
 	pgid, err := syscall.Getpgid(pid)
 	if err != nil {
@@ -182,6 +182,7 @@ func (e *osExecer) monitorMem(p *osProcess) {
 					State: execer.FAILED,
 					Error: msg,
 				}
+				memCh <- struct{}{}
 				p.mutex.Unlock()
 				p.Abort()
 				return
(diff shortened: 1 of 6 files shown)


runner/execer/os/os_execer.go
```diff
@@ -182,6 +182,7 @@ func (e *osExecer) monitorMem(p *osProcess) {
 					State: execer.FAILED,
 					Error: msg,
 				}
+				memCh <- struct{}{}
```
dgdonut: Check if memCh is nil before sending

runner/runners/single_test.go
```diff
@@ -122,14 +122,16 @@ func TestMemCap(t *testing.T) {
...
 	// Travis may be slow, wait a super long time? This may also be necessary due to slow debug output from os_execer? TBD.
 	if runs, _, err := r.Query(query, runner.Wait{Timeout: 10 * time.Second}); err != nil {
 		t.Fatalf(err.Error())
-	} else if len(runs) != 1 || !strings.Contains(runs[0].Error, "MemoryCap") {
+	} else if len(runs) != 1 {
+		t.Fatalf("Expected a single COMPLETE run, got %v", len(runs))
+	} else if runs[0].ExitCode != 1 {
 		status, _, err := r.StatusAll()
-		t.Fatalf("Expected result with FAILURE and matching err string, got: %v -- status %v -- err %v", runs, status, err)
+		t.Fatalf("Expected result with COMPLETE and an exit code of 1, got: %v -- status %v -- err %v", runs, status, err)
```
dgdonut: Can we still check the MemoryCap err string?

runner/runners/invoke.go
```diff
@@ -369,17 +371,29 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 			}).Info("Run timedout")
 		return runner.TimeoutStatus(id,
 			tags.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID, Tag: cmd.Tag})
+	case <-memCh:
```
dgdonut: It would be very nice to write the memory cap here as well to save ourselves the inevitable trouble of telling people what it was. What if we set up the MemCh's to pass an execer.ProcessStatus and receive that here, and mix in the st.Error string into the formatted message?

runner/runners/single_test.go
ryancouto: It isn't present atm, due to the specs defined in runner/status.go

`// Only valid if State == (FAILED || BADREQUEST || ABORTED)`
`Error string`

I'm open to changing that though to allow for COMPLETE tasks to have Error's as well

runner/runners/invoke.go
ryancouto: Good call, I like that. Will do

dgdonut (approved):

PR #413, opened 2019-02-25
Fix memory watcher & write to stdlog in addition to stdout/err

ryancouto:
Problem

1. Memory watcher wasn't including some processes that should've been in the summation of memory usage, only processes in the same process group as the watched process. This excluded processes with parent processes in the process group of the watched process, but spawned into their own process group.
2. After exceeding memory cap, error message was written to stdout & stderr, but not stdlog (combination of stdout & stderr)

Solution

1. Revamp memory watcher script to include aforementioned processes.
2. Write error message to stdlog in addition to stdout/err

Result

More accurate memory watching & error message now written to stdlog as well.

### DIFF
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -232,6 +232,8 @@ func (e *osExecer) monitorMem(p *osProcess, memCh chan execer.ProcessStatus) {
 // Query for all sets of (pid, pgid, ppid, rss). Given a pid, find all processes with pid as its pgid or ppid.
 // Given this list of pids, find all processes with a pgid or ppid in that set, and modify the set in place.
 // From there, sum the memory of all processes in aforementioned set.
+// TODO: Python script is hard to test and probably non-performant compared to a native Go solution.
+// 		 Refactor into unit-testable Go func.
 func (e *osExecer) memUsage(pid int) (execer.Memory, error) {
 	str := `
 PID=%d
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -247,13 +247,13 @@ total=0
 for line in \"$PSLIST\".split(';'):
   pid, pgid, ppid, rss = tuple(line.split())
   all_processes[pid] = {'pgid': pgid, 'ppid': ppid, 'rss': rss}
-  process_groups.get(pgid, []).append(pid)
-  parent_processes.get(ppid, {}).append(pid)
+  process_groups.setdefault(pgid, []).append(pid)
+  parent_processes.setdefault(ppid, []).append(pid)
   if pid == \"$PID\":
     proc_group_id = pgid
 
 # Add all processes from pid's process group to related_processes
-for pid in process_groups[proc_group_id]:
+for pid in process_groups.setdefault(proc_group_id, []):
   related_processes[pid] = all_processes[pid]
 
 # Add children processes of related_processes to children list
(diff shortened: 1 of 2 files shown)


runner/execer/os/os_execer.go
```diff
@@ -229,26 +229,33 @@ func (e *osExecer) monitorMem(p *osProcess, memCh chan execer.ProcessStatus) {
...
-  total += int(memory[p])
+  pid, pgid, ppid, rss = tuple(line.split())
+  all_processes[pid] = {'pgid': pgid, 'ppid': ppid, 'rss': rss}
+
+for pid, proc in all_processes.items():
+  if pid == \"$PID\" or proc['pgid'] == \"$PID\" or proc['ppid'] == \"$PID\":
+    related_processes[pid] = proc
+
+for rel_pid, _ in related_processes.items():
+  for pid, proc in all_processes.items():
```
dgdonut: maybe trade space for time here by bucketing all the proc's into pgid buckets in the first all_processes pass, so we don't have to go back through that repeatedly. since we are currently doing this 4 times/sec and shared machines have 1k+ procs typically. (and all our compute is counting against ticks mesos is allotting our exec'd process)

ryancouto: Good call. Still have to do a nested iteration over the children processes so we get any children of children etc etc. But should be a lot quicker than going through the full proc list again.

runner/execer/os/os_execer.go
```diff
@@ -239,20 +239,28 @@ PSLIST=$(ps -e -o pid= -o pgid= -o ppid= -o rss= | tr '\n' ';' | sed 's,;$,,')
...
-    if proc['pgid'] == rel_pid or proc['ppid'] == rel_pid:
-      related_processes[pid] = proc
+  process_groups.setdefault(pgid, []).append(pid)
+  parent_processes.setdefault(ppid, []).append(pid)
+  if pid == \"$PID\":
+    proc_group_id = pgid
+
+# Add all processes from pid's process group to related_processes
+for pid in process_groups.setdefault(proc_group_id, []):
+  related_processes[pid] = all_processes[pid]
```
dgdonut, line 259: Still have to go through parent_processes here to catch the proc['ppid'] == pid case we were trying to add

dgdonut:
It would be more effort at this point, but the way this code is structured is not ideal. Suggest refactoring so that we do an Exec for the ps command only, and pass the output to a native go function that does our parsing, which can then be unit tested with some handbaked pid/ppid/pgid output lines. I understand if you want to defer that, but let's at least put a todo here.

runner/execer/os/os_execer.go
ryancouto: Do we? We should get that in line 265

ryancouto (commented):
Yeah, was thinking the same after debugging a couple typos turned out to be way more effort than it should be. I'll create a ticket and add a TODO

runner/execer/os/os_execer.go
dgdonut: That's checking pids that were already in related procs, which we only added by pgid at 256-257. Won't the procs that set their own pgid (but still will have a chain to the PID through ppid) get left out?

ryancouto: $PID is in related_processes too, so when pid == $PID, we should get any procs chained through ppid

dgdonut: Ah, right, nevermind

dgdonut (approved):

PR #414, opened 2019-03-01
Memory watcher script conversion to golang

ryancouto:
Problem

memUsage script was brittle and untestable, as it was a python script passed to the local python interpreter

Solution

Refactored / converted it to golang

Result

It's more easily testable / debuggable


runner/execer/os/os_execer.go
```diff
@@ -229,54 +229,78 @@ func (e *osExecer) monitorMem(p *osProcess, memCh chan execer.ProcessStatus) {
...
+	allProcesses := make(map[int]proc)
+	relatedProcesses := make(map[int]proc)
+	processGroups := make(map[int][]proc)
+	parentProcesses := make(map[int][]proc)
+	lines := strings.Split(string(b), ";")
+	var procGroupID int
+	total := 0
+	for idx := 0; idx < len(lines); idx += 1 {
+		processInfo := strings.Fields(lines[idx])
+		for idx := 0; idx < len(processInfo); idx += 1 {
```
dgdonut: Suggest these parse blocks with:
`var p proc
n, err := fmt.Sscanf(lines[idx], "%d %d %d %d", &proc.pid, &proc.pgid, ...
<verify n, err>`
seems to work without having to worry about variable space sizes.

runner/execer/os/os_execer.go
```diff
@@ -229,54 +229,78 @@ func (e *osExecer) monitorMem(p *osProcess, memCh chan execer.ProcessStatus) {
...
-	} else {
-		u, err := strconv.Atoi(strings.TrimSpace(string(usageKB)))
-		return execer.Memory(u * 1024), err
 	}
+	allProcesses := make(map[int]proc)
+	relatedProcesses := make(map[int]proc)
+	processGroups := make(map[int][]proc)
+	parentProcesses := make(map[int][]proc)
+	lines := strings.Split(string(b), ";")
+	var procGroupID int
```
dgdonut: For testing, if we break this out into a stub func we can test by passing in some hardcoded ps output lines that we know to hit all the ppid/pgid/etc conditions

dgdonut (commented):
Yesss... Nice

runner/execer/os/os_execer.go
ryancouto: great call, thanks!

runner/execer/os/os_execer.go
ryancouto: done

runner/execer/os/os_execer.go
```diff
@@ -249,64 +249,82 @@ type osProcGetter struct{}
...
-	allProcesses, relatedProcesses, processGroups, parentProcesses, err := e.pg.getProcs()
+	allProcesses, processGroups, parentProcesses, err := e.pg.getProcs()
 	if err != nil {
 		return 0, nil
 	}
 	if _, ok := allProcesses[pid]; !ok {
 		return 0, fmt.Errorf("%d was not present in list of all processes", pid)
 	}
 	procGroupID := allProcesses[pid].pgid
+	log.Info(procGroupID)
```
dgdonut: Reminder to remove, or move to debug?

runner/execer/os/os_execer.go
```diff
@@ -249,64 +249,82 @@ type osProcGetter struct{}
 // Given this list of pids, find all processes with a pgid or ppid in that set, and modify the set in place.
```
dgdonut: Comment out of date - would just describe what the function does, but leave how it does it to inline comments

runner/execer/os/os_execer_test.go
```diff
@@ -31,7 +31,7 @@ func TestOsExecerMemUsage(t *testing.T) {
 	}
 }
 
-// Tests that memory of processes spawned by a process in original process's process group are counted
+// // Tests that memory of processes spawned by a process in original process's process group are counted
```
dgdonut: double "//"

dgdonut (approved):

PR #416, opened 2019-03-05
Remove scoot daemon related code

ryancouto:
Problem

This code isn't used

Solution

Remove it

Result

Codebase is cleaned up (slightly)


dgdonut (approved):

dgdonut:
bai

PR #417, opened 2019-03-08
Jbruno/dpt12331 scheduler throttle

JeanetteBruno:
_Problem_
Allow us to throttle the scheduler, limiting the size of jobs it will accept.

_Solution_
Added throttling API to the scheduler:
**Throttle(limit)**: 
* limit == -1->no throttling, 
* limit > -1-> don't accept new job requests if the job's task count puts the scheduler's active and waiting tasks over limit

**GetSchedulerStatus()**:
returns: {receivingJobs: <true/false>, currentTasks: <number of tasks currently running/waiting>, maxTasks: <throttle limit>}

Note: The implementation does not try to resolve race conditions between accepting job requests and setting the throttle limit.

**JIRA: DPT-12331**


dgdonut:
Before review in detail, pasting comment from ticket:
Suggest we have SetSchedulerStatus to correspond with the Get API, that takes equivalent param (i.e. maxtasks ) for throttling. Also suggest dropping ReceivingJobs since it seems like this could just be interpreted from maxtasks, and the only way we can set this based on the draft is from that value anyway.

JeanetteBruno:
agree with taking out receiving, but not about changing throttle_scheduler to setSchedulerStatus... setSchedulerStatus is too generic, throttle_scheduler is a more intuitive name.

JeanetteBruno:
ignore prior statement about keeping Throttle, changed it to SetSchedulerStatus.

binaries/scootapi/main.go
```diff
@@ -15,6 +15,8 @@ import (
 // 		get_job_status [job id]
 //		watch_job [job id]
 //		run_smoke_test
+//		get_scheduler_status
+//		throttle [limit]
```
dgdonut: change ref to set_scheduler_status

dgdonut (commented):
less test failures, lgtm!

binaries/scootapi/main.go
JeanetteBruno: done

sched/scheduler/stateful_scheduler.go
```diff
@@ -109,6 +112,7 @@ type SchedulerConfig struct {
 	MaxRequestors           int
 	MaxJobsPerRequestor     int
 	SoftMaxSchedulableTasks int
+	HardMaxSchedulableTasks int
```
dgdonut: Suggest adding to typedef comment and noting it's a suggestion vs a hard cap now (maybe rename the internal var?)

dgdonut (approved):

PR #419, opened 2019-03-14
saga/sagalogs - add GC capability to in-memory saga log

dgdonut:
The InMemorySagaLog implementation (in retrospect, obviously) consumes memory in proportion to the work done by the scheduler. As more work is done, more memory is consumed for the log, potentially exhausting the host environment's resources.

Implement a GC based on time expiration. This is a departure from strict interpretation of the Saga pattern, but in practice GC can be set well after any practical Saga would be expected to have completed (subject to interpretation by the operator). This is the simplest practical GC; the goal is not a more controllable solution (e.g. cap based on number of sagas, memory consumed, etc).

The GC happens in the background, according to a configured interval and expiration time. Sagas older than the expiration are completely purged. Logging messages to an expired saga will result in an error.


saga/sagalogs/memory.go
```diff
@@ -0,0 +1,155 @@
+// Package sagalogs provides implementations of SagaLog.
```
dgdonut, line 1: This was actually copied from in_memory_saga_log.go with only moderate modification, but github not handling that well.

saga/sagalogs/memory.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// Returns an Instance of a Saga based on an inMemorySagaLog.
+// This is an in-memory impl, and is not durable.
+// Implements GC of whole sagas based on time expiration.
+// GC should be set at a duration that realistically will not purge active Sagas.
+// gcExpiration: duration after which a Saga was created, it will be deleted.
+// 	A zero duration is interpretted as "never gc" (the Log will eventually consume all memory).
+// gcInterval: duration interval at which GC runs.
+func MakeInMemorySagaCoordinator(gcExpiration time.Duration, gcInterval time.Duration) saga.SagaCoordinator {
```
JeanetteBruno, line 35: can we have units on gcInterval variable name: gcInterval_ns?  (I think Duration is in nanosecond units)

saga/sagalogs/memory_test.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	if !isSagaInActiveList(sagaId2, slog) {
+		t.Fatalf("Expected Saga %s to be in active list", sagaId2)
+	}
+
+	// reset s1's created time to be older than our GC expiration
+	slog.(*inMemorySagaLog).setSagaCreatedTime(sagaId1, time.Now().Add(-2*time.Hour))
+
+	// invoke the GC
+	err := slog.(*inMemorySagaLog).gcSagas()
```
JeanetteBruno, line 103: this would mask errors with the ticker-based go routine for cleaning up sagas. should we just wait 2 sec instead? (make the interval 1 sec)

saga/sagalogs/memory.go
dgdonut: time.Duration type makes that moot - this stops being a "sec" field after reading it in from the config

saga/sagalogs/memory_test.go
```diff
@@ -0,0 +1,104 @@
...
+		if !reflect.DeepEqual(loggedMsgs[j], outMsg) {
+			t.Errorf("Expected Logged Message and Returned Message to be Equal.  Expected %v, Actual %v",
+				loggedMsgs[j], outMsg)
+		}
+	}
+}
+
+func TestMemoryGC(t *testing.T) {
+	// set GC slow enough that it won't happen during the test
+	slog := MakeInMemorySagaLog(1*time.Hour, 1*time.Hour)
```
JeanetteBruno, line 83: try to have different values here to pick up errors with parameter ordering

saga/sagalogs/memory.go
```diff
@@ -0,0 +1,155 @@
...
+
+	for _, id := range expired {
+		delete(slog.sagas, id)
+	}
+
+	return nil
+}
+
+func (slog *inMemorySagaLog) getExpiredSagaIds() []string {
+	slog.mutex.RLock()
```
JeanetteBruno, line 144: should we only gc completed saga's (ones with endSaga)?

saga/sagalogs/memory_test.go
dgdonut: yeah - will add some better param tests

saga/sagalogs/memory.go
dgdonut: I was on the fence here. I initially thought about having a hard and soft cap - clean up completed sagas after a soft interval and any after a hard interval. I think in practice, though, we don't have a use case for that, and decided to keep it simpler. Do you have a strong preference?

saga/sagalogs/memory.go
```diff
@@ -0,0 +1,155 @@
...
+	defer slog.mutex.RUnlock()
+
+	expired := []string{}
+	for id, ld := range slog.sagas {
+		if time.Since(ld.created) >= slog.gcExpiration {
```
JeanetteBruno: Since() returns a Duration object.  Duration represents int64 nanosecond count..... os gcExpiration is nanosecond?

saga/sagalogs/memory.go
JeanetteBruno: just a thought.... in the spirit of extreme programming's don't build it till you are going to use it....  should we just dump saga log completely - are we using it for anything meaningful?

saga/sagalogs/memory_test.go
dgdonut: I think we should try to avoid sleep-dependent tests aggressively, particularly with the go race detector. Ideally anything time-dependent would have a "sim" mode that could be stepped through with exact code paths. In here we could build abstraction to be able to step the same way as the ticker... but we'd be doing that literally to just test that tickers work (as the code stands now)

saga/sagalogs/memory.go
dgdonut: I believe we still need it for keeping track of tasks at all currently. I'm not against ditching it but that's way beyond scope. We may have to look into that closer for another internal issue ("reduce scheduler queueing latency by a lot").

saga/sagalogs/memory.go
JeanetteBruno: as long as we clearly document that gc collects any saga (regardless of state), then as is is fiine.

saga/sagalogs/memory.go
dgassaway: They're both time.Durations and so are comparable; although it's an int64 under the hood I think it's misleading to have to interpret the values this way.

JeanetteBruno (approved):

saga/sagalogs/memory.go
dgdonut: added more clear comments

PR #420, opened 2019-04-01
[WIP] replace vendor/ with `dep`

cosmicexplorer:
### Problem

I had difficulty getting scoot to run using the contents of the `vendor/` directory, due to some compile errors relating to the versions of `thrift` and the google protobuf libraries, which didn't seem to be pinned. It was found that the `dep` tool could do this pinning.

*Explain the context and why you're making that change.  What is the
problem you're trying to solve? In some cases there is not a problem
and this can be thought of being the motivation for your change.*

### Solution

- Use `dep` instead to pin the versions of libraries.
- Add documentation about running `dep ensure -update` to install dependencies.

*Describe the modifications you've done.*

### Result

Scoot can be built from the get-go for any new cloner of this repo!

*What will change as a result of your pull request? Note that sometimes
this section is unnecessary because it is self-explanatory based on
the solution.*


Gopkg.toml
```diff
@@ -0,0 +1,102 @@
+# Gopkg.toml example
```
JeanetteBruno, line 1: - Our deps.sh script generates .gitmodules with the branch for each.  Will this replace that?

- I'm seeing _Go toolchain, as of 1.11, has (experimentally) adopted an approach that sharply diverges from dep_ ... what is the toolchain alternative?

- can we have a comment at the top of this file explaining it's purpose and how developers use it.  Example:
This is the configuration file for the dep dependency management tool.
This file ?, go.mod and go.sum? must be kept up to date with constraint entries for 3rd party libraries used by Scoot.

- we need an update in _https://confluence.twitter.biz/display/DPT/Development+Practices -> Adding a dependency to a new go project in OSS_ - add a new 'Proposed Update' subsection with the instructions for adding a new library.  When this change is accepted, we can then replace the original instructions with the updates.

go.mod
```diff
@@ -0,0 +1,29 @@
+module github.com/twitter/scoot
```
JeanetteBruno, line 1: if this is not a generated file add comment about how to maintain this code

go.sum
```diff
@@ -0,0 +1,73 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
JeanetteBruno, line 1: if this is not a generated file add comment explaining how user maintains this file

JeanetteBruno:
Thanks for doing this, sorry you had issues with the versions.

cosmicexplorer:
Despite [`dep` claiming the `go-dep` package is available for debian-based OSes](https://github.com/golang/dep), the package isn't found in CI, even when I bump the ubuntu version. Not sure what the discrepancy is here, will try to figure it out.

dgdonut:
This seems like a hairy thing to tackle. We should probably avoid go dep and just convert to go modules (the now and future standard) to get away from vendor. This has to be coordinated with internal usage, which right now essentially depends on vendor, submodules, and our hacky scripts.

dgdonut:
Replaced by https://github.com/twitter/scoot/pull/435 to move in the direction of modules

PR #421, opened 2019-04-15
add lots of debug logging to investigate flakiness

cosmicexplorer:
*I'm not sure if this change is something Scoot would want to do differently.*

### Problem

When running a local cluster with `go run ./binaries/setup-cloud-scoot/main.go --strategy local.local` to run pants against, we uncovered some flakiness, which was resolved in pants with pantsbuild/pants#7422. We investigated that by pairing timestamped logs from pants (which are being done in pantsbuild/pants#7536) with timestamped logs from scoot when uploading files (from this change).

*Explain the context and why you're making that change. What is
the problem you're trying to solve? In some cases there is not a
problem and this can be thought of being the motivation for your change.*

### Solution

- Add some info to the debug logs in the grpc cas write method, including:
  - timestamps
  - a randomly assigned request id

*Describe the modifications you've done.*

### Result

CAS writes via Scoot and any flakiness around them are easier to debug!

*What will change as a result of your pull request? Note that sometimes
this section is unnecessary because it is self-explanatory based on
the solution.*

### DIFF
--- a/bazel/cas/service.go
+++ b/bazel/cas/service.go
@@ -8,10 +8,11 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"net"
 	"sync"
 	"time"
-
+	// "runtime/debug"
 	"github.com/golang/protobuf/proto"
 	log "github.com/sirupsen/logrus"
 	remoteexecution "github.com/twitter/scoot/bazel/remoteexecution"
@@ -496,9 +497,14 @@ func (s *casServer) Read(req *bytestream.ReadRequest, ser bytestream.ByteStream_
 // store.Stores do not support partial Writes, and neither does our implementation.
 // We can support partial Write by keeping buffers for inflight requests in the casServer.
 // When the entire Write is buffered, we can Write to the Store and return a response with the result.
-// NOTE We also no not currently attempt any resolution between multiple client UUIDs writing the same resource
+// NOTE We also so not currently attempt any resolution between multiple client UUIDs writing the same resource
 func (s *casServer) Write(ser bytestream.ByteStream_WriteServer) error {
-	log.Debug("Received CAS Write request")
+
+	// debug.PrintStack()
+
+	var intId = rand.Intn(200000)
+
+	log.Debugf("%s(%d) -- Received CAS Write request", time.Now().UTC(), intId)
 
 	if !s.IsInitialized() {
 		return status.Error(codes.Internal, "Server not initialized")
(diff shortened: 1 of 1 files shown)


bazel/cas/service.go
```diff
@@ -8,10 +8,11 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"net"
 	"sync"
 	"time"
-
+	// "runtime/debug"
```
ryancouto: we can remove this right?

bazel/cas/service.go
```diff
@@ -496,9 +497,14 @@ func (s *casServer) Read(req *bytestream.ReadRequest, ser bytestream.ByteStream_
 // store.Stores do not support partial Writes, and neither does our implementation.
 // We can support partial Write by keeping buffers for inflight requests in the casServer.
 // When the entire Write is buffered, we can Write to the Store and return a response with the result.
-// NOTE We also no not currently attempt any resolution between multiple client UUIDs writing the same resource
+// NOTE We also so not currently attempt any resolution between multiple client UUIDs writing the same resource
```
ryancouto: think this is supposed to be *do

bazel/cas/service.go
```diff
@@ -496,9 +497,14 @@ func (s *casServer) Read(req *bytestream.ReadRequest, ser bytestream.ByteStream_
...
 func (s *casServer) Write(ser bytestream.ByteStream_WriteServer) error {
-	log.Debug("Received CAS Write request")
+
+	// debug.PrintStack()
```
ryancouto: Remove

bazel/cas/service.go
cosmicexplorer: Yes!

bazel/cas/service.go
cosmicexplorer: Done!

bazel/cas/service.go
cosmicexplorer: Yep!! Thanks for the double catch.

bazel/cas/service.go
```diff
@@ -8,10 +8,10 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"net"
 	"sync"
 	"time"
-
```
ryancouto, line 14: Could you keep this newline? we like to keep our imports separated by stdlib / 3rdparty / scoot stuff

bazel/cas/service.go
```diff
@@ -8,10 +8,10 @@ import (
...
 	"github.com/golang/protobuf/proto"
 	log "github.com/sirupsen/logrus"
 	remoteexecution "github.com/twitter/scoot/bazel/remoteexecution"
```
ryancouto, line 17: related to above comment, as long as you're in here, could you move this down to the scoot stuff below? thanks!

ryancouto (dismissed):

dgdonut:
Don't logs already have timestamps in them from logrus, or do these not appear when running locally? If that's the case can we fix that at the logging layer instead of explicitly logging timestamps only in a single function of a single module?

dgdonut:
Closing to cleanup PRs, open & update per comments if you'd like to revisit

PR #423, opened 2019-06-05
runner/runners - RunnerID and small debug fixes

dgdonut:
- Add runner.RunnerID struct for passing some IDing info during Runner creation.
- Use RunnerID in remexec ExecutedOperationMetadata.Worker.
- Print remexec ingested file/dir digests.
- Fix stderr printing in bzutil.

Example of runner id info in remexec task:

> 			ExecutionMetadata:
> 				Worker: smf1/dgassaway/devel/scoot-worker/10
> 				QueueLatency: 467ms
> 				WorkerTotal: 271ms
> 				InputFetch: 33ms
> 				Execution: 44ms
> 				OutputUpload: 194ms  sourceLine="bzutil/main.go:272"

### DIFF
--- a/binaries/bzutil/main.go
+++ b/binaries/bzutil/main.go
@@ -43,6 +43,7 @@ var supportedCommands map[string]bool = map[string]bool{
 
 func main() {
 	log.AddHook(hooks.NewContextHook())
+	log.SetLevel(log.InfoLevel)
 
 	// TODO additional commands
 	// - upload Directory (tbd)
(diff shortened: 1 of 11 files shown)


binaries/bzutil/main.go
```diff
@@ -43,6 +43,7 @@ var supportedCommands map[string]bool = map[string]bool{
 
 func main() {
 	log.AddHook(hooks.NewContextHook())
+	log.SetLevel(log.InfoLevel)
```
ryancouto: I think this is the default... Could take a log_level flag if we want to be able to enable debug level similar to the other ./binaries/...

runner/runners/setup.go
```diff
@@ -21,6 +22,9 @@ func (m module) Install(b *ice.MagicBag) {
 		func(m execer.Memory, s stats.StatsReceiver) execer.Execer {
 			return execers.MakeSimExecerInterceptor(execers.NewSimExecer(), osexec.NewBoundedExecer(m, s))
 		},
+		func() runner.RunnerID {
+			return runner.EmptyID
```
ryancouto: Assuming we're going to override this internally... Why not use something identifying, e.g. `exec.Command("hostname")` instead of EmptyID?

ryancouto (approved):
LGTM - couple little nits / clarifying Qs.

binaries/bzutil/main.go
dgdonut: Default seemed to be error which is why I added - logs had disappeared! (adding level flags)

runner/runners/setup.go
dgdonut: adding simple hostname + rand int

PR #424, opened 2019-06-08
snapshot/bazel - add stderr fs_util materialize debug data, bump fs_util version

dgdonut:


ryancouto (approved):

snapshot/bazel/bz_tree.go
```diff
@@ -112,7 +116,13 @@ func (bc bzCommand) runCmd(args []string) ([]byte, error) {
...
 
 	log.Debugf("%s %s", fsUtilCmd, args)
-	return exec.Command(fsUtilCmd, args...).Output()
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+	cmd := exec.Command(fsUtilCmd, args...)
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	err = cmd.Run()
+	return stdout.Bytes(), stderr.Bytes(), err
```
ryancouto, line 125: may be overkill, but if the command fails to run, stderr is populated in err.Stderr. may want to check for that on calling funcs

dgdonut: Already gets checked/propagated in save/materialize funcs

PR #425, opened 2019-06-10
snapshot/gitdb - no bundlestore updateStream, less flaky nodes

dgdonut:
Remove updateStream ability from bundlestore snapshots. In effect this means workers can't rely on a stream (in practice, a git fetch) as a backup when a snapshot can't be checked out.

Scheduler node flakyness: we currently set nodes as flaky if they return a failed status. In practice this has almost universally meant that the node failed to checkout a snapshot on task run. In remexec, this is normal behavior (we add more context detailing the failure), and for scoot filers this has generally indicated a problem with the snapshot, not the worker.

We don't intend to overhaul either of these behaviors, but this represents an improvement over the status quo.

### DIFF
--- a/snapshot/git/gitdb/bundlestore.go
+++ b/snapshot/git/gitdb/bundlestore.go
@@ -190,7 +190,6 @@ func (s *bundlestoreSnapshot) ID() snap.ID {
 func (s *bundlestoreSnapshot) Kind() SnapshotKind { return s.kind }
 func (s *bundlestoreSnapshot) SHA() string        { return s.sha }
 
-// TODO --- this is the entry point ---
 func (s *bundlestoreSnapshot) Download(db *DB) error {
 	log.Infof("Downloading sha: %s", s.SHA())
 	if err := db.shaPresent(s.SHA()); err == nil {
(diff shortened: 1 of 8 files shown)


snapshot/git/gitdb/bundlestore.go
```diff
@@ -225,27 +225,10 @@ func (s *bundlestoreSnapshot) Download(db *DB) error {
...
-	}
-
-	if _, err := db.dataRepo.Run("bundle", "unbundle", filename); err != nil {
-		// if we still can't unbundle, then the bundle might be corrupt or the
-		// prereqs might not be in the stream, or maybe the git server is serving us
-		// stale data.
-		log.Infof("Couldn't download sha: %s, the final unbundling attempt returned error: %s", s.SHA(), err.Error())
-		return err
-	}
+	// NOTE updating from a stream for bundlestore snapshots is no longer supported
```
ryancouto: what if instead of removing this functionality from OSS, we just used a different snapshot-satisfying interface internally with a Download() that returns error? In the general case (non-journaled repos), fetching to update bundles still seems useful.

dgdonut: I'm against duplicating slightly altered functionality internally.

I was considering making this behavior configurable, and somehow marking it as disabled when setting up the internal db - this would be pretty easy to do without disabling anything, WDYT?

Longer term, I think this is something we should consider removing entirely, and instead of hiding this behavior in "gitdb" snapshot semantics, greatly simplify the snapshot operations, and delegate all update/stream behavior to filer updaters - a checkout failure in that case would fall back to an explicit filer update attempt.

ryancouto: I'm against the duplication too - I was thinking more along the lines of something like making the git.DB's stream field an interface, and creating a noFetchStreamBackend internally that errors on updateStream.

Making it configurable could end up being less involved than that though, so I'm fine with that.

Agreed re: filer changes.

ryancouto (approved):

PR #428, opened 2019-06-25
Rcouto/dpt 15121/create tmpdir if dne

ryancouto:
Problem

The dir used by the localOutputCreator to write stdout/err to can be deleted, leaving workers unable to create output.
 
Solution

Recreate the tempdir and all nonexistent parents in the case that it doesn't exist

### DIFF
--- a/runner/runners/local_output_test.go
+++ b/runner/runners/local_output_test.go
@@ -21,6 +21,7 @@ func TestLocalOutputCreator(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Unable to create temp dir: %v", err)
 	}
+	defer os.RemoveAll(td.Dir)
 	h, err := NewHttpOutputCreator(td, "")
 	if err != nil {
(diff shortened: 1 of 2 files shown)


runner/runners/local_output_test.go
```diff
@@ -0,0 +1,60 @@
...
+	log.AddHook(hooks.NewContextHook())
+	logrusLevel, _ := log.ParseLevel("debug")
+	log.SetLevel(logrusLevel)
+}
+
+func TestLocalOutputCreator(t *testing.T) {
+	td, err := temp.TempDirDefault()
+	if err != nil {
+		t.Fatalf("Unable to create temp dir: %v", err)
+	}
```
dgdonut, line 23: + defer remove all td.Dir for both tests?

dgdonut (approved):

runner/runners/local_output_test.go
ryancouto: yup - thanks! meant to put that in and blanked

PR #429, opened 2019-07-10
sched/scheduler - advance to next scheduler loop if there is pending work to do

dgdonut:
Circled back to this after doing a little scheduler queue latency profiling on my flight.

Most of our scheduler queue latency stemmed from the scheduler TickRate. With this change we don't sleep the full tick if there's pending work on the scheduler's channels. Gets us about a 10x reduction in queue latency, at least from a local cluster perspective.

Also updated bzutil to print digests to stdout by default, and fixed some hidden go fmt/vet errors.

### DIFF
--- a/sched/scheduler/stateful_scheduler.go
+++ b/sched/scheduler/stateful_scheduler.go
@@ -440,8 +440,10 @@ func (s *statefulScheduler) loop() {
 	for {
 		s.step()
 
-		// wait until our TickRate has elapsed or we have a pending action.
-		// resend messages asynchronously in case the channel is blocked.
+		// Wait until our TickRate has elapsed or we have a pending action.
+		// Detect pending action by monitoring statefulScheduler's job channels.
+		// Since "detect" means we pulled off of a channel, put it back,
+		// asynchronously in case the channel is blocked/full (it will be drained next step())
 		select {
 		case msg := <-s.checkJobCh:
 			go func() {
--- a/sched/scheduler/stateful_scheduler.go
+++ b/sched/scheduler/stateful_scheduler.go
@@ -157,6 +157,7 @@ type statefulScheduler struct {
 	checkJobCh    chan jobCheckMsg
 	addJobCh      chan jobAddedMsg
 	killJobCh     chan jobKillRequest
+	stepTicker    *time.Ticker
 
 	// Scheduler State
 	clusterState   *clusterState
@@ -280,6 +281,7 @@ func NewStatefulScheduler(
 		checkJobCh:    make(chan jobCheckMsg, 1),
 		addJobCh:      make(chan jobAddedMsg, 1),
 		killJobCh:     make(chan jobKillRequest, 1), // TODO - what should this value be?
+		stepTicker:    time.NewTicker(TickRate),
 
 		clusterState:     newClusterState(initialCluster, clusterUpdates, nodeReadyFn, stat),
 		inProgressJobs:   make([]*jobState, 0),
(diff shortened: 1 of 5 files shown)


illicitonion (approved):
Thank you good sir!

sched/scheduler/stateful_scheduler.go
```diff
@@ -280,6 +281,7 @@ func NewStatefulScheduler(
 		checkJobCh:    make(chan jobCheckMsg, 1),
 		addJobCh:      make(chan jobAddedMsg, 1),
 		killJobCh:     make(chan jobKillRequest, 1), // TODO - what should this value be?
+		stepTicker:    time.NewTicker(TickRate),
```
ryancouto, line 284: should we consider decreasing TickRate?

sched/scheduler/stateful_scheduler.go
```diff
@@ -437,7 +439,24 @@ func generateJobId() string {
 func (s *statefulScheduler) loop() {
 	for {
 		s.step()
-		time.Sleep(TickRate)
+
+		// wait until our TickRate has elapsed or we have a pending action.
+		// resend messages asynchronously in case the channel is blocked.
```
ryancouto: sending back on the chan we recvd from looks a little strange here... could we add a little more to this comment explaining that it's so we skip the tick and jump back to s.step() as soon as pending action comes in?

sched/scheduler/stateful_scheduler.go
```diff
@@ -437,7 +439,24 @@ func generateJobId() string {
...
+		select {
+		case msg := <-s.checkJobCh:
+			go func() {
+				s.checkJobCh <- msg
+			}()
+		case msg := <-s.addJobCh:
+			go func() {
+				s.addJobCh <- msg
```
ryancouto, line 454: If we have multiple addJob requests, won't this reorder the addJobCh? Could a job be stuck at the bottom of the addJobCh queue if we keep receiving new requests (>1 every TickRate)?

snapshot/store/http_store.go
```diff
@@ -131,7 +131,7 @@ func (s *httpStore) Write(name string, data io.Reader, ttl *TTLValue) error {
 	} else {
 		resp.Body.Close()
 		if resp.StatusCode != http.StatusOK {
-			log.Infof("Write response status error: %s %v -- %s", uri, resp.Status)
+			log.Infof("Write response status error: %s -- %s", uri, resp.Status)
```
ryancouto, line 134: nice finds on these

sched/scheduler/stateful_scheduler.go
dgdonut: Yes it'll reorder, but only jobs received in the last tickrate window, which won't be perceptible. We won't starve anything though, once we have any jobs to add, the addJob function will drain the channels until they are empty within one step.

snapshot/store/http_store.go
dgdonut: just stumbled into these using a newer go version locally!

sched/scheduler/stateful_scheduler.go
dgdonut: updating

sched/scheduler/stateful_scheduler.go
dgdonut: maybe, but not clear what the "correct" value is. per the todo in l476 ideally we would refactor so the main loop is only triggered by incoming actions, not any timer

PR #430, opened 2019-07-10
Increase CAS ByteStream[Read/Write] & [Get/Update]CacheResult retries from 2 w constant backoff -> 5 w exp backoff

ryancouto:
Problem

CAS can be unreachable due to transient network issues

Solution

Increase retries from 2->5 & add exponential backoff

Result

We'll retry more times

### DIFF
--- a/bazel/cas/client.go
+++ b/bazel/cas/client.go
@@ -47,25 +47,18 @@ func IsNotFoundError(err error) bool {
 // Read data as bytes from a CAS. Takes a Resolver for addressing and a bazel Digest to read.
 // Returns bytes read or an error. If the requested resource was not found,
 // returns a NotFoundError
-// If retries > 0, does simple retry attempts when encountering errors
-func ByteStreamRead(r dialer.Resolver, digest *remoteexecution.Digest, retries int) (bytes []byte, err error) {
+func ByteStreamRead(r dialer.Resolver, digest *remoteexecution.Digest, b backoff.BackOff) (bytes []byte, err error) {
 	// skip request processing for empty sha
 	if digest == nil || bazel.IsEmptyDigest(digest) {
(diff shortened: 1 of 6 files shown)


dgdonut:
I saw this in the email for the diff but not here:
"Add exp backoff up to 3 sec" - seems like we don't do this but my guess is it'd greatly improve success odds

ryancouto:
> I saw this in the email for the diff but not here:
> "Add exp backoff up to 3 sec" - seems like we don't do this but my guess is it'd greatly improve success odds

Yeah, had it in there via the `http://google.golang.org/grpc/internal/backoff` pkg. Turns out `internal` isn't importable, didn't have time to find an alternative yesterday but will get something in to that effect today.

bazel/cas/client.go
```diff
@@ -45,23 +46,18 @@ func IsNotFoundError(err error) bool {
...
-	}
-
-	for ; retries >= 0; retries-- {
+	backoff.Retry(func() error {
 		bytes, err = byteStreamRead(r, digest)
-
-		if err == nil || IsNotFoundError(err) {
-			break
+		if IsNotFoundError(err) {
+			return nil
```
dgdonut, line 57: I guess these returns work because the return values are defined in the function signature, but there should be either 2 values returned or just a naked return, so these read weird. Same for all in this client file.

dgdonut (approved):
Approving but see inline comment. If you want to pull into internal, feel free to commandeer the phab I have open to pull in PR430 and just land it.

bazel/cas/client.go
ryancouto: Yeah... little funky because backoff.Retry only takes functions with signature `func() error` so we can't return 2 vals here

PR #432, opened 2019-07-22
bazel/server - add grpc max connection idle time

dgdonut:
Don't let clients potentially use up finite connections when idle. To be combined internally with 
"export GRPC_GO_LOG_VERBOSITY_LEVEL=99 && export GRPC_GO_LOG_SEVERITY_LEVEL=info" for debugging.

### DIFF
--- a/bazel/constants.go
+++ b/bazel/constants.go
@@ -14,4 +14,11 @@ const (
 	// Nil-data/Empty SHA-256 data
 	EmptySha  = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
 	EmptySize = int64(0)
+
+	// GRPC Server connection-related setting limits (defaults)
+	MaxSimultaneousConnections = 0 // limits total simultaneous connections via the Listener
+	MaxRequestsPerSecond       = 0 // limits total incoming requests allowed per second
+	MaxRequestsBurst           = 0 // allows this many requests in a burst faster than MaxRPS average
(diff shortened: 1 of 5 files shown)


bazel/constants.go
```diff
@@ -14,4 +14,11 @@ const (
 	// Nil-data/Empty SHA-256 data
 	EmptySha  = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
 	EmptySize = int64(0)
+
+	// GRPC Server connection-related setting limits (defaults)
+	MaxSimultaneousConnections = 0 // limits total simultaneous connections via the Listener
+	MaxRequestsPerSecond       = 0 // limits total incoming requests allowed per second
+	MaxRequestsBurst           = 0 // allows this many requests in a burst faster than MaxRPS average
+	MaxConcurrentStreams       = 0 // limits concurrent streams _per client_
+	MaxConnIdleMins            = 1 // limits open idle connections before the server closes them
```
ryancouto, line 23: these first 4 constants already exist in bazel/cas/constants.go, we should remove those or move MaxConnIdleMins into there

ryancouto: oh... unless we're attempting to have different defaults for the CAS vs for the scheduler...

dgdonut: yep - leaving the default unlimited and having scheduler use that (was hardcoded in binaries/main.go previously)

JeanetteBruno (approved):

PR #433, opened 2019-07-25
Abort tasks with SIGTERM to allow for cleanup

ryancouto:
Problem

When we abort a task, we send it a SIGKILL and don't give it a chance to cleanup after itself

Solution

Pass a SIGTERM for aborted tasks, with a 10 second timeout leading to a SIGKILL. For tasks that exceed memCap, we still SIGKILL.

Result

Tasks will be allowed to cleanup after themselves.

### DIFF
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -412,15 +412,15 @@ func (p *osProcess) Abort() (result execer.ProcessStatus) {
 	doneCh := make(chan error)
 	var err error
 	go func() {
-		doneCh <- p.cmd.Process.Signal(os.Interrupt)
+		doneCh <- p.cmd.Process.Signal(syscall.SIGTERM)
 	}()
 	select {
 	case err = <-doneCh:
 		if err != nil {
-			log.Errorf("Error aborting command via SIGINT: %s", err)
+			log.Errorf("Error aborting command via SIGTERM: %s", err)
 			err = p.Kill()
 		} else {
-			log.Info("Process aborted via SIGINT")
+			log.Info("Process aborted via SIGTERM")
 		}
 	case <-time.After(10 * time.Second):
 		msg := "10 second timeout for graceful abort exceeded."
@@ -429,7 +429,7 @@ func (p *osProcess) Abort() (result execer.ProcessStatus) {
 		err = p.Kill()
 	}
 	if err != nil {
-		result.Error += "Couldn't kill process. Will still attempt cleanup."
+		result.Error += " Couldn't kill process. Will still attempt cleanup."
 	}
 	_, err = p.cmd.Process.Wait()
 	if err, ok := err.(*exec.ExitError); ok {
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -22,14 +22,19 @@ import (
 
 const bytesToKB = 1024
 
-func NewExecer() *osExecer {
-	return &osExecer{pg: &osProcGetter{}}
+// Used for mocking memCap monitoring
+type procGetter interface {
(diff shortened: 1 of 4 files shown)


runner/execer/os/os_execer.go
```diff
@@ -395,29 +395,68 @@ func (p *osProcess) Wait() (result execer.ProcessStatus) {
...
+
+	// Attempt to SIGTERM process, allowing for graceful exit
+	// SIGKILL after 10 seconds
+	doneCh := make(chan error)
+	var err error
+	go func() {
+		doneCh <- p.cmd.Process.Signal(syscall.SIGTERM)
+	}()
+	select {
+	case err = <-doneCh:
```
dgdonut: Does this really mean the process terminated? It looks like it just means the signal was sent successfully - we still have to wait and potentially kill.

dgdonut (commented):
Sorry for the delay in review.

Can we add a test with a dummy script that signal swallows SIGTERM and verifies kill in that case?

runner/execer/os/os_execer.go
ryancouto: good point, we shouldn't trust the process

runner/execer/os/os_execer.go
```diff
@@ -395,29 +388,141 @@ func (p *osProcess) Wait() (result execer.ProcessStatus) {
...
-	result.Error = "Aborted."
+	p.result.State = execer.FAILED
+	p.result.ExitCode = -1
+	p.result.Error = "Aborted."
+
+	sigCh := make(chan error)
+	var err error
+
+	go func() {
+		sigCh <- p.cmd.Process.Signal(syscall.SIGTERM)
```
dgdonut: why async just for this function? we can just call and check err immediately

runner/execer/os/os_execer.go
```diff
@@ -395,29 +388,141 @@ func (p *osProcess) Wait() (result execer.ProcessStatus) {
...
+	cmdDoneCh := make(chan error)
+	go func() {
+		cmdDoneCh <- p.cmd.Wait()
+	}()
+
+	for {
+		select {
+		case err = <-cmdDoneCh:
+			if err != nil {
+				msg := fmt.Sprintf("Command failed to finish: %v", err)
```
dgdonut: "failed to finish [+successfully]" - is that more 'correct'?

runner/execer/os/os_execer.go
ryancouto: ah good catch - relic from previous iteration of this PR i forgot to change.

runner/execer/os/os_execer.go
ryancouto: sgtm

runner/execer/os/os_execer_test.go
```diff
@@ -81,6 +83,46 @@ func TestParentProcGroupAndChildren(t *testing.T) {
...
+	proc.Abort()
+	usage, err = e.memUsage(pid)
+	if usage != 0 {
+		t.Fatalf("Expected memUsage to be 0 after Abort & Kill, was %d", usage)
+	}
+
+	time.Sleep(3 * time.Second)
+	usage, err = e.memUsage(pid)
+	if err == nil {
+		t.Fatalf("Expected %d to not exist as a process anymore.", pid)
```
ryancouto, line 119: Note: pids do get reused so we could potentially see a false positive on this condition... But it'll be super rare given the short duration of the sleep. Think it's more important to check for this and possibly run into a false positive than to assume it works as expected.

dgdonut: yeah that's fine

dgdonut (approved):
LVGTM!

PR #434, opened 2019-07-26
Dgassaway/dpt 15508/grpc limits

dgdonut:
* Bump grpc connection limits on CAS server by ~5x, and remove stream limit (probably doens't matter)
* Changed Resolver interface's ResolveAll -> ResolveMany, which supports capping the number of returned addresses (passing 0 behaves like ResolveAll).
* Cap number of resolved addresses passed to fs_util calls at 10 to try and cap total connections

### DIFF
--- a/common/dialer/resolver.go
+++ b/common/dialer/resolver.go
@@ -10,9 +10,10 @@ type Resolver interface {
 	// Resolve resolves a service, getting an address or URL
 	// If a Resolve() call completes successfully but finds no addresses, it will return ("", nil)
 	Resolve() (string, error)
-	// ResolveAll resolves a slice of all known addresses or URLs
-	// If a ResolveAll() call completes successfully but finds no addresses, it will return ([]string{}, nil)
-	ResolveAll() ([]string, error)
+	// ResolveMany resolves a slice of random addresses or URLs
+	// The int parameter specifies the maximum number of addresses to return. If 0, all addresses are returned.
+	// If the call completes successfully but finds no addresses, it will return ([]string{}, nil)
+	ResolveMany(int) ([]string, error)
 }
 
 // ConstantResolver always returns the same value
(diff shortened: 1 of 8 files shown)


snapshot/bazel/constants.go
```diff
@@ -22,4 +22,6 @@ const (
 	fsUtilCmdGlobWildCard = "**"
 
 	emptyID = bazel.SnapshotIDPrefix + "-" + bazel.EmptySha + "-0"
+
+	maxResolveToFSUtil = 5
```
ryancouto, line 28: Would be helpful to add a comment here detailing why this value exists

common/dialer/resolver.go
```diff
@@ -10,9 +10,10 @@ type Resolver interface {
 	// Resolve resolves a service, getting an address or URL
 	// If a Resolve() call completes successfully but finds no addresses, it will return ("", nil)
 	Resolve() (string, error)
-	// ResolveAll resolves a slice of all known addresses or URLs
-	// If a ResolveAll() call completes successfully but finds no addresses, it will return ([]string{}, nil)
-	ResolveAll() ([]string, error)
+	// ResolveMany resolves a slice of random addresses or URLs
+	// The int parameter specifies the maximum number of addresses to return. If 0, all addresses are returned.
```
ryancouto: Given zero values, this interface makes sense, but it still seems awkward... -1 for "return all" might make more sense.

ryancouto (approved):

common/dialer/resolver.go
ryancouto: Not blocking on this and don't really have a strong opinion, just thinking out loud

dgdonut: fair enough - will make it <= 0 to effectively return all (since there's no valid reason to call with 0 anyway)

Issue #426, opened 2019-06-18
Fail to run


theidexisted:
One line summary of the issue here.

### Expected behavior

Run success.
### Actual behavior

Encounter error as:
```
# github.com/twitter/scoot/common/dialer                                                                                                   
common/dialer/dialer.go:40:12: assignment mismatch: 1 variable but d.transportFactory.GetTransport returns 2 values
# github.com/twitter/scoot/scootapi/gen-go/scoot
scootapi/gen-go/scoot/cloudscoot.go:96:20: not enough arguments in call to oprot.Flush
        have ()                   
        want (context.Context)                                                                                                             
scootapi/gen-go/scoot/cloudscoot.go:120:15: assignment mismatch: 2 variables but error8.Read returns 1 values
scootapi/gen-go/scoot/cloudscoot.go:180:20: not enough arguments in call to oprot.Flush
        have ()                   
        want (context.Context)                                       
scootapi/gen-go/scoot/cloudscoot.go:204:16: assignment mismatch: 2 variables but error10.Read returns 1 values
scootapi/gen-go/scoot/cloudscoot.go:264:20: not enough arguments in call to oprot.Flush
        have ()                                                      
        want (context.Context)                                       
scootapi/gen-go/scoot/cloudscoot.go:288:16: assignment mismatch: 2 variables but error12.Read returns 1 values
scootapi/gen-go/scoot/cloudscoot.go:348:20: not enough arguments in call to oprot.Flush
        have ()                                                      
        want (context.Context)                                       
scootapi/gen-go/scoot/cloudscoot.go:372:16: assignment mismatch: 2 variables but error14.Read returns 1 values
scootapi/gen-go/scoot/cloudscoot.go:431:20: not enough arguments in call to oprot.Flush
        have ()                   
        want (context.Context)
scootapi/gen-go/scoot/cloudscoot.go:455:16: assignment mismatch: 2 variables but error16.Read returns 1 values
scootapi/gen-go/scoot/cloudscoot.go:455:16: too many errors  
```

### Steps to reproduce the behavior

```bash
go run ./binaries/setup-cloud-scoot/main.go --strategy local.local
```

ryancouto:
Could you provide some more information about your local setup? What go version do you have installed? What's your $GOPATH? Is this from a clean master branch?

theidexisted:
Thanks for your quick response.

GOPATH is set to the source folder scoot.
go version go1.12.2 linux/amd64.
It's a clean master branch.

[...]

PR #435, opened 2019-07-26
Update to go 1.12.7 and module-based dependencies

dgdonut:
The intention of this is twofold:
* Bump to a more modern golang version and use the language's supported dep management workflow
* Make the user and developer workflows for scoot easier. This should make it easier to get scoot up and running "out of the box" without worrying about arcane requirements (see https://github.com/twitter/scoot/issues/426), and make dep management for developers easier (as well as cleaning up some existing rough edges)

Reviewable but not in a rush - might let this simmer for a couple of weeks.


ryancouto:
yesssss awesome. will take a closer look this weekend/next week.

Abhik1998:
Hey is this the fix for the issue #426 ??

dgdonut:
Yes.

JeanetteBruno (approved):
First - Thanks for deleting unused files!

do we need to check in generated files?

dgdonut:
We should check in generated stuff, yes - it ensures consistency between users in case e.g. someone has a different local binary of the generating tool, or the generating tool pulls in new libraries when it's invoked, resulting in different code generated, etc. It also separates the "build & run" use case from the "develop" use case for this repo, and makes the former one easier.

PR #437, opened 2019-07-31
Cancel for Checkouter/Ingester, implement for bztree

dgdonut:
We needed a way to cleanup stuck fs_util operations that happen as underlying calls during Checkout/Ingest for Bazel tasks, as these were left running previously. Add a Cancel interface and implement in BzFiler (in bzCommand). Run fs_util via an OsExecer so that we can abort it on Cancel().

### DIFF
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -37,6 +37,7 @@ type WriterDelegater interface {
 	WriterDelegate() io.Writer
 }
 
+// Implements runner/execer.Execer
 type osExecer struct {
 	// Best effort monitoring of command to kill it if resident memory usage exceeds this cap. Ignored if zero.
 	memCap execer.Memory
@@ -44,6 +45,7 @@ type osExecer struct {
 	pg     procGetter
 }
 
+// Implements runner/execer.Process
 type osProcess struct {
 	cmd    *exec.Cmd
 	wg     *sync.WaitGroup
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -20,8 +20,6 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const bytesToKB = 1024
-
 // Used for mocking memCap monitoring
 type procGetter interface {
 	getProcs() (map[int]proc, map[int][]proc, map[int][]proc, error)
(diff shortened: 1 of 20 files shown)


runner/execer/os/os_execer.go
```diff
@@ -229,7 +231,7 @@ func (e *osExecer) monitorMem(p *osProcess, memCh chan execer.ProcessStatus) {
 						"tag":    p.Tag,
 						"jobID":  p.JobID,
 						"taskID": p.TaskID,
-					}).Debugf("ps after increasing mem_cap utilization for pid %d", pid)
+					}).Tracef("ps after increasing mem_cap utilization for pid %d", pid)
```
ryancouto: This log is super verbose so I get dropping it to Trace, but it's come in handy a few times when profiling high mem usage tasks

ryancouto (approved):

snapshot/bazel/fs_util_test.go
```diff
@@ -216,3 +219,42 @@ func TestMaterializeEmptyDir(t *testing.T) {
...
+	}
+	doneCh := make(chan execer.ProcessStatus)
+
+	go func() {
+		doneCh <- bc.exec(cmd)
+	}()
+
+	// simplest way to ensure cancel gets run after exec actually starts
+	go func() {
+		for {
```
ryancouto, line 246: do we need any sort of cleanup to ensure this goroutine gets garbage collected?

runner/execer/os/os_execer.go
dgdonut: yes, although when we finally switch staging to prod, we'll lose debug level anyway. with using the os execer for fs_util runs, the log really becomes unusable, with multiple ps outputs per task in remexec

snapshot/bazel/fs_util_test.go
dgdonut: the test completes normally, so not really, although it does look a little clunky.

dgdonut:
This ended up getting more involved. I made some changes to improve behavior when OsExecer's Wait() and Abort() collided, and to reduce the runtime of the execer tests.

The problem with Wait/Abort was, attaching to the process via cmd.Wait() immediately fails if something else is already Wait()ing for the process (which in practice is very likely). This means we were sigterm'ing, trying to wait but immediately erroring, and then immediately killing the task without waiting for timeout.

PR #441, opened 2019-08-21
Rcouto/offline worker panic

ryancouto:
Problem

Offlined workers can still have tasks assigned to them, resulting in a panic when the scheduler attempts to access their ID in the active nodes map

Solution

Don't assign to offlined nodes

Result

The OfflineWorker API will be usable, and won't result in scheduler panics.

### DIFF
--- a/sched/scheduler/cluster_state.go
+++ b/sched/scheduler/cluster_state.go
@@ -38,10 +38,8 @@ type clusterState struct {
 }
 
 func (c *clusterState) isOfflined(ns *nodeState) bool {
-	for _, v := range c.offlinedNodes {
-		if v == ns {
-			return true
-		}
+	if _, ok := c.offlinedNodes[ns.node.Id()]; ok {
+		return true
 	}
 	return false
 }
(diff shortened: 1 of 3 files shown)


sched/scheduler/cluster_state.go
```diff
@@ -37,6 +37,15 @@ type clusterState struct {
 	stats            stats.StatsReceiver           // for collecting stats about node availability
 }
 
+func (c *clusterState) isOfflined(ns *nodeState) bool {
+	for _, v := range c.offlinedNodes {
```
dgdonut: isn't this a map?

ryancouto: c.offlinedNodes? yeah, the value (v) is a *nodeState though

dgdonut: right, I mean, can't we lookup via ns.node.NodeID instead of iterating?

ryancouto: ah, gotcha. yep 🤦‍♂

dgdonut (approved):
thanks!

PR #443, opened 2019-08-23
Use temporary repos for bundlestore view server downloads

dgdonut:
Requests to an apiserver's view endpoint are made with git-based objects. For example, to view a task std* log from the view server, the apiserver would fetch the underlying binary data from an underlying store, treat that as a git bundle and unbundle it into the apiserver's "blank" git repo, and then ingest & serve the requested file.

This "blank" git repo (an instance of snapshot/git/gitdb) is never pruned. Over time, as many bundles are unbundled into this repo, operations within it slow to a crawl. Instead of pruning this repo, we create a new code path (DownloadTempRepo) that creates these objects in fleeting blank git repos that are wiped entirely after the request is served.

There is no advantage to treating these types of requests as git bundles; if anything it's slowing things down and makes these requests difficult to understand. Refactoring this out will be time consuming, so some todos have been added to that effect and we opt to only fix the pruning issue here.

### DIFF
--- a/snapshot/git/gitdb/bundlestore.go
+++ b/snapshot/git/gitdb/bundlestore.go
@@ -135,6 +135,7 @@ func (b *bundlestoreBackend) uploadLocalSnapshot(s *localSnapshot, db *DB) (sn s
 		return nil, err
 	}
 
+	// TODO clean up
 	d, err := db.tmp.TempDir("bundle-")
 	if err != nil {
 		return nil, err
--- a/snapshot/git/gitdb/bundlestore.go
+++ b/snapshot/git/gitdb/bundlestore.go
@@ -11,7 +11,9 @@ import (
 
 	log "github.com/sirupsen/logrus"
 
+	"github.com/twitter/scoot/os/temp"
 	snap "github.com/twitter/scoot/snapshot"
+	"github.com/twitter/scoot/snapshot/git/repo"
 	"github.com/twitter/scoot/snapshot/store"
 )
 
@@ -135,7 +137,6 @@ func (b *bundlestoreBackend) uploadLocalSnapshot(s *localSnapshot, db *DB) (sn s
 		return nil, err
 	}
 
-	// TODO clean up
 	d, err := db.tmp.TempDir("bundle-")
 	if err != nil {
 		return nil, err
@@ -200,9 +201,12 @@ func (s *bundlestoreSnapshot) Download(db *DB) error {
 		return nil
 	}
 
-	// TODO(dbentley): keep stats about bundlestore downloading
-	// TODO(dbentley): keep stats about how long it takes to unbundle
-	filename, err := s.downloadBundle(db)
+	dlDir, filename, err := s.downloadBundle(db)
+	if dlDir != "" {
+		defer func() {
+			os.RemoveAll(dlDir)
+		}()
+	}
 	if err != nil {
 		log.Info("Unable to download bundle: ", err)
 		return err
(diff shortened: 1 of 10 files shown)


dgdonut:
Tests in github.com/twitter/scoot/sched/scheduler are flaky due to reliance on time.Sleep.

snapshot/git/gitdb/bundlestore.go
```diff
@@ -199,9 +201,10 @@ func (s *bundlestoreSnapshot) Download(db *DB) error {
 		return nil
 	}
 
-	// TODO(dbentley): keep stats about bundlestore downloading
```
JeanetteBruno, line 202: while we're in this area of the code, can you add a comment (for go docs) explaining Download()....(document externally visible features):
Update's db's repo with this bundlestoreSnapshot's SHA. The snapshot must be a git 'bundle' file.  If necessary, and db.bundles.cfg.AllowStreamUpdate is True, the db's repo will also be updated with missing ?commits? needed to update db's repo with bundlestoreSnapshot's SHA.  Returns True if the repo contains the bundlestoreSnapshot's SHA, error otherwise.

snapshot/git/gitdb/bundlestore.go
```diff
@@ -254,29 +257,57 @@ func (s *bundlestoreSnapshot) Download(db *DB) error {
 	return db.shaPresent(s.sha)
 }
 
-func (s *bundlestoreSnapshot) downloadBundle(db *DB) (filename string, err error) {
+// TODO separate the use cases that need git/repo/stream semantics from things that can be passed
+// around as binary bundles, and simplify the use cases accordingly
+func (s *bundlestoreSnapshot) DownloadTempRepo(db *DB, tmp *temp.TempDir) (*repo.Repository, error) {
```
JeanetteBruno, line 271: can we also have a comment here

tmp is not used - assuming we have it to conform to an interface and the use cases we want to separate.  Perhaps naming it 'notUsed' or 'notNeededForXXXUseCase' would be better?

snapshot/git/gitdb/checkout.go
```diff
@@ -11,21 +11,31 @@ import (
 	"github.com/twitter/scoot/snapshot/git/repo"
 )
 
+// NOTE we assume in practice this only gets used where downloading and reading
+// the file to a tempdir, which is then deleted, is acceptable
```
JeanetteBruno: don't understand the 'is acceptable' clause.

also comment explaining externally visible results.

snapshot/git/gitdb/stream.go
```diff
@@ -86,6 +88,10 @@ func (s *streamSnapshot) Download(db *DB) error {
 	return db.shaPresent(s.SHA())
 }
 
+func (s *streamSnapshot) DownloadTempRepo(db *DB, tmp *temp.TempDir) (*repo.Repository, error) {
```
JeanetteBruno, line 91: do we use streamSnapshot?  (see comment regarding updating the design docs)

snapshot/git/gitdb/local_data.go
```diff
@@ -39,3 +42,7 @@ func (s *localSnapshot) Download(db *DB) error {
 	// a localSnapshot is either present already or we have no way to download it
 	return db.shaPresent(s.SHA())
 }
+
```
JeanetteBruno, line 45: is localSnapshot used?

snapshot/git/gitdb/git_tags.go
```diff
@@ -99,6 +101,10 @@ func (s *tagsSnapshot) Download(db *DB) error {
 	return db.shaPresent(s.SHA())
 }
 
+func (s *tagsSnapshot) DownloadTempRepo(db *DB, tmp *temp.TempDir) (*repo.Repository, error) {
```
JeanetteBruno, line 104: is tagsSnapshot used?

snapshot/git/gitdb/db.go
```diff
@@ -12,6 +12,10 @@ import (
 	"github.com/twitter/scoot/snapshot/store"
 )
 
+// TODO The interfaces and functionality here should be refactored to only use git when necessary.
```
JeanetteBruno, line 15: we might want to revisit the overall snapshot/bundlestore implementation - does remexec/cloud exec need all the snapshots (fake_checkouter, fake_filer, etc.), gitdb, bundlestore pieces?  Can we implement a simpler bundlestore server (it seems the bazel server is a much simpler implementation?)

JeanetteBruno (approved):
I'm approving this given a minimal understanding of the architecture of this component.

It would be nice to have an architecture diagram, go docs, (anything) explaining the form/function of all these pieces (bundlestore, snapshots, git, store, db)

do we have tests anywhere?  Shouldn't they be updated to confirm the new behavior?

snapshot/git/gitdb/db.go
dgdonut: Agree but this is going to be complicated and time consuming to unravel. I would like us to eventually dedicate an amount of time specifically for cleaning up this repo.

dgdonut:
Will add/clean up existing comments

snapshot/git/gitdb/stream.go
dgdonut: there might be use cases for all of these snapshots

snapshot/git/gitdb/bundlestore.go
dgdonut: tmp does get used in this function

PR #445, opened 2019-08-30
Apratti/dpt 14866

anthonyjpratti:
Problem

Scoot can serve expired resources.

Solution

Propagate ttl from the underlying data store, and the Get code paths in groupcache.
Update groupcache(done in separate pr) to respect ttl.

Result

Scoot now respects ttl.


snapshot/store/groupcache_store.go
```diff
@@ -42,30 +42,33 @@ func MakeGroupcacheStore(underlying Store, cfg *GroupcacheConfig, ttlc *TTLConfi
...
+			}
+			return ttl, dest.SetBytes(data)
 		}),
-		groupcache.PutterFunc(func(ctx groupcache.Context, bundleName string, data []byte, ttl time.Duration) error {
+		groupcache.PutterFunc(func(ctx groupcache.Context, bundleName string, data []byte, ttl *time.Time) error {
 			log.Info("New bundle, write and populate cache: ", bundleName)
 			stat.Counter(stats.GroupcacheWriteUnderlyingCounter).Inc(1)
 			defer stat.Latency(stats.GroupcacheWriteUnderlyingLatency_ms).Time().Stop()
 
 			// Convert duration back to TTLValue
```
dgdonut: update comment

snapshot/store/groupcache_store.go
```diff
@@ -155,10 +159,8 @@ func (s *groupcacheStore) Write(name string, data io.Reader, ttl *TTLValue) erro
 	c := make([]byte, len(b))
 	copy(c, b)
 
-	d := GetDurationTTL(ttl)
-
 	log.Infof("Write() bundle %s: length: %d ttl: %s", name, len(b), ttl)
-	if err := s.cache.Put(nil, name, c, d); err != nil {
+	if err := s.cache.Put(nil, name, c, &ttl.TTL); err != nil {
```
dgdonut: May need a defense against nil ttl parameter here

snapshot/store/http_store.go
```diff
@@ -46,11 +46,21 @@ type httpStore struct {
 	client  Client
 }
 
-func (s *httpStore) OpenForRead(name string) (io.ReadCloser, error) {
+func getTTLValue(resp *http.Response) *TTLValue {
+	ttlk := http.CanonicalHeaderKey("x-ton-expires")
```
dgdonut: If we do this we should use the DefaultTTLKey in snapshot/store/store.go as the default, and if the httpStore config has a TTLConfig with TTLKey specified, to use that instead (so this will work with internal scoot)

snapshot/store/http_store.go
```diff
@@ -46,11 +46,21 @@ type httpStore struct {
...
+	expire := resp.Header.Get(ttlk)
+	ttl, err := time.Parse(time.RFC1123, expire)
```
dgdonut: If we do this we should also make the format string a const value in store.go and add as part of the TTLConfig as like TTLFormat

dgdonut (commented):
I'm a little rushed at the moment so will come back to review the addition of TTL to the OpenForRead API.

Also, unrelated but ping for https://github.com/twitter/groupcache/pull/9 - we can probably include that in here once it lands (I had an error building groupcache because of deps not being pinned there)

snapshot/store/store.go
```diff
@@ -50,7 +50,7 @@ type StoreRead interface {
 	Exists(name string) (bool, error)
 
 	// Open the bundle for streaming read. It is the caller's responsibility to call Close().
```
dgdonut, line 70: Since we're substantially changing the Store interfaces let's update the comments in this file a bit more.

Let's be explicit and go in this direction: if a non-nil TTL is used and that TTL has expired at read time, OpenForRead implementations should behave as if it was not found, and Exists should return false.

snapshot/store/groupcache_store.go
```diff
@@ -120,23 +123,24 @@ type groupcacheStore struct {
...
 
-func (s *groupcacheStore) OpenForRead(name string) (io.ReadCloser, error) {
+func (s *groupcacheStore) OpenForRead(name string) (io.ReadCloser, *TTLValue, error) {
 	log.Info("Read() checking for cached bundle: ", name)
 	defer s.stat.Latency(stats.GroupcacheReadLatency_ms).Time().Stop()
 	s.stat.Counter(stats.GroupcacheReadCounter).Inc(1)
 	var data []byte
-	if err := s.cache.Get(nil, name, groupcache.AllocatingByteSliceSink(&data)); err != nil {
-		return nil, err
+	if _, err := s.cache.Get(nil, name, groupcache.AllocatingByteSliceSink(&data)); err != nil {
```
dgdonut: Let's actually get the ttl from Get and populate a *TTLValue in the return if it wasn't null. We'll have to use the rest of the TTLValue fields from the TTLConfig

go.mod
```diff
@@ -21,7 +21,7 @@ require (
 	github.com/sirupsen/logrus v1.4.2
 	github.com/spf13/cobra v0.0.5
 	github.com/twitter/go-bindata v0.0.0-20180208183117-2fa2cba09795 // indirect
-	github.com/twitter/groupcache v0.0.0-20190208044251-7852b63ca833
+	github.com/twitter/groupcache v0.0.0-20190830195751-11158c6be16d
```
dgdonut: mod init landed, please update

snapshot/store/store.go
```diff
@@ -50,7 +50,7 @@ type StoreRead interface {
...
-	OpenForRead(name string) (io.ReadCloser, error)
+	OpenForRead(name string) (io.ReadCloser, *TTLValue, error)
```
anthonyjpratti: @dgassaway Thoughts on returning a struct with both io readcloser, and a ttlvalue?

snapshot/store/store.go
JeanetteBruno: (I apologize if I missed this in design discussions)... I am surprised to see the interface for read changing.... just would have expected read to return nil, 404 if the requested resource's ttl has expired.

anthonyjpratti: Yea it wasn't ideal, but it was the only way to propagate TTL from the underlying store to groupcache. I did have a better idea this morning that is backwards compatible.

```
OpenForRead(name string) (Resource, error)

type Resource struct {
	io.ReadCloser
	TTLValue *TTLValue
}
```

snapshot/store/store.go
dgdonut: I'm fine either way. In a way the struct makes more sense. If we go that route Write() should use it as well.

snapshot/store/store.go
dgdonut: Yeah it's required for how we want to use it in groupcache. It was also odd that the read interface left it out - you're able to write with a ttl but never read it?

snapshot/store/http_store.go
anthonyjpratti: I might be missing something, but It does not appear that the http store is configurable. Should I make it configurable?

dgdonut: Yep, you are right, will need to add TTLConfig

snapshot/store/groupcache_store.go
anthonyjpratti: Done.

snapshot/store/groupcache_store.go
anthonyjpratti: Done.

snapshot/store/http_store.go
anthonyjpratti: Done.

snapshot/store/groupcache_store.go
anthonyjpratti: Done.

snapshot/store/store.go
anthonyjpratti: I was thinking about keeping write as is, to preserve backwards compatibility. Thoughts on that?

dgdonut: We don't need to worry about backwards compat there, since this is just a code interface change, i.e. not changing a networked interface

snapshot/store/groupcache_store.go
dgdonut: It's actually ok to store a nil ttl, I meant just check here before dereferencing

snapshot/bundlestore/server_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	//"io/ioutil"
 	"net"
 	"net/http"
+
 	//"reflect"
```
ryancouto: nit: commented out packages & unnecessary line break

snapshot/store/groupcache_store.go
anthonyjpratti: ack

snapshot/store/store.go
anthonyjpratti: kk, done.

snapshot/store/store.go
```diff
@@ -44,13 +50,24 @@ func GetDurationTTL(t *TTLValue) time.Duration {
 	return d
 }
 
+// Resource is io readable as well as ttl
+type Resource struct {
+	io.ReadCloser
+	TTLValue *TTLValue
+}
+
+// NewResource constrcuts a new resource.
```
ryancouto: nit: constructs

snapshot/store/store.go
```diff
@@ -10,7 +10,12 @@ import (
...
 var DefaultTTL time.Duration = time.Hour * 24 * 180 //180 days. If zero, no ttl will be applied by default.
-const DefaultTTLKey string = "x-scoot-expires"      //the primary use for this is communicating ttl(RFC1123) over http.
+const (
+	// DefaultTTLKey is primarily used for communicating ttl(RFC1123) over http.
+	DefaultTTLKey string = "x-scoot-expires"
+	// DefaultTTLFormat is to specify the format for parsing ttl(RFC1123).
+	DefaultTTLFormat string = time.RFC1123
+)
 
 // Stores should generally support TTL, at this time only httpStore implements it.
```
ryancouto: as long as you're in here, would you mind editing this comment / adding one to TTLConfig?

anthonyjpratti: Done.

snapshot/store/store.go
anthonyjpratti: Oof.
Done.

snapshot/bundlestore/server_test.go
anthonyjpratti: Done.

dgdonut (approved):
I think everything has been addressed here!

PR #446, opened 2019-09-05
snapshot/store - add Length to store Resource and populate

dgdonut:
Explicitly set ContentLength in http_store.

The previous change in #445 altered the Store.Write API to use a Resource with an embedded io.ReadCloser. This had the side effect of obscuring the length of the data in the http store Write implementation, which only worked on more transparent Reader values. Add an explicit Length to Resources.

### DIFF
--- a/snapshot/git/gitdb/bundlestore.go
+++ b/snapshot/git/gitdb/bundlestore.go
@@ -330,11 +330,11 @@ func (b *bundlestoreBackend) uploadFile(filePath string, ttl *store.TTLValue) (s
 	if err != nil {
 		return "", err
 	}
+	defer f.Close()
 	fi, err := f.Stat()
 	if err != nil {
 		return "", err
 	}
-	defer f.Close()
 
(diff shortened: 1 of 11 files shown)


dgdonut:
Needs to pull in https://github.com/twitter/groupcache/pull/10

JeanetteBruno (approved):

snapshot/git/gitdb/bundlestore.go
```diff
@@ -330,14 +330,18 @@ func (b *bundlestoreBackend) uploadFile(filePath string, ttl *store.TTLValue) (s
 	if err != nil {
 		return "", err
 	}
+	fi, err := f.Stat()
+	if err != nil {
+		return "", err
+	}
 	defer f.Close()
```
ryancouto: we still want to Close() even if f.Stat() fails, right?

dgdonut: yes, thanks

PR #447, opened 2019-09-06
Define & propagate exit codes for various error states on workers & kill workers for cleaning failures

ryancouto:
Problem

We've observed workers getting into bad states where they continuously fail to `git clean -ffdx` their workspaces, requiring manual intervention to kill and have the worker reassigned to a fresh, uncorrupted host.

Solution

1. Define and propagate exit codes for various error states on workers (defined in common/errors/exit_codes.go)
2. Track last exit code in the runner
3. Have runner take itself down if it records multiple git clean errors in a row

Result

Workers will automatically be killed when they fail to `git clean -ffdx`


snapshot/git/gitdb/README.md
```diff
@@ -13,6 +13,16 @@ GitDB is an implementation of ScootDB (snapshot.DB) that stores local values in
 ## Backends
 GitDB uses different Backends to identify, upload and download Snapshots.
 
+## Exit Codes
+GitDB defines a set of exit codes that are returned depending on various git related errors (see git/gitdb/constants.go for definitive list):
```
ryancouto: note to self: update to point to common/errors/exit_codes.go

common/errors/exit_codes.go
```diff
@@ -0,0 +1,25 @@
+package errors
+
+type ExitCode int
+
+const (
```
dgdonut, line 5: Suggest we avoid collisions with nix-y things like sysexit.h and use 200+

runner/execer/execer.go
```diff
@@ -46,7 +46,7 @@ func (s ProcessState) String() string {
 }
 
 func (s ProcessState) IsDone() bool {
-	return s == COMPLETE || s == FAILED
+	return s != RUNNING
```
dgdonut, line 49: what about PENDING or other states?

common/errors/errors.go
```diff
@@ -0,0 +1,20 @@
+package errors
+
+type Error struct {
```
dgdonut: A little confusing with similarity to standard errors package, do we want to call it ScootError or something else?

dgdonut (commented):
Will revisit in more detail, but one overall comment:

Do we really need the determination to be made at the scheduler and introduce all the complexity that comes with that (namely the API extension)? Is there a simpler way the worker could keep track of its own failures and self-terminate? Does it seem likely we would use kills from the scheduler for other things in the future?

runner/execer/execer.go
ryancouto: That doesn't exist in the execer, all we have is UNKNOWN, RUNNING, COMPLETE, and FAILED

common/errors/exit_codes.go
ryancouto: Good point, will do

common/errors/errors.go
ryancouto: Yeah, I actually had it as ScootError at one point, but changed it bc I felt like that name 
1. Didn't tell me anything about why it's different from a standard error
&
2. Implied that this type of error should in some way be unique to Scoot, when in reality it's just an error with an exit code 

I do agree it's confusing though, maybe CodeError?

ryancouto:
> Will revisit in more detail, but one overall comment:
> 
> Do we really need the determination to be made at the scheduler and introduce all the complexity that comes with that (namely the API extension)? Is there a simpler way the worker could keep track of its own failures and self-terminate? Does it seem likely we would use kills from the scheduler for other things in the future?

Good point, and something I considered too. I'm torn on it, but I went with the API extension mainly because I could see us wanting to kill for other types of error variants, and having that happen from one location seems easier to maintain than kills sprinkled across the worker codebase. Additionally, we've historically tried to keep the worker as dumb as possible, stuffing logic in the scheduler, which motivated the decision a bit too.

I don't feel strongly, and I'm ok changing to handle all on the worker side if you think that's the right approach.

common/errors/exit_codes.go
ryancouto: Good point, will do

common/errors/errors.go
JeanetteBruno: ExitCodenError or ExitCodeError

runner/runners/queue.go
```diff
@@ -316,6 +316,11 @@ func (c *QueueController) Release() {
 	close(c.reqCh)
 }
 
+// Kills worker process
+func (c *QueueController) Kill() {
+	log.Fatal("Kill Requested")
```
dgdonut: What if a run is in progress? We would need to attempt to abort & give the opportunity to cleanup properly. As called from the scheduler that "shouldn't" be happening, but from an API standpoint there's no reason it couldn't be. Making this API  based opens us up to these kinds of problems in the future.

sched/scheduler/task_runner.go
```diff
@@ -138,6 +140,28 @@ func (r *taskRunner) run() error {
 			"tag":        taskErr.st.Tag,
 			"err":        taskErr,
 		}).Info("End task")
+
+	// Check if failure was due to failure to git clean. If so, kill worker
```
dgdonut: The other problem I have with the scheduler approach is that it's clunky to tune this to act on consecutive failures. We aren't certain we don't have flaky one-off command failures, so it's reasonable in the case we solving for now to require a few failures in a row before terminating.

dgdonut (commented):
The more I think about it the more I lean towards leaving the scheduler out of this. In general the scheduler being relied on to do too much is already an issue and I don't want to reinforce that. In this case the worker is capable in being self-sufficient in determining that it's in a fatal state and can terminate, similar to the behavior we expect when a worker fails to initialize source for some reason.

runner/runners/invoke.go
```diff
@@ -193,10 +199,8 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 				log.Errorf("Error canceling checkout: %s", err)
 			}
 			if err := <-checkoutCh; err != nil {
-				// If there was an error there should be no lingering gitdb locks, so return.
-				return
+				log.Errorf("Checkout errored: %s", err)
```
dgdonut, line 202: Don't have full context on the deleted comment and return - why is the return safe to get rid of?

runner/runners/queue.go
```diff
@@ -316,6 +320,11 @@ func (c *QueueController) Release() {
 	close(c.reqCh)
 }
 
+// Kills worker process
+func (c *QueueController) Kill() {
```
dgdonut: remove?

scootapi/scoot.thrift
```diff
@@ -29,14 +29,14 @@ enum JobType {
...
-  TIMEDOUT = 6     # Run timed out and was killed.
-  BADREQUEST = 7   # Request rejected due to unexpected failure.
+  UNKNOWN = 0      # Catch-all for indeterminate RunStatusStates. Also an "end state"
+  PENDING = 1      # Waiting to run
+  RUNNING = 2      # Running
+  COMPLETE = 3     # Succeeded or failed yielding an exit code
+  FAILED = 4       # Run mechanism failed and is no longer running
+  ABORTED = 5      # User requested that the run be killed, or task preempted by Scheduler
+  TIMEDOUT = 6     # Run timed out and was killed
+  #BADREQUEST = 7  # DEPRECATED
```
dgdonut: If we're deleting code level compatibility for this everywhere else, prefer we just delete it from here

snapshot/git/gitdb/README.md
```diff
@@ -13,6 +13,16 @@ GitDB is an implementation of ScootDB (snapshot.DB) that stores local values in
 ## Backends
 GitDB uses different Backends to identify, upload and download Snapshots.
 
+## Exit Codes
+GitDB defines a set of exit codes that are returned depending on various git related errors (see common/errors/exit_codes.go for definitive list):
+* Failed to git clean:        211
```
dgdonut: I wouldn't even enumerate these here to avoid defining 2x, just direct to the above file

workerapi/worker.thrift
```diff
@@ -11,14 +11,14 @@
...
-  TIMEDOUT = 6     # Run timed out and was killed.
-  BADREQUEST = 7   # Request rejected due to unexpected failure.
+  UNKNOWN = 0      # Catch-all for indeterminate RunStatusStates. Also an "end state"
+  PENDING = 1      # Waiting to run
+  RUNNING = 2      # Running
+  COMPLETE = 3     # Succeeded or failed yielding an exit code
+  FAILED = 4       # Run mechanism failed and is no longer running
+  ABORTED = 5      # User requested that the run be killed, or task preempted by Scheduler
+  TIMEDOUT = 6     # Run timed out and was killed
+  #BADREQUEST = 7  # DEPRECATED
```
dgdonut: same comment re: comment

runner/runners/invoke.go
ryancouto: The only difference is we now will still release the checkout (i.e. unlock the DB.workTreeLock) even if there was an error checking it out. I don't know why we assumed there wouldn't be any lingering gitdb locks if the checkout failed in the first place.

runner/runners/queue.go
ryancouto: Yep, thanks

scootapi/scoot.thrift
ryancouto: 👍

snapshot/git/gitdb/README.md
ryancouto: 👍

workerapi/worker.thrift
ryancouto: 👍

dgdonut (approved):
thanks!

PR #449, opened 2019-09-17
Use allocator (limiter) to limit concurrent CAS requests

dgdonut:
Preventing client requests from bringing down apiservers using grpc connection/stream/etc settings has been insufficient. Here we limit in the CAS API request handlers themselves based on the size of the requests as determined by the requested digests.

We do this with a new allocator package that manages leases on amounts of bytes. We cap the overall allocator capacity and force CAS requests to reserve amounts from this, or else we return grpc/Unavailable for a request.

This will have a side effect of capping the max size of data read/written via the CAS based on the allocator's max.

Bumps fs_util version.


snapshot/bazel/constants.go
```diff
@@ -19,11 +19,15 @@ const (
...
 	fsUtilCmdGlobWildCard = "**"
 
 	emptyID = bazel.SnapshotIDPrefix + "-" + bazel.EmptySha + "-0"
 
 	// limit amount of CAS addresses resolved in BzFiler calls to fs_util to this
 	// value, in order to limit number of connections
 	maxResolveToFSUtil = 5
+
+	// connection-limit fs_util flag setting
+	connLimit = 3
```
ryancouto, line 32: why 3?

dgdonut: This is the default (but setting explicit). We don't have a better reason to try setting higher or lower at this point

ryancouto (approved):

ryancouto:
hm... looks like this is in a funky state... pulled in my PR's changes as part of this PR

dgdonut:
recreating this PR

PR #450, opened 2019-09-17
Use allocator (limiter) to limit concurrent CAS requests

dgdonut:
Preventing client requests from bringing down apiservers using grpc connection/stream/etc settings has been insufficient. Here we limit in the CAS API request handlers themselves based on the size of the requests as determined by the requested digests.

We do this with a new allocator package that manages leases on amounts of bytes. We cap the overall allocator capacity and force CAS requests to reserve amounts from this, or else we return grpc/Unavailable for a request.

This will have a side effect of capping the max size of data read/written via the CAS based on the allocator's max.

Bumps fs_util version.


dgdonut:
This is identical to #449, which got clobbered by rebasing

ryancouto (approved):

PR #455, opened 2019-10-01
added apiserver load test framework

JeanetteBruno:
Problem
We need a load testing framework for the apiserver

Solution

The code starts a load test on apiserver.  The load test can be run as a single cli command or we can use the cli to start an http server waiting for http requests to start a load test.

Result

We'll be using this code to run performance tests on the api server


dgdonut:
Can we please include a description of how the test framework is designed, and specifically what it's capable of measuring on the apiserver?

PR #457, opened 2019-10-04
added apiserver load test framework

JeanetteBruno:
This pull request contains a tool for load testing the apiserver.

The main logic in the tool starts N concurrent upload/download actions using the apiserver's cas streamwriter/reader.  Uploading/downloading test data sets (generated by the framework) collecting latency stats.
The test can be run via command line (from mac) or can be deployed as an http server (on mesos) that starts the test via an http request.
The tests are parameterized to allow the request to define 
- the type of action, 
- number of concurrent actions, 
- range of data set sizes to use in the test (1kb, 10kb, 1m, 10m, 1g sizes)
- whether to run the test once or repeat running the test at a set frequency stopping after an elapsed time.


ryancouto (commented):
general comment - i'm not sure if ./binaries is the best place for the load_test library. could drop it into ./tests

ryancouto (commented):
also - do we want to run this as part of the travis PR check?

binaries/apiserver/load_test/http_api.go
```diff
@@ -0,0 +1,100 @@
...
+		} else {
+			lt.GetEndpoint(w, r)
+		}
+	default:
+		resp := fmt.Sprintf("Method %s not implemented", r.Method)
+		json.NewEncoder(w).Encode(resp)
+	}
+}
+
+func (lt *ApiserverLoadTester) GetEndpoint(w http.ResponseWriter, r *http.Request) {
```
dgdonut: Somewhat nit for a load tester - seems confusing having GET used to start tests - can we make this PUT or POST? (same with KillEndpoint (DELETE?))

binaries/apiserver/load_test/main/main.go
```diff
@@ -0,0 +1,164 @@
+/*
+This is the apiserver load tester.
+
+It is a cli that generates load on the apiserver.
+The load scenario is parameterized in terms of action (upload/download/both),
+file size (min, max), number of (concurrent) actions, frequency for running a scenario (once, every N minutes)
+
+when running locally, don't forget to open a tunnel to the apiserver:
+ssh nest.smfc.twitter.com -N -L 50002:<apiserver:grpc port>
```
dgdonut: please remove twitter internal-specific docs

binaries/apiserver/load_test/http_api.go
```diff
@@ -0,0 +1,100 @@
+package load_test
```
dgdonut: go packages shouldn't have underscores, i.e. rename to "loadtest"

binaries/apiserver/load_test/load_tester.go
```diff
@@ -0,0 +1,409 @@
+package load_test
+
+import (
+	"crypto/sha256"
+	"fmt"
+	"github.com/cenkalti/backoff"
```
dgdonut: please separate std imports, 3rdparty imports, and scoot imports

binaries/apiserver/load_test/load_tester.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	if lt.getKillRequested() {
+		lt.drainStopChAndResetStatus()
+		return nil
+	}
+
+	if lt.freq == 0 {
+		lt.runOneIteration() // run the test once
+		lt.drainStopChAndResetStatus()
```
dgdonut: nit - prefer to return in this block so the else block below can be removed, and that code is indented less

dgdonut:
Agree with this not being under apiserver for now, I'd suggest a perftest top level dir. I don't think this should run under travis; the goal of that is correctness vs speed and the environment is probably too unstable anyway.

binaries/apiserver/load_test/http_api.go
JeanetteBruno: agreed, I had it as POST then realized that I can't pass params in on url (using curl).   I moved to to get just to simplify curling.  If I make it POST we'll need to use -H "Content-Type: application/json"  and -d '{<param def json>}' in curl statements.

if we leave it as GET I can add the above as a comment.

binaries/apiserver/load_test/main/main.go
JeanetteBruno: oops

binaries/apiserver/load_test/http_api.go
dgdonut: Ok, get + comment sounds fine

JeanetteBruno: added comment

bazel/cas/conn_interfaces/conn_interfaces.go
```diff
@@ -0,0 +1,15 @@
+package conn_interfaces
```
dgdonut: rename to connection or something without underscores

bazel/cas/client.go
```diff
@@ -28,6 +30,46 @@ type NotFoundError struct {
 	Err string
 }
 
+
+// define default 3rd party dependencies
```
dgdonut: what does this comment mean?

bazel/cas/client.go
```diff
@@ -277,3 +319,155 @@ func updateCacheFromClient(acc remoteexecution.ActionCacheClient,
...
+	for _, content := range contents {
+		requests = append(requests, &remoteexecution.BatchUpdateBlobsRequest_Request{
+			Digest:               content.Digest,
+			Data:                 content.Data,
+		})
+	}
+	request := &remoteexecution.BatchUpdateBlobsRequest{
+		InstanceName:         bazel.DefaultInstanceName,
+		Requests:             requests,
+		XXX_NoUnkeyedLiteral: struct{}{},
```
dgdonut: we shouldn't explicitly define any field we are just setting to a zero value

bazel/cas/client.go
```diff
@@ -277,3 +319,155 @@ func updateCacheFromClient(acc remoteexecution.ActionCacheClient,
...
+		var err error
+		log.Debugf("Try #%d", try)
+		casCli.uploadedDigests, err = casCli.batchUpdateWriter(r, request, ctx)
+		try += 1
+		return err
+	}, b)
+
+	// if errors are found return an error result
+	if e != nil {
+		return casCli.uploadedDigests, fmt.Errorf("CAS BatchUpdateBlobs() returned error:%s", e.Error())
```
dgdonut: shouldn't we just return nil and the error here? can casCli.uploadedDigests be usable in this situation?

bazel/cas/client.go
```diff
@@ -277,3 +319,155 @@ func updateCacheFromClient(acc remoteexecution.ActionCacheClient,
...
+	}
+
+	return casCli.uploadedDigests, nil
```
dgdonut: Why is uploadedDigests part of the CASClient? Looks like we can remove and track only in this function.

binaries/bzutil/main.go
```diff
@@ -130,6 +156,26 @@ func main() {
 		}
 		parseAndSetLevel(*actionLogLevel)
 		uploadBzAction(*actionAddr, *actionCommandDigest, *actionRootDigest, *actionNoCache, *actionJson)
+	} else if batchUpload.Parsed() {
+		parseAndSetLevel(*batchUploadLogLevel)
+		if strings.ToLower(strings.Trim(*batchUploadTest, " ")) == "true" && (*batchUploadDir != "") {
+			fmt.Printf("Supply either -dir or -test=true not both.\n")
+			return
```
dgdonut: log.Fatal instead of returning

binaries/bzutil/main.go
```diff
@@ -130,6 +156,26 @@ func main() {
...
+		}
+		digestsStr, e := batchUploadFiles(*batchUploadAddr, *batchUploadDir,
+			strings.ToLower(strings.Trim(*batchUploadTest, " ")) == "true")
+		if e != nil {
+			log.Fatalf("batch update returned %s", e.Error())
+		} else {
+			fmt.Printf("the following digests were uploaded:\n%s\n", digestsStr)
```
dgdonut: This program should keep stdout machine readable - please just output newline-separated digests in the same format as other commands: "<hash>/<size>"

binaries/bzutil/main.go
```diff
@@ -130,6 +156,26 @@ func main() {
...
+		}
+	} else if batchDownload.Parsed() {
+		parseAndSetLevel(*batchDownloadLogLevel)
+		e := downloadTheBatch(*batchDownloadAddr, *batchDownloadDigests, *batchDownloadDir)
+		if e != nil {
+			log.Fatal(e)
+		}
+		fmt.Printf("contents downloaded to: %s", *batchDownloadDir)
```
dgdonut: this could be a log but not printf

binaries/bzutil/main.go
```diff
@@ -342,3 +388,101 @@ func printSupported() {
...
+	for i := 0; i < 10; i++ {
+		data := make([]byte, (i+1)*1000)
+		rand.Read(data)
+		uploadContent := makeUploadContent(data)
+		contents = append(contents, uploadContent)
+	}
+	return contents, nil
+}
+
+func makeUploadContent(data []byte) (cas.BatchUploadContent) {
```
dgdonut: This feels like we are conflating the perftest tool with bzutil. bzutil should ideally be as thin a wrapper around remexec client actions as possible. In that case bzutil would just implement the core functionality, and any wrapper tool would be where inputs were generated. The read subcommand is close to what we want although it does deviate by using json inputs. Everything else expects a digest to be in the form "<hash>/<size>".

bazel/cas/client.go
JeanetteBruno: casCli.uploadedDigests will contain the list of digests that were uploaded successfully, e will contain error messages for the digests that were not uploaded successfully.

bazel/cas/client.go
JeanetteBruno: done

binaries/bzutil/main.go
JeanetteBruno: done

binaries/bzutil/main.go
dgdonut: Realizing now github reformatted my comment, the digests look like: "1b00e10d51c107c1a1f06ebdc09dea3c45e06fd257481d085d4e37566f6a6041/76"

JeanetteBruno: done

binaries/bzutil/main.go
JeanetteBruno: done

binaries/bzutil/main.go
JeanetteBruno: this method doesn't make the data being uploaded... its making the BatchUploadContent data structure used to upload a batch of data blobs.  (I'll rename) BatchUploadContent is custom to bzutil (our bazel client utility to allow bazel users to provide a single list of (digests, data blob) pairs.  
If we don't want to introduce a new data structure we can simply use 2 lists (a list of digests and a list of data blobs).  I consider 2 lists to be more fragile: the list producers need to ensure that the 2 sets of entries are aligned (make sure they don't have a bug that adds one entry to the list but not to the other).

You are right that adding the "test" option to the cli conflates testing with a thin client wrapper concept (the perf testing actually is not using the test option.... I was just using it to manually test the cli).  I've removed the test option from the options and the code for generating sample data.

binaries/bzutil/main.go
dgdonut: Another late follow up to this one - can we prevent default dir param here? If test is false, and dir is not specified - then it might just start uploading stuff in the current directory, which seems unwanted

binaries/bzutil/main.go
dgdonut: Ok, sorry, I think I misread what this is doing. Seems reasonable as is

binaries/bzutil/main.go
```diff
@@ -64,6 +72,20 @@ func main() {
...
+	batchUpload := flag.NewFlagSet(batchUploadStr, flag.ExitOnError)
+	batchUploadAddr := batchUpload.String("cas_addr", scootapi.DefaultApiBundlestore_GRPC, "'host:port' of grpc CAS server")
+	batchUploadDir := batchUpload.String("dir", "", "dir containing files to upload: '/dir'")
+	batchUploadTest := batchUpload.String("test", "false", "when true, will upload 10 files randomly ranging from 1k to 1m")
+	batchUploadLogLevel := batchUpload.String("log_level", "info", "Log everything at this level and above (error|info|debug), default info")
+
+	// Batch Download
+	batchDownload := flag.NewFlagSet(batchDownloadStr, flag.ExitOnError)
+	batchDownloadAddr := batchDownload.String("cas_addr", scootapi.DefaultApiBundlestore_GRPC, "'host:port' of grpc CAS server")
+	batchDownloadDigests := batchDownload.String("digests", "", "digests identifying the contents to download: '[{Hash:<hash>, SizeBytes:<size>},...]")
```
dgdonut: Last consistency nit, all other digests get supplied in the hash/size format, if json is easier to integrate with the test tool prefer we have 2 flags that specify the digests in either plaintext or json forms

bazel/cas/conn_interfaces/conn_interfaces.go
JeanetteBruno: renamed to connection-api.  didn't like the no underscores option (connectionapi)

bazel/cas/client.go
JeanetteBruno: rewrote comment (hopefully its clearer now)

binaries/bzutil/main.go
JeanetteBruno: done

bazel/cas/conn_interfaces/conn_interfaces.go
ryancouto: The directory is now connection-api, but the package name is still connection_api. The golang style guide endorses using single word names (https://golang.org/doc/effective_go.html#names), could we just use "connection"?

binaries/bzutil/main.go
```diff
@@ -342,3 +383,108 @@ func printSupported() {
...
+
+func downloadTheBatch(casAddr string, digestsStr string, toDir string) error {
+	casClient := cas.MakeCASClient()
+	resolver := dialer.NewConstantResolver(casAddr)
+
+	// parse the digests into []*Digests
+	entries := strings.Split(digestsStr, ",")
+	digestsPtr := make([]*remoteexecution.Digest, len(entries))
+	for i, entry := range entries {
+		digestParts := strings.Split(entry, "/")
```
dgdonut: There's a lib function for this: bazel.DigestFromString in bazel/digest.go

bazel/cas/conn_interfaces/conn_interfaces.go
dgdonut: +1

dgdonut (approved):
Thanks for seeing this through!

JeanetteBruno:
abandoned, replaced with [468](https://github.com/twitter/scoot/pull/468)

PR #458, opened 2019-10-04
Scheduler reorganization

ryancouto:
Problem

scoot's codebase is disorganized and tough to grok

Solution

move all of the scheduler code into one directory

Result

easier to grok scheduler stuff


Makefile
```diff
@@ -99,19 +99,19 @@ test-all: test-unit-property-integration coverage
 
 ############## standalone binary & integration tests
 
-swarmtest: install
+smoketest:
```
dgdonut, line 103: why did these lose dependency on install?

scheduler/client/addrs.go
```diff
@@ -1,4 +1,4 @@
-package scootapi
+package client
```
dgdonut: Putting addrs under client isn't clear to me. Now we have various packages across the repo importing scheduler/client - that's misleading

dgdonut (commented):
Overall in the right direction but I have some reservations.

The biggest thing that seems "off" to me is the domain package structure. Domain seems more like an api-related package as it just deals with converting between thrift and internal data structures. And moving all of the core scheduler functionality under it makes that less discoverable.

Without going so far as refactoring how all of the API pieces fit together, do you think you can make that clearer without polluting the main scheduler dir too much? Off the top of my head a "server" dir under scheduler would encompass all of the main component functionality. Ideally, in the future, not much outside of scheduler would need to import this package which would reduce the ambiguity of the name.

Makefile
ryancouto: seemed redundant to install multiple times in travis without any code changes, so i added an install directive in the travis make command and removed from here.

scheduler/client/addrs.go
ryancouto: fair - that does seem misleading to me too. will find a more suitable place for them.

ryancouto:
> Overall in the right direction but I have some reservations.
> 
> The biggest thing that seems "off" to me is the domain package structure. Domain seems more like an api-related package as it just deals with converting between thrift and internal data structures. And moving all of the core scheduler functionality under it makes that less discoverable.
> 
> Without going so far as refactoring how all of the API pieces fit together, do you think you can make that clearer without polluting the main scheduler dir too much? Off the top of my head a "server" dir under scheduler would encompass all of the main component functionality. Ideally, in the future, not much outside of scheduler would need to import this package which would reduce the ambiguity of the name.

Thanks for the feedback! Makes sense, and yep will do.

common/log/helpers/helpers.go
```diff
@@ -3,7 +3,7 @@ package helpers
 import (
 	log "github.com/sirupsen/logrus"
 
-	"github.com/twitter/scoot/scootapi/gen-go/scoot"
+	"github.com/twitter/scoot/scheduler/api/thrift/gen-go/scoot"
```
JeanetteBruno, line 6: having something in common use (import) from a specific service is a red flag (tends to lead to cyclic dependencies)... is LogRunStatus() used by > 1 service?  If so, then scoot.JobStatus is used by > 1 service and should also be in common.  If LogRunStatus() is only used by scheduler, then it can be moved to scheduler.

scheduler/client/README.md
```diff
@@ -17,7 +17,7 @@ The setup of these is defined by server.Defaults(), which sets up the ice.MagicB
 * Thrift interface
 * Configuration Schema
 
-The actual implementations in scoot/scootapi/server handle Cloud Server API request handling from the Thrift interface down. The main elements here are:
```
JeanetteBruno, line 20: have a comment explaining what is offered in the client directory:
- ? main() exposing a cli that issues commands ?via thrift? to the scheduler
- ?a library of functions that issue commands ?via thrift? to the scheduler.  These functions can be used in a custom client.

scheduler/domain/README.md
```diff
@@ -5,7 +5,7 @@ API, and distributing the jobs to workers and maintaining and communicating
 the state of jobs run.
 
 The scheduler code contains the following packages:
-* __sched__ - scheduler go objects: jobs, tasks and states, and thrift versions of these objects
+* __domain__ - scheduler go objects: jobs, tasks and states, and thrift versions of these objects
   * __gen-go/sched__ - generated code from the thrift definitions
   * __scheduler__ - interfaces and implementations for job scheduling
   * __worker__ - interface for scheduler to run tasks on a worker
```
JeanetteBruno: this is a relatively simplistic statement that always confused me... is it more accurate to say: starting tasks on workers, and tracking and managing responses from workers.  (not just 'run tasks on a worker')

JeanetteBruno:
general comments
the following can be done in a future reorg:
- put bazel under apiserver and scheduler or in common
- should configs go under their service (eg: conifg/scootconfig/scheduler_config_modules.go)  or will that cause cyclic dependency?
- should saga be moved to scheduler/service/saga?  (is saga used by anything other than scheduler?)
general pattern to follow:  (the current reorg is close)
- serviceX/
- serviceX/api - just the api definitions used by both server and client (thrift files, generated files)
- serviceX/client - client implementations
- serviceX/client/serviceX_client - main() (so we get a serviceX_client binary)
- serviceX/client/http - http client
- serviceX/client/thrift - thrift client
- serviceX/server - server implementation
- serviceX/server/serviceX_server/ - main() (so we get a serviceX_server binary)
- serviceX/server/thrift - thrift entry implementation
- serviceX/server/http - http entry implementation
- common non-service specific functionality can be used by >1 service
(does that match what I put up on the whiteboard in NYC - recommend we document the overall pattern we're following in top level doc.go(?)

common/log/helpers/helpers.go
ryancouto: It's used by both the scootcl thrift client and the bzutil grpc client to get the job's status - agreed though re: service imports into common

dgdonut:
We are getting to the point where we are trying to over-pattern and organize this in a way that's forced and doesn't have clear benefits. Lets address the specific inline comments at this stage and call this done. In the future I don't think reorganizing for its own sake makes a good use of effort.

scheduler/domain/setup/api.go
```diff
@@ -2,11 +2,11 @@ package setup
 
 import (
```
dgdonut: Echoing offline comment I made, but let's move setup to the scheduler top level so domain has no subpackages.

scheduler/api/grpc/doc.go
```diff
@@ -0,0 +1,4 @@
+/*
```
dgdonut: Until we populate these we shouldn't over pattern, let's get rid of all the doc only packages.

scheduler/domain/README.md
```diff
@@ -8,5 +8,5 @@ The scheduler code contains the following packages:
 * __domain__ - scheduler go objects: jobs, tasks and states, and thrift versions of these objects
   * __gen-go/sched__ - generated code from the thrift definitions
   * __scheduler__ - interfaces and implementations for job scheduling
-  * __worker__ - interface for scheduler to run tasks on a worker
+  * __worker__ - interface for scheduler to starts tasks on a worker & manage responses re: status of in progress tasks.
```
dgdonut: s/starts/start/

dgdonut: Also re: domain - worker sub-pkg doesn't feel like it belongs here either. Maybe a top-level pkg under scheduler, or within setup (looks related to setup/workers.go). + update this readme with changes

dgdonut (approved):

PR #461, opened 2019-10-15
Return before release of nil checkout

ryancouto:
Problem

Failing to return here results in a nil pointer dereference

Solution

Return early

Result

No more nil pointers

### DIFF
--- a/runner/runners/invoke.go
+++ b/runner/runners/invoke.go
@@ -200,7 +200,12 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 			}
 			if err := <-checkoutCh; err != nil {
 				log.Errorf("Checkout errored: %s", err)
+				// If there was an error there should be no lingering gitdb locks, so return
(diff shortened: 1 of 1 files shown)


runner/runners/invoke.go
```diff
@@ -200,7 +200,12 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 			}
 			if err := <-checkoutCh; err != nil {
 				log.Errorf("Checkout errored: %s", err)
+				// If there was an error there should be no lingering gitdb locks, so return
+				// In addition, co should be nil, so failing to return and calling co.Release()
+				// will result in a nil pointer dereference
+				return
```
dgdonut, line 206: What does this end up returning - should we make a status struct to return here?

ryancouto: still returns the AbortStatus declared below - this just breaks the goroutine early

dgdonut: ah - yep, misunderstood this block, thanks!

dgdonut (approved):

PR #462, opened 2019-10-16
Add go report card badge, enforce stricter gofmt, remove templates

dgdonut:
- use -s option in gofmt and fix existing failures
- travis verifies that gofmt without -w has no output via check_fmt.sh script
- add goreportcard badge to README
- remove PR/issue templates: I think these templates actually encourage weaker PR summaries, by giving the wrong impression that completing the template results in a useful summary.


README.md
```diff
@@ -1,7 +1,8 @@
 # Scoot
+[![GoDoc](https://godoc.org/github.com/twitter/scoot?status.svg)](https://godoc.org/github.com/twitter/scoot)
 [![Build Status](https://travis-ci.org/twitter/scoot.svg?branch=master)](https://travis-ci.org/twitter/scoot)
+[![GoReportCard](https://goreportcard.com/badge/github.com/twitter/scoot)](https://goreportcard.com/report/github.com/twitter/scoot)
```
ryancouto, line 4: v cool

Makefile
```diff
@@ -23,7 +23,7 @@ build:
 	go build ./...
 
 fmt:
-	go fmt ./...
+	gofmt -l -w -s .
```
ryancouto, line 26: why not bash check_fmt.sh here?

dgdonut: we have 2 different use cases - 1) I'm a developer and I want to gofmt all the code in my branch, and 2) I'm travis and I want to verify that a branch is all gofmt compliant and fail if it's not. We currently run the same command for both cases - so travis is actually fixing any non-fmted code and proceeding, hence the need for check_fmt. As a developer, though, I don't want an error message if my code isn't formatted, I just want to fix it.

ryancouto (approved):

PR #463, opened 2019-10-17
cas,bzutil - add client FindMissingBlobs requests

dgdonut:
Adds a bazel API FindMissingBlobs function to the CAS client library, with simple mock test. This capability makes Scoot closer to being API-complete on the client, and will be used in internal testing.

Added a placeholder to bazel-integration - will add a lightweight find missing test, but so far verified with a local cluster and manual bzutil calls.

Minor:
- added a DigestsEqual comparator
- moved scripts to scripts/ dir

This will likely conflict with #457


JeanetteBruno:
Yes, this will conflict with [#457](https://github.com/twitter/scoot/pull/457).  Once #457 has landed we can/should update the find missing blob unit test to start the test at  bzutil/main.go

JeanetteBruno (commented):
do we want to wait for #457?

JeanetteBruno (approved):

dgdonut:
Yep we can let 457 merge first and I'll update this.

PR #464, opened 2019-10-17
Use groupcache.Contain to optimize HEAD request

anthonyjpratti:
This will improve performance of RPC method FindMissingBlobs.

Depends on https://github.com/twitter/groupcache/pull/11.


dgdonut:
Can you link include a link to the groupcache pr this depends on?

snapshot/bundlestore/server.go
```diff
@@ -43,11 +43,12 @@ func MakeServer(s store.Store, ttl *store.TTLConfig, stat stats.StatsReceiver, g
 // Implements http.Handler interface
 func (s *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	s.storeConfig.Stat.Counter(stats.BundlestoreRequestCounter).Inc(1)
+	//TODO(apratti): write ttl in headers in each of these
```
dgdonut: I don't think this is part of the contract this server's API guarantees

snapshot/bundlestore/http_server.go
```diff
@@ -79,6 +79,28 @@ func (s *httpServer) HandleUpload(w http.ResponseWriter, req *http.Request) {
...
+		return
+	}
+	ok, err := s.storeConfig.Store.Exists(bundleName)
+	if err != nil || !ok {
+		log.Infof("Check err: %v --> StatusNotFound (from %v)", err, req.RemoteAddr)
+		http.NotFound(w, req)
+		s.storeConfig.Stat.Counter(stats.BundlestoreCheckErrCounter).Inc(1)
+		return
+	}
+	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
```
dgdonut, line 100: why set this header?

snapshot/store/groupcache_store.go
```diff
@@ -65,6 +65,17 @@ func MakeGroupcacheStore(underlying Store, cfg *GroupcacheConfig, ttlc *TTLConfi
 			}
 			return ttl, dest.SetBytes(data)
 		}),
+		groupcache.ContainerFunc(func(ctx groupcache.Context, bundleName string) (bool, error) {
+			log.Info("Not cached, try to check bundle existence and populate cache metadata: ", bundleName)
```
dgdonut: log msg - doesn't try to populate cache metadata

snapshot/bundlestore/http_server.go
anthonyjpratti: To match the GET response, the http lib adds this automatically for GETs.

snapshot/bundlestore/server.go
anthonyjpratti: Will remove

snapshot/store/groupcache_store.go
anthonyjpratti: Will fix

dgdonut (approved):

PR #472, opened 2019-10-31
Jbruno/dpt126129deploy apiserver benchmark service

JeanetteBruno:
submitting updates to the apiserver benchmark tool (from lessons learned using it):
- ability to set logging level with each test request when deployed as a service (helps debugging)
- ability to direct the each test to a different apiserver when deployed as a service
- code cleanup (names, dead code)
- fixed stats reporting when deployed as a service: reset stats at the beginning of each test, get stats from local file so as to report stats on repeated runs of a test.

### DIFF
--- a/perftests/apiserver/loadtest/apiserver_benchmarker/main.go
+++ b/perftests/apiserver/loadtest/apiserver_benchmarker/main.go
@@ -34,7 +34,7 @@ func main() {
 	// parse the arguments
 	a, startAsServer, server, port, err := argParse()
 	if err != nil {
-		log.Fatal(err)
+		log.Errorf(err.Error())
 		return
 	}
 
@@ -52,7 +52,7 @@ func main() {
 	} else { // run the test directly from cli
 		e := lt.RunLoadTest()
 		if e != nil {
-			log.Errorf("%s", e.Error())
+			log.Fatal(e)
 		}
 	}
 
@@ -77,10 +77,10 @@ func argParse() (*loadtest.Args, bool, string, string, error) {
 	portFlag := flag.String("port", "", "The (load test) server's portFlag.")
 	flag.Parse()
 
-	level, err := log.ParseLevel(*logLevelFlag)
-	if err != nil {
-		log.Fatal(err)
-		return nil, false, "", "", fmt.Errorf("bad log level: %s", *logLevelFlag)
+	level, levelErr := log.ParseLevel(*logLevelFlag)
+	if levelErr != nil {
+		log.Error(levelErr)
+		level = log.InfoLevel // giving it a valid value so we can verify the rest of the args
 	}
 
 	a := &loadtest.Args{
@@ -94,7 +94,7 @@ func argParse() (*loadtest.Args, bool, string, string, error) {
 		CasGrpcAddr: *casGrpcAddr,
 		Batch:       *batch,
 	}
-	if !validateArgs(a, *startAsServerFlag, *portFlag) {
+	if !validateArgs(a, *startAsServerFlag, *portFlag) || levelErr != nil {
 		return a, false, "", "", fmt.Errorf("bad arugment value(s)")
 	}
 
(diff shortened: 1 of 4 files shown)


perftests/apiserver/loadtest/apiserver_benchmarker/main.go
```diff
@@ -74,16 +69,22 @@ func argParse() (*loadtest.Args, bool, string, string, error) {
...
 	totalTime := flag.Int("total_time", 30, "Total number of minutes to allow the test to run (default 30)")
 	casGrpcAddr := flag.String("cas_addr", "", "cas grpc address")
 	startAsServerFlag := flag.Bool("start_as_http_server", false, "start an http service listening for get to trigger the test. Default false.")
 	serverNameFlag := flag.String("server_name", "localhost", "name for this load test server. Default localhost")
 	portFlag := flag.String("port", "", "The (load test) server's portFlag.")
 	flag.Parse()
 
+	level, err := log.ParseLevel(*logLevelFlag)
+	if err != nil {
+		log.Fatal(err)
```
ryancouto: don't need to log Fatal here if you're returning an err

perftests/apiserver/loadtest/apiserver_benchmarker/main.go
```diff
@@ -144,11 +145,12 @@ func validateArgs(a *loadtest.Args, startAsServer bool, port string) bool {
 		err = true
```
ryancouto, line 145: Can we return the particular err that isn't valid and bubble it up to argParse()? As it currently exists, we won't even get to the end of validateArgs if one isn't valid, as log.Fatal calls an os.Exit(1)

perftests/apiserver/loadtest/load_tester.go
```diff
@@ -110,24 +129,23 @@ func MakeApiserverLoadTester(a *Args) *ApiserverLoadTester {
...
 
 	// initialize thes stats
 	statsReceiver, _ := stats.NewCustomStatsReceiver(stats.NewFinagleStatsRegistry, 0)
 	lt.stat = statsReceiver.Scope("cas_streaming")
 
-	lt.stopTestIterations = make(chan bool)
-
 	tdir := os.TempDir()
-	_, err := os.Open(fmt.Sprintf("%scloud_exec", tdir))
+	_, err := os.Open(fmt.Sprintf("%sCloudExec", tdir))
```
ryancouto: why CloudExec?

perftests/apiserver/loadtest/load_tester.go
```diff
@@ -141,62 +159,75 @@ func (lt *ApiserverLoadTester) RunLoadTest() error {
...
 
 	// run the test once
 	lt.runOneIteration()
 	if lt.getKillRequested() {
-		lt.drainStopChAndResetStatus()
+		lt.resetStatusToWaitingToStart()
 		return nil
 	}
 
 	// set up a timer to signal running the test every <freq> minutes
```
dgdonut: comment -> seconds

perftests/apiserver/loadtest/apiserver_benchmarker/main.go
JeanetteBruno: changing to error and updating logic to perform all edits

perftests/apiserver/loadtest/apiserver_benchmarker/main.go
JeanetteBruno: changed fatalf to errorf. performing all edits before returning.

perftests/apiserver/loadtest/load_tester.go
JeanetteBruno: forgot we are using scoot

perftests/apiserver/loadtest/load_tester.go
JeanetteBruno: fixed

dgdonut (approved):

PR #474, opened 2019-11-01
Timeout on exec fs_util commands and wait for ingests to return during abort

dgdonut:
This should help address a couple variants of issues we've been seeing.

The first is panics in bz_tree indicating bzCommands are being reused when that's not intended. It's clear from the invoker that asynchronous ingests can be left running in Abort scenarios, so we wait blocking after calling Cancel to make sure the worker can't run new tasks until it's finished.

We've also seen cases where fs_util has been running for ~days on end - this is rare, but essentially bricks the worker. For now, add a 30 minute timeout for all exec'd fs_util calls.

### DIFF
--- a/snapshot/bazel/constants.go
+++ b/snapshot/bazel/constants.go
@@ -32,5 +32,5 @@ const (
 	connLimit = 3
 
 	// max time execd fs_util commands can run before timing them out
-	fsUtilTimeoutSec = 1800
+	fsUtilTimeoutSec = 600
 )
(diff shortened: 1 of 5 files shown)


snapshot/bazel/constants.go
```diff
@@ -30,4 +30,7 @@ const (
 
 	// connection-limit fs_util flag setting
 	connLimit = 3
+
+	// max time execd fs_util commands can run before timing them out
+	fsUtilTimeoutSec = 1800
```
ryancouto: this seems a little high. could we go down to ~10 min?

ryancouto (approved):

snapshot/bazel/constants.go
dgdonut: I originally had this as 10 and then upped it - will take this as a sign that that was a better value

PR #475, opened 2019-11-13
kill the worker on any repeated git-based error

JeanetteBruno:
We had an incident where the worker had a stuck git index.lock file which was causing the worker to fail all tests.  We expanded the fail worker logic to include any repeated git error.

Does this include too many failure types?  Are there more that should be included?

### DIFF
--- a/common/errors/exit_codes.go
+++ b/common/errors/exit_codes.go
@@ -9,7 +9,6 @@ const (
 	GenericCheckoutFailureExitCode = 210
 	CleanFailureExitCode           = 211
 	CheckoutFailureExitCode        = 212
-	ReleaseCheckoutFailureExitCode = 213
 	BundleUploadFailureExitCode    = 214
 	ReadFileAllFailureExitCode     = 215
 	ExportGitCommitFailureExitCode = 216
--- a/common/errors/exit_codes.go
+++ b/common/errors/exit_codes.go
@@ -9,6 +9,7 @@ const (
 	GenericCheckoutFailureExitCode = 210
 	CleanFailureExitCode           = 211
 	CheckoutFailureExitCode        = 212
+	ReleaseCheckoutFailureCode     = 213
 	BundleUploadFailureExitCode    = 214
 	ReadFileAllFailureExitCode     = 215
 	ExportGitCommitFailureExitCode = 216
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -13,6 +13,7 @@ import (
 	"syscall"
 	"time"
 
+	scoot_e "github.com/twitter/scoot/common/errors"
 	"github.com/twitter/scoot/common/log/tags"
 	"github.com/twitter/scoot/common/stats"
 	"github.com/twitter/scoot/runner/execer"
@@ -386,7 +387,7 @@ func (p *osProcess) Wait() (result execer.ProcessStatus) {
 		// we can get the commands exit code
 		if status, ok := err.Sys().(syscall.WaitStatus); ok {
 			result.State = execer.COMPLETE
-			result.ExitCode = status.ExitStatus()
+			result.ExitCode = scoot_e.ExitCode(status.ExitStatus())
 			// stdout and stderr are collected and set by (invoke.go) runner
 			return result
 		}
@@ -558,7 +559,7 @@ func (p *osProcess) KillAndWait(resultError string) {
 	_, err = p.cmd.Process.Wait()
 	if err, ok := err.(*exec.ExitError); ok {
 		if status, ok := err.Sys().(syscall.WaitStatus); ok {
-			p.result.ExitCode = status.ExitStatus()
+			p.result.ExitCode = scoot_e.ExitCode(status.ExitStatus())
 		}
 	}
 }
--- a/runner/execer/os/os_execer.go
+++ b/runner/execer/os/os_execer.go
@@ -13,7 +13,7 @@ import (
 	"syscall"
 	"time"
 
-	scoot_e "github.com/twitter/scoot/common/errors"
+	scooterror "github.com/twitter/scoot/common/errors"
 	"github.com/twitter/scoot/common/log/tags"
 	"github.com/twitter/scoot/common/stats"
 	"github.com/twitter/scoot/runner/execer"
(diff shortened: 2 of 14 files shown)


runner/runners/queue.go
```diff
@@ -454,9 +454,13 @@ func (c *QueueController) runAndWatch(cmdID cmdAndID) chan runner.RunStatus {
 				}).Info("Queue received status update")
 			c.statusManager.Update(st)
 			if st.State.IsDone() {
-				if st.ExitCode == errors.CleanFailureExitCode && c.lastExitCode == errors.CleanFailureExitCode {
+				if st.ExitCode == int(c.lastExitCode) &&
+					(st.ExitCode == errors.CleanFailureExitCode ||
+						st.ExitCode == errors.CheckoutFailureExitCode ||
+						st.ExitCode == errors.GenericCheckoutFailureExitCode ||
+						st.ExitCode == errors.ReleaseCheckoutFailureExitCode) {
```
dgdonut: This isn't used to I would remove the check for it.

runner/runners/queue.go
```diff
@@ -454,9 +454,13 @@ func (c *QueueController) runAndWatch(cmdID cmdAndID) chan runner.RunStatus {
...
 					c.inv.stat.Counter(stats.WorkerServerKillGauge).Inc(1)
```
dgdonut: Noticing this now but this stat is useless, the worker goes down immediately after so it can never be read.

runner/runners/queue.go
```diff
@@ -454,9 +454,13 @@ func (c *QueueController) runAndWatch(cmdID cmdAndID) chan runner.RunStatus {
 				}).Info("Queue received status update")
 			c.statusManager.Update(st)
 			if st.State.IsDone() {
-				if st.ExitCode == errors.CleanFailureExitCode && c.lastExitCode == errors.CleanFailureExitCode {
+				if st.ExitCode == int(c.lastExitCode) &&
```
dgdonut: Noticing this now, but can we just change the type of ExitCode in runner/status.go to errors.ExitCode so we don't have to int() convert these everywhere.

dgdonut: Noticing this now (again) but this is very significant logic that's kind of hidden in this function, can we break the checks out to a separate function that tells this one whether to crash or not

runner/runners/queue.go
```diff
@@ -454,9 +454,13 @@ func (c *QueueController) runAndWatch(cmdID cmdAndID) chan runner.RunStatus {
...
+					(st.ExitCode == errors.CleanFailureExitCode ||
+						st.ExitCode == errors.CheckoutFailureExitCode ||
+						st.ExitCode == errors.GenericCheckoutFailureExitCode ||
```
dgdonut: Do we know specifically which code the lock file scenario will trigger? If it will trigger CheckoutFailureExitCode that's good, but it looks like GenericCheckoutFailureExitCode is too generic - I think the latter will get triggered if a checkout fails due to a snapshot lacking the proper commits, which could let a bad snapshot poison the system and cause a substantial worker outage.

dgdonut (commented):
Maybe the CLA changed or something and requires re-signing, can you try that and see if the CLA check passes?

runner/runners/queue.go
JeanetteBruno: do we want to sleep for 1 minute (the stats latch time) to give stats a chance to pick up this count?  This might tell us if a lot of workers are restarting due to errors.

runner/runners/queue.go
JeanetteBruno: we end up casting int to errors.ExitCode more often than we were casting errors.ExitCode to int, but the various data structures are now using errors.ExitCode more consistently.

JeanetteBruno: we end up casting int to errors.ExitCode more often than we were casting errors.ExitCode to int, but the various data structures are now using errors.ExitCode more consistently.

JeanetteBruno: put comment in wrong place....

runner/runners/queue.go
JeanetteBruno: the error was CheckoutFailureExitCode, removed GenericCheckoutFailureExitCode.

runner/execer/os/os_execer.go
```diff
@@ -13,6 +13,7 @@ import (
 	"syscall"
 	"time"
 
+	scoot_e "github.com/twitter/scoot/common/errors"
```
ryancouto: No underscores in pkg names. We could either import scooterror or just throw std lib errors stuff we're using into internal, e.g. in common/errors/errors.go throw a
`func New(string) *exitCodeError { ... }` that assigns some generic exitCode

common/errors/exit_codes.go
```diff
@@ -9,7 +9,6 @@ const (
 	GenericCheckoutFailureExitCode = 210
 	CleanFailureExitCode           = 211
 	CheckoutFailureExitCode        = 212
-	ReleaseCheckoutFailureExitCode = 213
```
ryancouto, line 12: whoops - looks like I missed implementing this one. In snapshot/git/gitdb/checkout.go releaseCheckout we should be returning this error via `errors.NewError(err, errors.ReleaseCheckoutFailureExitCode)` instead of just err

runner/runners/queue.go
```diff
@@ -466,3 +467,15 @@ func (c *QueueController) runAndWatch(cmdID cmdAndID) chan runner.RunStatus {
 	}()
 	return watchCh
 }
+
+/*
+killForPersistentError returns true when one of the critical errors has occurred 2 times in a row.
+*/
+func (c *QueueController) killForPersistenError(exitCode errors.ExitCode) bool {
+	if exitCode == c.lastExitCode &&
```
ryancouto: could just `return  exitCode == c.lastExitCode &&
		(exitCode == errors.CleanFailureExitCode ||
			exitCode == errors.CheckoutFailureExitCode)`

runner/runners/queue.go
dgdonut: ok, didn't remember how many execer things were using int, but that's good that it's more consistent

runner/runners/queue.go
dgdonut: For now I'd recommend no, we would probably have to sleep longer to make sure obs had time to pick up the emitted metric to avoid flakiness. It would make sense to try to monitor all worker instance failures internally and not just from this cause if we were to spend more time on this.

snapshot/git/gitdb/db.go
```diff
@@ -15,6 +15,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// TODO verify that we really use this code - is this only used by FSSnapshot?  Are we using FSSnapshot?
```
dgdonut: this code is very heavily used

dgdonut (approved):

PR #479, opened 2020-03-10
Add batch saga update APIs to speed up scheduler kill jobs

dgdonut:
The scheduler's processing of kill job requests in the case of jobs with large numbers of unstarted tasks is very bad - O(N) where N is the number of unstarted tasks. For each N, the scheduler applies separate start and end task saga updates, each of which independently write-lock the Saga data structure as well as the entire in-memory saga log. This is now O(1). When unstarted task counts are >~ 10,000 this results in a perf increase of over two orders of magnitude, as show by go test -bench of the added Benchmark.

I will resist substantial cleanup or refactor of the saga or other APIs until we are able to dedicate the time to properly overhaul them.


saga/saga.go
```diff
@@ -196,20 +213,25 @@ type sagaUpdate struct {
...
+		if err != nil {
+			return oldState, err
+		}
 	}
 
-	//try durably storing the message
-	err = log.LogMessage(msg)
+	// try durably storing the message
+	if len(msgs) == 1 {
+		err = log.LogMessage(msgs[0])
```
dgdonut, line 224: just using LogBatchMessages breaks like 20 million mock tests and I am refusing to fix them

scheduler/server/stateful_scheduler_test.go
```diff
@@ -594,6 +592,32 @@ func Test_StatefulScheduler_GetSomeThrottledStatus(t *testing.T) {
...
+		duration: 5 * time.Second,
+	}
+	taskDurations["foo"].update(21 * time.Second)
+	taskDurations["foo"].update(25 * time.Second)
+	if taskDurations["foo"].duration != 17*time.Second {
+		t.Fatalf("Expected 17 seconds, got %v", taskDurations["foo"].duration)
+	}
+}
+
+func BenchmarkProcessKillJobsRequests(b *testing.B) {
```
ryancouto, line 608: `*testing.B`
oooh cool

saga/saga.go
```diff
@@ -144,24 +149,36 @@ func (s *Saga) StartCompensatingTask(taskId string, data []byte) error {
 // Returns an error if it fails
 //
 func (s *Saga) EndCompensatingTask(taskId string, results []byte) error {
-	return s.updateSagaState(MakeEndCompTaskMessage(s.id, taskId, results))
+	return s.updateSagaState([]SagaMessage{MakeEndCompTaskMessage(s.id, taskId, results)})
+}
+
+//
```
ryancouto: extra //

ryancouto (approved):

saga/saga.go
```diff
@@ -144,24 +149,36 @@ func (s *Saga) StartCompensatingTask(taskId string, data []byte) error {
...
 
 	result := <-resultCh
 
 	// after we successfully log an EndSaga message close the channel
 	// no more messages should be logged
-	if msg.MsgType == EndSaga {
-		close(s.updateCh)
+	for _, msg := range msgs {
+		if msg.MsgType == EndSaga {
+			close(s.updateCh)
```
JeanetteBruno, line 173: can we break here?

saga/saga.go
JeanetteBruno: can we just remove the call to log.LogMessage() - we aren't really doing an durable store anyway.  (same as above) (and remove the broken tests?)

dgdonut: I don't follow, we are doing a durable store here.

saga/saga.go
dgdonut: removed all of these

saga/saga.go
dgdonut: yep, done

saga/saga.go
dgdonut: I just spent a while trying to consolidate the mock based tests and it's too painful, deferring it to wider refactor

PR #489, opened 2020-11-03
added load based scheduling algorithm

JeanetteBruno:
this replaces the jbruno/schedulerAlgorithm pull request.

I have not implemented the rebalancing yet.

This code implements the computation demonstrated in https://docs.google.com/spreadsheets/d/1QP-okxrVkmKBMtc4OzWe8cFU15u84b9aXENeKqGogbc/edit#gid=554928957 (the allocations tab)


scheduler/api/thrift/load_based_alg_settings.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"github.com/twitter/scoot/scheduler/server"
+)
+
+/*
+GetClassLoadPcts get the target load pcts for the classes
+*/
+func GetClassLoadPcts(scheduler server.Scheduler) map[string]int32 {
+
```
dgdonut: nit: blank lines on these

scheduler/server/requestor_counts_test.go
```diff
@@ -15,6 +15,7 @@ func Test_StatefulSchedulerRequestorCounts(t *testing.T) {
 	log.SetLevel(log.ErrorLevel)
 	sc := sagalogs.MakeInMemorySagaCoordinatorNoGC()
 	s, _, statsRegistry := initializeServices(sc, false)
+	s.SetSchedulingAlg(&OrigSchedulingAlg{})
```
dgdonut, line 301: Is this test only relevant to the orig sched algorithm? If so can we rename the file and test function so that's clear? (if not can we leverage this test w/ the new alg as well?)

scheduler/client/client.go
```diff
@@ -166,6 +167,72 @@ func (c *CloudScootClient) GetSchedulerStatus() (*scoot.SchedulerStatus, error)
...
+		c.closeConnection()
+	}
+	return classLoadPcts, err
+}
+
+/*
+SetClassLoadPcts set the target worker load % for each job class
+*/
+func (c *CloudScootClient) SetClassLoadPcts(classLoads map[string]int32) error {
+	err := c.checkForClient()
```
dgdonut: nit but could condense these blocks into: if err := c.checkForClient(); err != nil { ...

scheduler/client/client.go
```diff
@@ -166,6 +167,72 @@ func (c *CloudScootClient) GetSchedulerStatus() (*scoot.SchedulerStatus, error)
...
+	if err != nil {
+		return err
+	}
+
+	err = c.client.SetClassLoadPcts(classLoads)
```
dgdonut: can just return c.client.* in both set fns

scheduler/server/scheduler.go
```diff
@@ -23,9 +23,19 @@ type Scheduler interface {
 	SetSchedulerStatus(maxTasks int) error
 
 	GetSchedulerStatus() (int, int)
+
+	GetClassLoadPcts() map[string]int32
+
+	SetClassLoadPcts(classLoads map[string]int32)
```
ryancouto: why doesn't this return an error?

scheduler/api/server.go
```diff
@@ -78,3 +78,25 @@ func (h *Handler) GetSchedulerStatus() (*scoot.SchedulerStatus, error) {
...
+
+// Implements GetClassLoadPcts Cloud Scoot API
+func (h *Handler) GetClassLoadPcts() (map[string]int32, error) {
+	return schedthrift.GetClassLoadPcts(h.scheduler), nil
+}
+
+// Implements SetClassLoadPcts Cloud Scoot API
+func (h *Handler) SetClassLoadPcts(classLoadPcts map[string]int32) error {
+	schedthrift.SetClassLoadPcts(h.scheduler, classLoadPcts)
+	return nil
```
ryancouto: why not just return an err from SetClassLoadPcts and return that here?

scheduler/api/server.go
```diff
@@ -78,3 +78,25 @@ func (h *Handler) GetSchedulerStatus() (*scoot.SchedulerStatus, error) {
...
+
+// Implements GetRequestorToClassMap Cloud Scoot API
+func (h *Handler) GetRequestorToClassMap() (map[string]string, error) {
+	return schedthrift.GetRequestorToClassMap(h.scheduler), nil
+}
+
+// Implements SetRequestorToClassMap Cloud Scoot API
+func (h *Handler) SetRequestorToClassMap(requestToClassMap map[string]string) error {
+	schedthrift.SetRequestorToClassMap(h.scheduler, requestToClassMap)
+	return nil
```
ryancouto: ditto re: returning an err

scheduler/api/thrift/load_based_alg_settings.go
ryancouto: nit: remove newlines

scheduler/client/client.go
```diff
@@ -166,6 +167,72 @@ func (c *CloudScootClient) GetSchedulerStatus() (*scoot.SchedulerStatus, error)
 	return schedulerStatus, err
 }
 
+/*
+GetClassLoadPcts get the target load pcts for the classes
```
ryancouto: nit: no need for multiline comments

scheduler/server/job_state.go
```diff
@@ -22,10 +24,13 @@ type jobState struct {
...
 
 	// track tasks by state (completed, running, not started) for priority based scheduling algorithm
-	// TODO remove if we don't use the priority based scheduling algorithm
+	// Completed and Running are only used by priority based scheduling algorithm
 	Completed  map[string]*taskState
 	Running    map[string]*taskState
 	NotStarted map[string]*taskState
+
+	jobClass                       string
+	tasksByJobClassAndStartTimeSec map[string]map[time.Time]map[string]*taskState
```
ryancouto: is there any way we can simplify this field? the nesting of the triple map is confusing imo

scheduler/server/job_state.go
```diff
@@ -170,12 +195,16 @@ func (j *jobState) taskCompleted(taskId string, running bool) {
 		delete(j.NotStarted, taskId)
 	}
 	j.Completed[taskId] = taskState
+
+	// remove the task from the map of tasks by start time
+	j.removeTaskFromStartTimeMap(taskState.JobId, taskId, startTimeSec)
```
ryancouto, line 206: is this tasksByJobClassAndStartTimeSec just in progress tasks?

scheduler/server/job_state.go
```diff
@@ -197,3 +228,41 @@ func (j *jobState) getJobStatus() domain.Status {
 	}
 	return domain.InProgress
 }
+
+func (j *jobState) addTaskToStartTimeMap(jobClass string, task *taskState, startTimeSec time.Time) {
+	if j.tasksByJobClassAndStartTimeSec == nil {
+		return
```
ryancouto, line 240: should we log something here?

scheduler/server/orig_scheduling_alg.go
```diff
@@ -7,10 +7,47 @@ import (
...
+	}
+
+	// Udate SoftMaxSchedulableTasks based on number of healthy nodes and the total number of tasks.
+	// Setting the max to num healthy nodes means that each job can be fully scheduled.
+	// (This gets used in config.GetNodeScaleFactor())
+	var cfg OrigSchedulingAlgConfig
+	if oa.Config != nil {
+		cfg = *oa.Config
+		if cfg.SoftMaxSchedulableTasks == 0 {
+			cfg.SoftMaxSchedulableTasks = DefaultSoftMaxSchedulableTasks
```
ryancouto: does this actually modify oa.Config? or just the dereferenced value cfg

scheduler/server/scheduler.go
```diff
@@ -23,9 +23,19 @@ type Scheduler interface {
...
+
+	GetRequestorToClassMap() map[string]string
+
+	SetRequestorToClassMap(requestorToClassMap map[string]string)
```
ryancouto: ditto

scheduler/server/scheduler.go
JeanetteBruno: adding

scheduler/api/server.go
JeanetteBruno: adding

scheduler/api/server.go
JeanetteBruno: adding

scheduler/server/job_state.go
JeanetteBruno: converting to nested structs - hopefully that is better

scheduler/server/job_state.go
JeanetteBruno: yes - these are the tasks that can be stopped for rebalancing

scheduler/server/job_state.go
JeanetteBruno: this would only happen if/when we use a different scheduling algorithm - so if we revert to the orig scheduling algorithm, this logging will spam the logs.

scheduler/server/orig_scheduling_alg.go
JeanetteBruno: reverted to match the code pattern in the original scheduling algorithm - updated version will not modify oa.Config

scheduler/server/scheduler.go
JeanetteBruno: done

scheduler/server/job_state.go
```diff
@@ -42,6 +47,23 @@ type taskState struct {
 
 type taskStatesByDuration []*taskState
 
+// the following types are used to access task State objects by class, startTime, taskID
+type taskStateByTaskID map[string]*taskState
+type taskTimeBuckets map[time.Time]taskStateByTaskID
+type tasksByClassAndStartTimeSec map[string]taskTimeBuckets
+
+// GetTasksByJobClassAndStart get the tasks for the given class with start time matching the input
+func GetTasksByJobClassAndStart(tasksByJobClassAndStartTimeSec tasksByClassAndStartTimeSec, class string, start time.Time) map[string]*taskState {
```
dgdonut: this type/var name combination are extremely confusing - can we substantially shorten or differentiate one of them?

scheduler/server/job_state.go
```diff
@@ -42,6 +47,23 @@ type taskState struct {
 
 type taskStatesByDuration []*taskState
 
+// the following types are used to access task State objects by class, startTime, taskID
+type taskStateByTaskID map[string]*taskState
```
dgdonut: we have a lot of uses of map[string]*taskState that don't use this type alias - should we convert them all?

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,723 @@
...
+// When there are not enough tasks to use all of the class's entitled workers, the algorithm
+// will allow other classes to run their tasks on the unused workers.  We refer to
+// this as loaning worker to other classes.
+// Note: workers are not assigned to specific classes.  The class % concept is simply a counting
+// mechanism.
+//
+// Each scheduling iteration tries to bring the task allocation back to the original class
+// entitlement as defined in the class %s. It could be the case that long running tasks slowly create
+// an imbalance in the worker to class numbers (long running tasks accumulate loaned workers).
+// As such, the algorithm periodically rebalances the running workers back toward the original target
```
dgdonut, line 61: This behavior needs to be configurable to the extent that it can be disabled completely. We can't assume that rebalancing will be as effective as we think.

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,723 @@
...
+		}
+		targetNumWorkers := int(math.Floor(float64(numWorkers) * float64(jc.origTargetLoadPct) / 100.0))
+		jc.origNumTargetedWorkers = targetNumWorkers
+		totalWorkers += targetNumWorkers
+	}
+	lbs.jobClasses[lbs.classByDescLoadPct[0]].origNumTargetedWorkers = numWorkers - totalWorkers
+}
+
+// initJobClassesMap builds the map of requestor (class name) to jobClass objects
+// if we see a job whose class % not defined, assign the job to the class with the
```
dgdonut, line 228: Why default to the "least number of workers" class? We could have a critical class configured with ~10% that gets interrupted by this. Can we have the notion of a default or reserve class that gets the % not allocated in the config mapping and these get dumped into that bucket?

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,723 @@
...
+		if className != "" {
+			jc, ok = lbs.jobClasses[className]
+			if !ok {
+				// the class name was not recognized, use the lowest priority class
+				lbs.config.stat.Counter(stats.SchedLBSUnknownJobCnt).Inc(1)
+				jc = lbs.jobClasses[classNameWithLeastWorkers]
+				log.Errorf("%s is not a recognized job class assigning to lowest priority class (%s)", className, classNameWithLeastWorkers)
+			}
+		} else {
+			// if the requestor is not recognized, use the lowest priority class
```
dgdonut: we should get away from using the term "priority" throughout this algorithm's internals, as we're not defining levels of prioritization

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,723 @@
...
+			}
+			jc.jobsByNumRunningTasks[job.TasksRunning] = append(jc.jobsByNumRunningTasks[job.TasksRunning], jobWaitingTaskIds{jobState: job, waitingTaskIDs: waitingTaskIds})
+			if job.TasksRunning > jc.maxTaskRunningMapIndex {
+				jc.maxTaskRunningMapIndex = job.TasksRunning
+			}
+			jc.origNumRunningTasks += job.TasksRunning
+			jc.origNumWaitingTasks += len(job.NotStarted)
+		}
+
+		jc.numWaitingTasks = jc.origNumWaitingTasks
```
dgdonut, line 278: what does this assignment do? the jc var is about to get overwritten in the next loop iteration

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,723 @@
...
+	})
+	lbs.config.classByDescLoadPct = keys
+
+	// set the load pcts - normalizing them if the don't add up to 100
+	lbs.config.classLoadPcts = map[string]int{}
+	pctTotal := 0
+	for _, val := range classLoadPcts {
+		pctTotal += int(val)
+	}
+	if pctTotal != 100 {
```
dgdonut, line 711: Why not just reject this? (see also other comment about having a "reserve" bucket/class)

scheduler/server/scheduler.go
```diff
@@ -23,9 +23,19 @@ type Scheduler interface {
 	SetSchedulerStatus(maxTasks int) error
 
 	GetSchedulerStatus() (int, int)
+
+	GetClassLoadPcts() (map[string]int32, error)
```
dgdonut, line 27: We're kind of mixing the idea of replaceable algorithms and fully migrating to the class-based alg. At this point we've basically changed the API to require the use of a class-based alg. Maybe once this is finalized, we don't care and we get rid of the "orig" alg altogether.

JeanetteBruno: agreed - I think at this point, we could still switch between the original and new algorithm ... should we be targeting being able to switch back to the old algorithm without redeploying?

perftests/scheduler_simulator/test_alg.go
```diff
@@ -173,22 +173,23 @@ func (st *SchedulingAlgTester) RunTest() error {
...
-			// pause to simulate the frequency in which the jobs arrived in production
-			deltaFromStart, e := st.extractWaitDurationFromJobDef(jobDef)
-			if e != nil {
-				return fmt.Errorf("Couldn't get deltaStartDuration:%s, skipping job", e.Error())
-			}
+			// // pause to simulate the frequency in which the jobs arrived in production
+			// deltaFromStart, e := st.extractWaitDurationFromJobDef(jobDef)
+			// if e != nil {
+			// 	return fmt.Errorf("Couldn't get deltaStartDuration:%s, skipping job", e.Error())
+			// }
```
anthonyjpratti, line 187: Why is this commented?

perftests/scheduler_simulator/test_alg.go
```diff
@@ -284,37 +285,35 @@ func (st *SchedulingAlgTester) watchForAllDone(allJobsStartedCh chan bool,
...
-	return time.Duration(d), nil
-}
+// // extract the time from Basis field
+// func (st *SchedulingAlgTester) extractWaitDurationFromJobDef(jobDef *domain.JobDefinition) (time.Duration, error) {
+// 	d, e := strconv.Atoi(jobDef.Basis)
+// 	if e != nil {
+// 		return time.Duration(0), fmt.Errorf("couldn't parse duration from job def basis:%s", e.Error())
+// 	}
+// 	return time.Duration(d), nil
+// }
```
anthonyjpratti, line 316: Why is this commented?

scheduler/api/server.go
```diff
@@ -31,50 +32,106 @@ func MakeServer(handler scoot.CloudScoot,
...
+// the rebalance threshold before rebalancing.  <= 0 implies no rebalancing
+func (h *Handler) GetRebalanceMinDuration() (int32, error) {
+	d, err := schedthrift.GetRebalanceMinDuration(h.scheduler)
+	log.Infof("GetRebalanceMinDuration returning: %d, err:%s", d, err)
+	return d, err
+}
+
+// SetRebalanceMinDuration set the duration(minutes) that the scheduler needs to be exceeding
+// the rebalance threshold before rebalancing.  <= 0 implies no rebalancing
+func (h *Handler) SetRebalanceMinDuration(durationMin int32) error {
```
anthonyjpratti, line 119: Could durationMin be a time.Duration? We could then convert that to an int32 if that's require by thrift.

scheduler/api/thrift/scoot.thrift
```diff
@@ -158,4 +158,32 @@ service CloudScoot {
     1: InvalidRequest ir
     2: ScootServerError err
   )
+  map<string, i32> GetClassLoadPcts() throws (
```
anthonyjpratti, line 161: Is Pcts a shortening of Percents? I recommend avoiding shortenings for public apis.

scheduler/client/cli/cli.go
```diff
@@ -23,6 +24,18 @@ type simpleCLIClient struct {
 	scootClient *client.CloudScootClient
 }
 
+// returnError extend the error with Invalid Request, Scoot server error, or Error getting status message
+func getReturnError(err error) error {
```
anthonyjpratti, line 28: Mismatch between comment and function name

scheduler/client/cli/sched_alg_params.go
```diff
@@ -0,0 +1,170 @@
...
+		fmt.Println("Rebalance Duration:", g.params.RebalanceMinDuration, " (minutes)")
+		log.Infof("Rebalance Threshold:%d", g.params.RebalanceThreshold)
+		fmt.Println("Rebalance Threshold:", g.params.RebalanceThreshold)
+
+	}
+
+	return nil
+}
+
+func (s *lbsSchedAlgParams) registerFlags() *cobra.Command {
```
anthonyjpratti, line 94: IMO, modules shouldn't have 2 registerFlags functions. Could you explain the need for two?

perftests/scheduler_simulator/test_alg.go
JeanetteBruno: removed in pr 490

perftests/scheduler_simulator/test_alg.go
JeanetteBruno: removed in pr 490

scheduler/client/cli/sched_alg_params.go
JeanetteBruno: we're starting to get a lot of subcommands (with 10 files for handling them).  It seems natural to me to group the set/get pairs together rather than have them in 2 different files.

JeanetteBruno:
use pr 490
I've updated it as per comments except for:
- Drew's comments about not normalizing %'s to 100, and not maintaining the ability to switch back to the old algorithm
- Anthony's comment about not putting to registerFlags functions in one file

PR #490, opened 2020-11-19
add new scheduler

JeanetteBruno:
This contains the same changes as pull request: added load based scheduling algorithm #489.  I'm creating this pull request in hopes that travis will behave better with a clean pull request. (Travis does not seem to be building for pr 489, I changed a travis integration setting after creating 489, so that may have some impact on how it is processing 489's changes.)


common/stats/stats_names.go
```diff
@@ -321,6 +321,47 @@ const (
 	*/
 	SchedScheduleTasksLatency_ms = "schedScheduleTasksLatency_ms"
 
+	/*--------------------- load based scheduler stats ---------------------------*/
+	/*
+		number of time Load Based Scheduler saw an unrecognized requestor
+	*/
+	SchedLBSUnknownJobCnt = "schedLBSUnknownJobCnt"
```
anthonyjpratti: Why Cnt instead of Counter? Seems the convention is Counter.

anthonyjpratti:
Git commit titles should be in present tense.

"Add new scheduler"

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,780 @@
...
+	origNumWaitingTasks    int
+	origNumRunningTasks    int
+	origTargetLoadPct      int // the target % of workers for this class
+	origNumTargetedWorkers int // the original number of workers allocated for this class by target % (total workers * origTargetLoadPct)
+
+	numTasksToStart int // number of tasks that can be started (when negative -> number of tasks to stop)
+	numWaitingTasks int // number of tasks still waiting to be started
+
+	tempEntitlement   int // temporary field to hold intermediate entitled num workers
+	tempNormalizedPct int // temporary field to hold the normalized load %
```
anthonyjpratti: Please match above convention of comment before field.

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,780 @@
...
+		jc.tempNormalizedPct)
+}
+
+// NewJobClass a job class with its target % worker load
+func NewJobClass(name string, targetLoadPct int) *jobClass {
+	return &jobClass{className: name, origTargetLoadPct: targetLoadPct, jobsByNumRunningTasks: map[int][]jobWaitingTaskIds{}}
+}
+
+// GetTasksToBeAssigned - the entry point to the load based scheduling algorithm
+// It returns the list of tasks that should be assigned to nodes as per the worker-class %s:
```
anthonyjpratti: Consider rephrasing "assigned to nodes as per the worker-class %s:" as  "assigned to nodes according to worker-class"

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,780 @@
...
+// Allocate available workers to classes based on target % allocations for each class and the current and
+// running load for each class.
+// - When a class has tasks waiting to start, the algorithm will determing the number of workers the class it 'entitled' to:
```
anthonyjpratti: Grammatical error please rephrase

perftests/scheduler_simulator/test_alg.go
```diff
@@ -69,18 +69,20 @@ type timeSummary struct {
...
+	testsEnd            time.Time
+	realStart           time.Time
+	jobDefsMap          map[int][]*domain.JobDefinition
+	pRatios             []int
+	clusterSize         int
+	finishTimeFilename  string
+	comparisonMap       map[string]*timeSummary
+	comparisonMapMu     sync.RWMutex
+	timeout             time.Duration
+	classLoadPercents       map[string]int32
```
anthonyjpratti: Please format.

perftests/scheduler_simulator/test_alg.go
```diff
@@ -107,28 +109,26 @@ func MakeSchedulingAlgTester(testsStart, testsEnd time.Time, jobDefsMap map[int]
...
-		pRatios:            pRatios,
-		clusterSize:        clusterSize,
+		statsFileName:       statsFile,
+		finishTimeFilename:  finishTimesFilename,
+		realStart:           time.Now(),
+		testsStart:          testsStart,
+		testsEnd:            testsEnd,
+		jobDefsMap:          jobDefsMap,
+		clusterSize:         clusterSize,
+		classLoadPercents:       classLoadPercents,
```
anthonyjpratti: Please format.

perftests/scheduler_simulator/test_alg.go
```diff
@@ -166,32 +164,29 @@ func (st *SchedulingAlgTester) RunTest() error {
...
-				return fmt.Errorf("Couldn't get deltaStartDuration:%s, skipping job", e.Error())
-			}
-			n := time.Now()
-			startTime := shadowStart.Add(deltaFromStart)
-			if startTime.After(n) {
-				time.Sleep(startTime.Sub(n)) // this pause emulates the jobs' run frequency
+			jobStart := simStart.Add(time.Duration(secondsAfterStart) * time.Second)
+			if time.Now().Before(jobStart) {
+				log.Warnf("now: %s, job start: %s", time.Now().Format(time.RFC3339), jobStart.Format(time.RFC3339))
+				<-time.After(jobStart.Sub(time.Now())) // wait till time reaches startTime0				<-
```
anthonyjpratti: Why are you using time.After instead of time.Sleep?

perftests/scheduler_simulator/test_alg.go
```diff
@@ -166,32 +164,29 @@ func (st *SchedulingAlgTester) RunTest() error {
...
 			}
 
 			// give the job to the scheduler
+			log.Warnf("%s submitting job:jobType:%s, req:%s, tag:%s, basis: %s, tasks:%d", time.Now().Format(time.RFC3339), jobDef.JobType, jobDef.Requestor, jobDef.Tag, jobDef.Basis, len(jobDef.Tasks))
```
anthonyjpratti: Does jobDef have a String method? Might make this clearer.

perftests/scheduler_simulator/test_alg.go
```diff
@@ -339,36 +323,37 @@ func (st *SchedulingAlgTester) getExternals(clusterSize int) *externalDeps {
...
-		runningCnt := s[runningStatName]
-		waitingCnt := s[waitingStatName]
-		line = append(line, []byte(fmt.Sprintf("%s, job:%s, priority:%d, running:%v, waiting:%v, ",
-			timePP, req, priority, runningCnt, waitingCnt))...)
+	for className, loadPct := range st.classLoadPercents {
+		runningStatName := fmt.Sprintf("schedNumRunningTasksGauge_%s", className)
+		waitingStatName := fmt.Sprintf("schedNumWaitingTasksGauge_%s", className)
+		runningCnt, ok1 := s[runningStatName]
+		waitingCnt, ok2 := s[waitingStatName]
+		if ok1 || ok2 {
```
anthonyjpratti: Are you sure you want an or? Wouldn't that mean runningStatName or waitingStatName might be nil?

scheduler/api/server.go
```diff
@@ -78,3 +81,60 @@ func (h *Handler) GetSchedulerStatus() (*scoot.SchedulerStatus, error) {
...
+
+// SetRequestorToClassMap Implements SetRequestorToClassMap Cloud Scoot API
+func (h *Handler) SetRequestorToClassMap(requestToClassMap map[string]string) error {
+	log.Infof("SetRequestorToClassMap to %v", requestToClassMap)
+	return schedthrift.SetRequestorToClassMap(h.scheduler, requestToClassMap)
+}
+
+// GetRebalanceMinDuration get the duration(minutes) that the scheduler needs to be exceeding
+// the rebalance threshold before rebalancing.  <= 0 implies no rebalancing
+func (h *Handler) GetRebalanceMinDuration() (int32, error) {
```
anthonyjpratti: This is ambiguous. Does min mean minimum or minute in this context?

scheduler/client/cli/sched_alg_params.go
```diff
@@ -0,0 +1,174 @@
...
+
+	clpFilePath    string
+	reqMapFilePath string
+}
+
+// getLBSSchedAlgParams structure for getting the load based scheduling params.
+type getLBSSchedAlgParams struct {
+	printAsJSON bool
+	params      lbsSchedAlgParams
+}
```
anthonyjpratti: Why is this needed?

scheduler/client/cli/cli.go
```diff
@@ -23,6 +24,18 @@ type simpleCLIClient struct {
 	scootClient *client.CloudScootClient
 }
 
+// getReturnError extend the error with Invalid Request, Scoot server error, or Error getting status message
+func getReturnError(err error) error {
```
anthonyjpratti: I think returnError is clear enough, and is the convention we use internally.

scheduler/client/cli/sched_alg_params.go
```diff
@@ -0,0 +1,174 @@
...
+		fmt.Println("Requestor (reg exp) to class map:")
+		for requestorRe, class := range g.params.RequestorMap {
+			log.Infof("%s:%s", requestorRe, class)
+			fmt.Println(requestorRe, ":", class)
+		}
+		log.Infof("Rebalance Duration:%d (minutes)", g.params.RebalanceMinDuration)
+		fmt.Println("Rebalance Duration:", g.params.RebalanceMinDuration, " (minutes)")
+		log.Infof("Rebalance Threshold:%d", g.params.RebalanceThreshold)
+		fmt.Println("Rebalance Threshold:", g.params.RebalanceThreshold)
+
```
anthonyjpratti, line 89: nit: no newline

scheduler/client/cli/sched_alg_params.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	return nil
+}
+
+func (s *lbsSchedAlgParams) registerFlags() *cobra.Command {
```
anthonyjpratti: Why isn't this named setLbsSchedAlgParams

scheduler/client/client.go
```diff
@@ -35,10 +36,9 @@ func NewCloudScootClient(config CloudScootClientConfig) *CloudScootClient {
 }
 
 // RunJob API. Schedules a Job to run asynchronously via CloudExecScoot based on
-//the specified job. If successful the JobId is returned if not an error.
+//the specified job. If successful the jobId is returned if not an error.
 func (c *CloudScootClient) RunJob(jobDef *scoot.JobDefinition) (r *scoot.JobId, err error) {
-	err = c.checkForClient()
-	if err != nil {
+	if err := c.checkForClient(); err != nil {
```
anthonyjpratti: These stylistic changes are good changes, but do make it harder to focus on the important changes in this CL.

scheduler/client/client.go
```diff
@@ -166,6 +161,98 @@ func (c *CloudScootClient) GetSchedulerStatus() (*scoot.SchedulerStatus, error)
...
 }
 
+// GetClassLoadPercents get the target load pcts for the classes
+func (c *CloudScootClient) GetClassLoadPercents() (map[string]int32, error) {
+	if err := c.checkForClient(); err != nil {
+		return nil, err
+	}
+	classLoadPercents, err := c.client.GetClassLoadPercents()
+	// if an error occurred reset the connection, could be a broken pipe or other
+	// unrecoverable error.  reset connection so a new clean one gets created
```
anthonyjpratti: nit: two spaces are error should only be one

scheduler/client/client.go
```diff
@@ -166,6 +161,98 @@ func (c *CloudScootClient) GetSchedulerStatus() (*scoot.SchedulerStatus, error)
 	return schedulerStatus, err
 }
 
+// GetClassLoadPercents get the target load pcts for the classes
+func (c *CloudScootClient) GetClassLoadPercents() (map[string]int32, error) {
+	if err := c.checkForClient(); err != nil {
+		return nil, err
+	}
+	classLoadPercents, err := c.client.GetClassLoadPercents()
+	// if an error occurred reset the connection, could be a broken pipe or other
```
anthonyjpratti, line 175: Why don't we do this for all methods?

scheduler/client/scootcl/main.go
```diff
@@ -17,6 +17,8 @@ import (
 //		run_smoke_test
 //		get_scheduler_status
 //		set_scheduler_status [limit]
+//      get_scheduling_alg_params
+//      set_scheduling_alg_params
```
anthonyjpratti: Please format. Looks like others used tabs over spaces.

scheduler/server/job_state.go
```diff
@@ -22,10 +26,13 @@ type jobState struct {
 	TimeMarker     time.Time    //when was this job last marked (i.e. for reporting purposes)
 
 	// track tasks by state (completed, running, not started) for priority based scheduling algorithm
-	// TODO remove if we don't use the priority based scheduling algorithm
-	Completed  map[string]*taskState
-	Running    map[string]*taskState
-	NotStarted map[string]*taskState
+	// Completed and Running are only used by priority based scheduling algorithm
```
anthonyjpratti: I saw pratios being removed. Are we not removing the priority based impl? If we don't remove that implementation, we should keep the tests associated with it, right?

scheduler/server/job_state.go
```diff
@@ -42,6 +49,22 @@ type taskState struct {
...
+// getTasksByJobClassAndStart get the tasks for the given class with start time matching the input
+func getTasksByJobClassAndStart(classTasksByStartTime tasksByClassAndStartTimeSec, class string, start time.Time) taskStateByTaskID {
+	if _, ok := classTasksByStartTime[class]; !ok {
+		return nil
+	}
+	if tasks, ok := classTasksByStartTime[class][start]; ok {
+		return tasks
+	}
+
+	return nil
```
anthonyjpratti: Map is a reference type, so these checks are not needed. If class is not in classTasksByStartTime, then classTasksByStartTime[class] will return nil, which means that classTasksByStartTime[class][start] will return nil. so this entire func can be a one liner.

scheduler/server/job_state.go
```diff
@@ -170,12 +201,16 @@ func (j *jobState) taskCompleted(taskId string, running bool) {
 		delete(j.NotStarted, taskId)
```
anthonyjpratti, line 195: That TODO on the prev line is worrying...

scheduler/server/job_state.go
```diff
@@ -197,3 +234,43 @@ func (j *jobState) getJobStatus() domain.Status {
 	}
 	return domain.InProgress
 }
+
+// addTaskToStartTimeMap add the running task to the map that bins running tasks by their start time
+func (j *jobState) addTaskToStartTimeMap(jobClass string, task *taskState, startTimeSec time.Time) {
+	if j.tasksByJobClassAndStartTimeSec == nil {
+		return
+	}
```
anthonyjpratti: Is this an ok state for this to be in?

scheduler/server/job_state.go
```diff
@@ -197,3 +234,43 @@ func (j *jobState) getJobStatus() domain.Status {
...
+	if _, ok := j.tasksByJobClassAndStartTimeSec[jobClass]; !ok {
+		j.tasksByJobClassAndStartTimeSec[jobClass] = taskTimeBuckets{}
+	}
+	if _, ok := j.tasksByJobClassAndStartTimeSec[jobClass][startTimeSec]; !ok {
+		j.tasksByJobClassAndStartTimeSec[jobClass][startTimeSec] = taskStateByTaskID{}
+	}
+	j.tasksByJobClassAndStartTimeSec[jobClass][startTimeSec][fmt.Sprintf("%s_%s", task.JobId, task.TaskId)] = task
```
anthonyjpratti: Keys can be structs, please prefer that over concatenation.

scheduler/server/job_state.go
```diff
@@ -197,3 +234,43 @@ func (j *jobState) getJobStatus() domain.Status {
...
+
+// addTaskToStartTimeMap add the running task to the map that bins running tasks by their start time
+func (j *jobState) addTaskToStartTimeMap(jobClass string, task *taskState, startTimeSec time.Time) {
+	if j.tasksByJobClassAndStartTimeSec == nil {
+		return
+	}
+	if _, ok := j.tasksByJobClassAndStartTimeSec[jobClass]; !ok {
+		j.tasksByJobClassAndStartTimeSec[jobClass] = taskTimeBuckets{}
+	}
+	if _, ok := j.tasksByJobClassAndStartTimeSec[jobClass][startTimeSec]; !ok {
```
anthonyjpratti: Looking up on jobClass is unnecessary. if we do an early return when ok is true, we can then preserve the reference to both taskTimeBuckets and taskStateByTaskID.

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,780 @@
...
+const (
+	under int = iota
+	over
+)
+
+// defaults for the LoadBasedScheduler algorithm: only one class and all jobs map to that class
+var DefaultLoadBasedSchedulerClassPercents = map[string]int32{"c0": 100}
+var DefaultRequestorToClassMap = map[string]string{".*": "c0"}
+var DefaultMinRebalanceTime = time.Duration(4 * time.Minute)
+var MaxTaskDuration = time.Duration(4 * time.Hour)
```
anthonyjpratti: Prefer var blocks

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,780 @@
...
+
+type LoadBasedAlgConfig struct {
+	classLoadPercents           map[string]int
+	classLoadPercentsMu         sync.RWMutex
```
anthonyjpratti: Please format.

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,780 @@
...
+type LoadBasedAlg struct {
+	config     *LoadBasedAlgConfig
+	jobClasses map[string]*jobClass
+
+	totalUnusedEntitlement          int
+	exceededRebalanceThresholdStart time.Time
+
+	// local copy of load pcts and requestor map to use during assignment computation
+	// to insulate the computation from external changes to the configuration
+	classLoadPercents         map[string]int
```
anthonyjpratti: Please format.

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,780 @@
...
+	rebalanceThresholdMu   sync.RWMutex
+	rebalanceMinDuration   time.Duration
+	rebalanceMinDurationMu sync.RWMutex
+
+	classByDescLoadPct []string
+
+	stat stats.StatsReceiver
+}
+
+// LoadBasedAlg the scheduling algorithm allocates job tasks to workers using a class map as
```
anthonyjpratti: In the README, you say that workers are allocated to tasks and it is flipped here. So which is it?

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,780 @@
...
+// follows:
+// - each job maps to a 'class' (based on the job's requestor value)
+// - classes are assigned a % of the number of scoot workers.
+// When the algorithm is assigning tasks to workers it will try to start tasks such that the
```
anthonyjpratti: super nit: missing comma after introductory clause.

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,780 @@
...
+	lbs.classByDescLoadPct = lbs.getClassByDescLoadPct()
+
+	numWorkers := len(cs.nodes)
+	lbs.initOrigNumTargetedWorkers(numWorkers)
+
+	lbs.initJobClassesMap(jobsByRequestor)
+
+	rebalanced := false
+	var stopTasks []*taskState
+	if lbs.getRebalanceMinDuration() > 0 && lbs.getRebalanceThreshold() > 0 {
```
anthonyjpratti: Why not call getCurrentPercentsSpread here? If it's a perf concern, could you please comment that here.

scheduler/server/orig_scheduling_alg.go
```diff
@@ -7,10 +7,42 @@ import (
 	"github.com/twitter/scoot/scheduler/domain"
 )
 
-type OrigSchedulingAlg struct{}
+// Decrease the NodeScaleFactor by taking the percentage defined by NodeScaleAdjust[Priority].
+// Note: the use case here is to hit an SLA for each job priority, and this is a coarse way to do so.
```
anthonyjpratti: We have SLAs for scoot?

scheduler/server/orig_scheduling_alg.go
```diff
@@ -7,10 +7,42 @@ import (
 	"github.com/twitter/scoot/scheduler/domain"
 )
 
-type OrigSchedulingAlg struct{}
```
anthonyjpratti: I'm surprised this files contents changed. Shouldn't it stay the same mostly if the impl hasn't changed?

scheduler/server/orig_scheduling_alg.go
```diff
@@ -7,10 +7,42 @@ import (
...
-	requestors map[string][]*jobState, cfg SchedulerConfig) []*taskState {
+	requestors map[string][]*jobState) (startTasks []*taskState, stopTasks []*taskState) {
+
+	// Exit if there are no unscheduled tasks.
+	totalOutstandingTasks := 0
+	for _, j := range jobs {
+		totalOutstandingTasks += (len(j.Tasks) - j.TasksCompleted)
+	}
+
+	// Udate SoftMaxSchedulableTasks based on number of healthy nodes and the total number of tasks.
```
anthonyjpratti: nit: typo Udate to Update

scheduler/server/orig_scheduling_alg_test.go
```diff
@@ -23,9 +308,9 @@ func Test_StatefulSchedulerRequestorCounts(t *testing.T) {
 		{"numTasks": "1", "requestor": "p2Requestor", "priority": "2"},
 	}
 	jobIds := addRequestorTestJobsToScheduler(phase1Profiles, s)
-	s.step()                          // start the first set of tasks
-	time.Sleep(50 * time.Millisecond) // let task threads run to complete
-	s.step()                          // get first set of completed tasks and start next set
+	s.step()                           // start the first set of tasks
+	time.Sleep(500 * time.Millisecond) // let task threads run to complete
```
anthonyjpratti: This seems like a pretty large increase. Do you know why that is needed?

scheduler/server/job_state.go
anthonyjpratti: We deleted the priority algo below, so why do we have these fields still?

scheduler/server/priority_based_scheduling_alg.go
```diff
@@ -1,255 +0,0 @@
-package server
```
anthonyjpratti, line 1: If this is deleted, why do we still have comments that reference the priority based algo?

scheduler/server/stateful_scheduler.go
```diff
@@ -267,12 +259,22 @@ func NewStatefulScheduler(
 	if config.MaxJobsPerRequestor == 0 {
 		config.MaxJobsPerRequestor = DefaultMaxJobsPerRequestor
 	}
-	if config.SoftMaxSchedulableTasks == 0 {
-		config.SoftMaxSchedulableTasks = DefaultSoftMaxSchedulableTasks
+	config.SchedAlgConfig = &LoadBasedAlgConfig{
+		rebalanceThreshold:   0,
+		rebalanceMinDuration: 0 * time.Minute,
```
anthonyjpratti: Zero values here are already both zero, so this is not needed.

scheduler/server/job_state.go
JeanetteBruno: they are used by the new scheduling algorithm - I'll update the comment.

perftests/scheduler_simulator/test_alg.go
JeanetteBruno: I don't think it matters much... time. Sleep(d) will let the current goroutine enter sleeping sub-state, but still stay in running state, whereas, the channel receive operation <-time. After(d) will let the current goroutine enter blocking state.... I'm thinking that putting the test in blocking state might help with switching between threads.

perftests/scheduler_simulator/test_alg.go
JeanetteBruno: The intent of the condition is to exclude printing info on classes that don't have running or waiting tasks.

scheduler/client/cli/sched_alg_params.go
JeanetteBruno: to allow us to query the scheduler for the current load based algorithm parameters.

scheduler/api/server.go
JeanetteBruno: changed all RebalanceMinDuration to RebalanceMinimumDuration

scheduler/client/cli/cli.go
JeanetteBruno: changed

scheduler/client/cli/sched_alg_params.go
JeanetteBruno: lbsSchedAlgParams is the struct that holds all the algorithm params.  It carries setter function, but is also used by the getter.  I could have created a setLBSAlgoParams that is strictly a pass through to the lbsSchedAlgParam struct to isolate the setter function but thought that approach was more complicated than necessary.

scheduler/client/client.go
JeanetteBruno: reverted

scheduler/client/client.go
JeanetteBruno: good question - added it to the new functions

scheduler/client/scootcl/main.go
JeanetteBruno: done

scheduler/server/job_state.go
JeanetteBruno: fixed comment - these fields are being used

scheduler/server/job_state.go
JeanetteBruno: data structure change eliminated the method

scheduler/server/job_state.go
JeanetteBruno: done

scheduler/server/job_state.go
JeanetteBruno: added logging an error

scheduler/server/job_state.go
JeanetteBruno: done

scheduler/server/load_based_sched_alg.go
JeanetteBruno: done

scheduler/server/orig_scheduling_alg.go
JeanetteBruno: removed file

scheduler/server/orig_scheduling_alg_test.go
JeanetteBruno: removed file

scheduler/server/priority_based_scheduling_alg.go
JeanetteBruno: fixed comments

scheduler/server/stateful_scheduler.go
JeanetteBruno: removed

scheduler/server/task_scheduler.go
```diff
@@ -20,101 +20,108 @@ type taskAssignment struct {
...
 
-	// Udate SoftMaxSchedulableTasks based on number of healthy nodes and the total number of tasks.
-	// Setting the max to num healthy nodes means that each job can be fully scheduled.
-	// (This gets used in config.GetNodeScaleFactor())
-	cfg := SchedulerConfig{SoftMaxSchedulableTasks: max(totalOutstandingTasks, len(cs.nodes))}
-	if config != nil {
-		cfg = *config
+	tasks, stopTasks := s.config.SchedAlg.GetTasksToBeAssigned(jobs, stat, cs, requestors)
+	// Exit if no tasks qualify to be scheduled.
+	if len(tasks) == 0 {
```
JeanetteBruno, line 43: if len(stopTasks) > 0 then log an error - this should never happen

also verify that task is aborted (blocking) before moving on to starting tasks.

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,780 @@
...
+// if we see a job whose class % not defined, assign the job to the class with the
+// least number of workers
+func (lbs *LoadBasedAlg) initJobClassesMap(jobsByRequestor map[string][]*jobState) {
+	classNameWithLeastWorkers := lbs.classByDescLoadPct[len(lbs.classByDescLoadPct)-1]
+	// fill the jobClasses map with the state of the running jobs
+	for requestor, jobs := range jobsByRequestor {
+		var jc *jobClass
+		var ok bool
+		className := GetRequestorClass(requestor, lbs.requestorReToClassMap)
+		if className != "" {
```
JeanetteBruno: don't need to test for className != "" , !ok should be sufficient to know to use SchedLBSUnknownJobCounter and classNameWithLeastWorkers

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,780 @@
...
+		jc := lbs.jobClasses[className]
+		currPct := int(math.Floor(float64(jc.origNumRunningTasks) * 100.0 / float64(totalWorkers)))
+		pctDiff := jc.origTargetLoadPct - currPct
+		if pctDiff < 0 || jc.numWaitingTasks > 0 { // only consider classes using loaned workers, or with waiting tasks
+			minPct = min(minPct, pctDiff)
+			maxPct = max(maxPct, pctDiff)
+		}
+	}
+
+	return maxPct - minPct
```
JeanetteBruno, line 674: should this be subtract or add (minPct will be < 0 for loaned workers) - check the spreadsheet calc.

perftests/scheduler_simulator/fake_worker_cli.go
```diff
@@ -76,7 +76,9 @@ func (fw *FakeWorker) Run(cmd *runner.Command) (runner.RunStatus, error) {
 
 	fw.state = runner.RUNNING
 	go func(fw *FakeWorker) {
-		time.Sleep(time.Duration(duration) * time.Second)
+		t := time.NewTicker(time.Duration(duration) * time.Second)
```
dgdonut, line 79: <-time.After(...) would do the same thing

scheduler/server/job_state.go
```diff
@@ -151,13 +171,18 @@ func (j *jobState) taskStarted(taskId string, tr *taskRunner) {
 		delete(j.Completed, taskId)
 	}
 	j.Running[taskId] = taskState
+
+	// add the task to the map of tasks by start time
+	startTimeSec := taskState.TimeStarted.Truncate(time.Second)
```
dgdonut, line 176: We are always truncating to the nearest time.Second but this is not enforced anywhere (e.g. by some sort of wrapping mechanism). If we ever change or add code and forget to do this it will probably break things.

scheduler/domain/definitions.go
```diff
@@ -52,6 +52,10 @@ type JobDefinition struct {
 	Tasks     []TaskDefinition
 }
 
+func (jd *JobDefinition) String() string {
+	return fmt.Sprintf("jobType:%s, req:%s, tag:%s, basis: %s, tasks:%d", jd.JobType, jd.Requestor, jd.Tag, jd.Basis, len(jd.Tasks))
+}
+
```
dgdonut, line 58: Somewhat of a tangent - what is happening with the Priority constructs defined in this file and used in various places?

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,781 @@
...
+)
+
+const (
+	under int = iota
+	over
+)
+
+// defaults for the LoadBasedScheduler algorithm: only one class and all jobs map to that class
+var (
+	DefaultLoadBasedSchedulerClassPercents = map[string]int32{"c0": 100}
```
dgdonut: Realizing now, we are making the scheduler more stateful. When schedulers restart they will lose any configured class %s. We will need a mechanism to preserve this between instances.

scheduler/server/stateful_scheduler.go
```diff
@@ -1162,6 +1146,29 @@ func (s *statefulScheduler) processKillJobRequests(reqs []jobKillRequest) {
...
+func (s *statefulScheduler) abortTask(jobState *jobState, task *taskState, logFields log.Fields) []saga.SagaMessage {
+	logFields["taskID"] = task.TaskId
+	log.WithFields(logFields).Info("aborting task")
+	var updateMessages []saga.SagaMessage
+
+	if task.Status == domain.InProgress {
+		task.TaskRunner.Abort(true, UserRequestedErrStr)
+	} else if task.Status == domain.NotStarted {
+		st := runner.AbortStatus("", tags.LogTags{JobID: jobState.Job.Id, TaskID: task.TaskId})
+		st.Error = UserRequestedErrStr
```
dgdonut, line 1159: It's not UserRequested

scheduler/server/stateful_scheduler.go
```diff
@@ -1162,6 +1146,29 @@ func (s *statefulScheduler) processKillJobRequests(reqs []jobKillRequest) {
 	}
 }
 
+// abortTask abort a task - will be triggered by killing a job and when the scheduling algorithm rebalances the workers
```
dgdonut, line 1149: Is this substantially different than the existing abort task mechanisms? Why did we need it?

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,781 @@
...
+	// local copy of load pcts and requestor map to use during assignment computation
+	// to insulate the computation from external changes to the configuration
+	classLoadPercents     map[string]int
+	requestorReToClassMap map[string]string
+
+	classByDescLoadPct             []string
+	tasksByJobClassAndStartTimeSec map[taskClassAndStartKey]taskStateByJobIDTaskID
+}
+
+// NewLoadBasedAlg allocate a new LoadBaseSchedAlg object.  If the load %'s don't add up to 100
```
dgdonut: Doesn't seem to return errors or care what the %s add up to

dgdonut (commented):
How is default class bucketing going to work, I don't see this mentioned anywhere any more.

perftests/scheduler_simulator/fake_worker_cli.go
JeanetteBruno: I know - I happened to read that  using ticker/stop is more efficient wrt to garbage collection when I was looking at this part of the code so converted it. I expect we'll increase the simulation to use 10K workers managing 100k tasks as some point, so converted it.

scheduler/domain/definitions.go
JeanetteBruno: It's not used anymore, but I'm not taking it out due to the extent of the changes.  
I've created https://jira.twitter.biz/browse/MGCI-1310 for clean up issues.

scheduler/server/job_state.go
JeanetteBruno: it's only local to this function - the whole intent of this function is to step backwards in time 1 second at a time.  Only the initializer (ln 572) and time decrement (ln 584)  should be modifying the time value.  If we're doing something else to modify the time value, yes we would need to compute the time then 'wrap' with Truncate() (In essence Truncate() is the only wrapper we could have in common to all lines of code impacting the time index.)

scheduler/server/job_state.go
JeanetteBruno: thanks for catching that.  I added it last year when I was starting to work with the (earlier) new algorithm and wasn't really sure if we needed to trap for error or what..... since then I realized we don't need it.

scheduler/server/stateful_scheduler.go
JeanetteBruno: no - it's copying the abort functionality from inline in processKillJobRequests() and exposing it so it can be used by both processKillJobRequests(), and the taskScheduler.go getTaskAssignments when stopping tasks for rebalancing.

scheduler/server/stateful_scheduler.go
JeanetteBruno: thanks! made into parameter.  Rebalance aborts will use "RebalanceRequested"

scheduler/server/task_scheduler.go
JeanetteBruno: abort is blocking, added error message

scheduler/server/load_based_sched_alg.go
JeanetteBruno: looked at it some more - think its ok (for now - till we have a better handle on thresholds, etc)

scheduler/server/load_based_sched_alg.go
JeanetteBruno: it normalizes the %s automatically

scheduler/server/load_based_sched_alg.go
JeanetteBruno: we need to talk about this, I think there are other parameters (throttle?) we might also want to preserve... think for now we should set meaningful default values (after running simulations to understand what they should be).  Then have another ticket for preserving parameter values across restarts.

perftests/scheduler_simulator/fake_worker_cli.go
ryancouto: could also just skip assigning to t
`<-time.NewTicker(time.Second).C`

scheduler/client/cli/sched_alg_params.go
ryancouto: I think I'd prefer that approach - isolating the setter

scheduler/client/cli/sched_alg_params.go
```diff
@@ -0,0 +1,174 @@
...
+		Short: "SetSchedAlgParams",
+	}
+	r.Flags().StringVar(&s.clpFilePath, "class_pcts_file", "", "JSON file to read class load percents from.")
+	r.Flags().StringVar(&s.reqMapFilePath, "requestor_map_file", "", "JSON file to read requestor to class map from.")
+	r.Flags().IntVar(&s.RebalanceMinimumDuration, "rebalance_min", -1, "The number of minutes the tasks must be over rebalance threshold to trigger rebalance. 0 implies no rebalance. (Default of -1 implies no entry)")
+	r.Flags().IntVar(&s.RebalanceThreshold, "rebalance_threshold", -1, "The rebalance threshold. 0 implies no rebalance. (Default of -1 implies no entry)")
+	return r
+}
+
+func (s *lbsSchedAlgParams) run(cl *simpleCLIClient, cmd *cobra.Command, args []string) error {
```
ryancouto: I understand the desire to minimize the # of files, but I personally would prefer to have separate files for each command.

scheduler/client/client.go
ryancouto: Going further - why don't we just reset the connection regardless? Is there any value in persisting the client? Could be another cleanup item to add to MGCI-1310

scheduler/server/job_state.go
```diff
@@ -21,11 +24,14 @@ type jobState struct {
...
-	// track tasks by state (completed, running, not started) for priority based scheduling algorithm
-	// TODO remove if we don't use the priority based scheduling algorithm
-	Completed  map[string]*taskState
-	Running    map[string]*taskState
-	NotStarted map[string]*taskState
+	// track tasks by state (completed, running, not started) for scheduling algorithm
+	// Completed and Running are only used by the scheduling algorithm
+	Completed  taskStateByTaskID
+	Running    taskStateByTaskID
+	NotStarted taskStateByTaskID
```
ryancouto, line 31: I think we should consider renaming NotStarted to Waiting. We use both interchangeably, but Waiting seems like a better descriptor imo.

scheduler/server/job_state.go
```diff
@@ -42,6 +48,17 @@ type taskState struct {
 
 type taskStatesByDuration []*taskState
 
+// the following types are used to access task State objects by class, startTime, taskID
+type taskClassAndStartKey struct {
```
ryancouto, line 52: don't think we need to include "key" in this name. the fact that it's used as a key in maps doesn't tell me anything about the struct itself.

scheduler/server/job_state.go
```diff
@@ -42,6 +48,17 @@ type taskState struct {
...
+	class string
+	start time.Time
+}
+type jobIDTaskIDKey struct {
```
ryancouto, line 56: ditto re: "key"

scheduler/server/job_state.go
```diff
@@ -197,3 +228,39 @@ func (j *jobState) getJobStatus() domain.Status {
...
 	return domain.InProgress
 }
+
+// addTaskToStartTimeMap add the running task to the map that bins running tasks by their class and start time
+func (j *jobState) addTaskToStartTimeMap(jobClass string, task *taskState, startTimeSec time.Time) {
+	if j.tasksByJobClassAndStartTimeSec == nil {
+		log.Errorf("tasksByJobClassAndStartTime map not found.  Skipping adding task to it. jobID: %s, taskID:%s, jobClass:%s, startTime:%s", task.JobId, task.TaskId, jobClass, startTimeSec.Format("2006-01-02 15:04:05 -0700 MST"))
+		return
+	}
+	timeBucket := taskClassAndStartKey{class: jobClass, start: startTimeSec}
```
ryancouto: it's a timeBucket, but it's also a classBucket, right?

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,781 @@
...
+
+	// jobsByNumRunningTasks is a map that bins jobs by their number of running tasks.  Given that the algorithm has
+	// determined it will start n tasks from class A, the tasks selected for starting from class A will give prefence
+	// to jobs with the least number of running tasks.
+	jobsByNumRunningTasks map[int][]jobWaitingTaskIds
+	// the largest key value in the jobsByNumRunningTasks map
+	maxTaskRunningMapIndex int
+
+	origNumWaitingTasks int
+	origNumRunningTasks int
```
ryancouto: let's add comments for these fields too

scheduler/server/task_scheduler.go
```diff
@@ -13,108 +13,121 @@ type taskAssignment struct {
...
-func getTaskAssignments(cs *clusterState, jobs []*jobState,
-	requestors map[string][]*jobState, config *SchedulerConfig, stat stats.StatsReceiver,
-	schedAlg SchedulingAlgorithm) (
-	[]taskAssignment, map[string]*nodeGroup,
-) {
+func (s *statefulScheduler) getTaskAssignments() []taskAssignment {
+	cs := s.clusterState
+	jobs := s.inProgressJobs
+	requestors := s.requestorMap
+	stat := s.stat
```
ryancouto: why make these assignments?

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,781 @@
...
+// Jobs within a class are binned by the number of running tasks (ranking jobs by the number of active tasks).
+// When starting tasks for a class, the tasks are first pulled from jobs with the least number of active tasks.
+func (lbs *LoadBasedAlg) GetTasksToBeAssigned(jobsNotUsed []*jobState, stat stats.StatsReceiver, cs *clusterState,
+	jobsByRequestor map[string][]*jobState) ([]*taskState, []*taskState) {
+	log.Debugf("in LoadBasedAlg.GetTasksToBeAssigned: numWorkers:%d, numIdleWorkers:%d", len(cs.nodes), cs.numFree())
+
+	// make local copies of the load pct structures
+	lbs.classLoadPercents = lbs.LocalCopyClassLoadPercents()
+	lbs.requestorReToClassMap = lbs.getRequestorToClassMap()
+	lbs.classByDescLoadPct = lbs.getClassByDescLoadPct()
```
ryancouto, line 183: why do we assign these here if we already have access to it via the config?

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,781 @@
...
+
+	numWorkers := len(cs.nodes)
+	lbs.initOrigNumTargetedWorkers(numWorkers)
+
+	lbs.initJobClassesMap(jobsByRequestor)
+
+	rebalanced := false
+	var stopTasks []*taskState
+	if lbs.getRebalanceMinimumDuration() > 0 && lbs.getRebalanceThreshold() > 0 {
+		// currentPctSpread is the delta between the highest and lowest
```
ryancouto: between highest and lowest what?

JeanetteBruno:
> How is default class bucketing going to work, I don't see this mentioned anywhere any more.
not sure if your asking about unkown requestors or default class percents and requestor to class maps so I'll answer both....

**default class %s and requestor map (default values on scheduler restart):**
stateful_scheduler.go creates the LoadBasedAlg giving it DefaultLoadBasedSchedulerClassPercents, and DefaultRequestorToClassMap.  We'll initialize those 2 constants with the values determined by our simulation runs.

**unknown requestors:**
the default class %s will include an unknown requestor class that is allocated 1% of the workers.  jobs with unknown requestors will be assigned to this class.

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,781 @@
...
+		currentPctSpread := lbs.getCurrentPercentsSpread(numWorkers)
+		if currentPctSpread > lbs.getRebalanceThreshold() {
+			nilTime := time.Time{}
+			if lbs.exceededRebalanceThresholdStart == nilTime {
```
ryancouto, line 197: when would this be true? just on the first iteration?

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,781 @@
...
+		stat.Gauge(fmt.Sprintf("%s_%s", stats.SchedJobClassTasksStarting, jc.className)).Update(int64(jc.numTasksToStart))
+		stat.Gauge(fmt.Sprintf("%s_%s", stats.SchedJobClassTasksWaiting, jc.className)).Update(int64(jc.origNumWaitingTasks - jc.numTasksToStart))
+		stat.Gauge(fmt.Sprintf("%s_%s", stats.SchedJobClassTasksRunning, jc.className)).Update(int64(jc.origNumRunningTasks))
+		stat.Gauge(fmt.Sprintf("%s_%s", stats.SchedJobClassDefinedPct, jc.className)).Update(int64(jc.origTargetLoadPct))
+		finalPct := int(math.Round(float64(jc.origNumRunningTasks+jc.numTasksToStart) / float64(int64(numWorkers)*100.0)))
+		stat.Gauge(fmt.Sprintf("%s_%s", stats.SchedJobClassActualPct, jc.className)).Update(int64(finalPct))
+	}
+
+	log.Debugf("Returning %d start tasks, %d stop tasks", len(tasksToStart), len(stopTasks))
+	return tasksToStart, stopTasks
```
ryancouto: let's stay consistent with naming - tasksToStart & tasksToStop or startTasks & stopTasks

perftests/scheduler_simulator/fake_worker_cli.go
anthonyjpratti: Is there a reason to prefer <-time.After over Sleep?

scheduler/client/cli/sched_alg_params.go
anthonyjpratti: I'd follow what is convention in the other files, which appears to be a "set" struct and a "get" struct

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,781 @@
...
+// getTasksToStartForJobClass get the tasks to start list for a given jobClass.  The jobClass's numTasksToStart
+// field will contain the number of tasks to start for this job class.  The jobClass's jobsByNumRunningTasks is
+// a map from an integer value (number of tasks running) to the list of jobs with that number of tasks running
+// For a given jobClass, we start adding tasks from the jobs with the least number of tasks running.
+// (Note: when a task is started for a job, the job is moved to the ‘next’ bin and placed at the end of that bin’s job list.)
+func (lbs *LoadBasedAlg) getTasksToStartForJobClass(jc *jobClass) []*taskState {
+	tasks := []*taskState{}
+
+	startingTaskCnt := 0
+	// work our way through the class's jobs, starting with jobs with the least number of running tasks,
```
ryancouto, line 523: why least running tasks vs lowest pct of running tasks relative to waiting?

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,781 @@
...
+		}
+		// jobs contains list of jobs and their waiting taskIds. (Each job in this list has the same number of running tasks.)
+		// Allocate one task from each job till we've allocated numTasksToStart for the jobClass, or have allocated 1 task from
+		// each job in this list.  As we allocate a task for a job, move the job to the end of jc.jobsByNumRunningTasks[numRunningTasks+1].
+		for _, job := range jobs {
+			if job.waitingTaskIDs != nil && len(job.waitingTaskIDs) > 0 {
+				// get the next task to start from the job
+				tasks = append(tasks, lbs.getJobsNextTask(job))
+
+				// move the job to jobsByRunning tasks with numRunningTasks + 1 entry.  Note: we don't have to pull it from
```
ryancouto: *jobsByRunningTasks

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,781 @@
...
+				startingTaskCnt++
+				if startingTaskCnt == jc.numTasksToStart {
+					return tasks
+				}
+			}
+
+		}
+	}
+
+	return tasks // note: we should never hit this line
```
ryancouto: why not? why include it?

scheduler/server/load_based_sched_alg.go
```diff
@@ -0,0 +1,781 @@
...
+	lbs.config.classByDescLoadPct = keys
+
+	// set the load pcts - normalizing them if the don't add up to 100
+	lbs.config.classLoadPercents = map[string]int{}
+	pctTotal := 0
+	for _, val := range classLoadPercents {
+		pctTotal += int(val)
+	}
+	if pctTotal != 100 {
+		log.Errorf("LoadBalanced scheduling %%'s don't add up to 100, normalizing them")
```
ryancouto, line 733: let's break normalization out into a helper so we can more easily test it

perftests/scheduler_simulator/fake_worker_cli.go
JeanetteBruno: need access to t so we can stop the ticker.

scheduler/client/cli/sched_alg_params.go
JeanetteBruno: I'm going to push back on this again.... my reasoning

1. this file makes 11 files in the directory (excluding doc.go) which moves us into 'create subdirectories range'
2. both cli subcommands use the lbsSchedAlgParams struct, so the getter would be using a data structure defined in the setter's go file adding a little more clumsiness to the code
3. it seem's very natural to me to have setter/getter functionality defined together.

but adding to mgci-1310 - who ever picks it up can make the change if they wish

scheduler/client/cli/sched_alg_params.go
JeanetteBruno: see comment above

scheduler/client/client.go
JeanetteBruno: added to ticket

scheduler/server/job_state.go
JeanetteBruno: added to mgci-1310

scheduler/server/job_state.go
JeanetteBruno: I have 'Key' in it as an indicator that it is only used (useful) for access the map (allowing us to find tasks by class and start time)

scheduler/server/job_state.go
JeanetteBruno: see comment above

scheduler/server/job_state.go
JeanetteBruno: actually it's the key for a set of tasks bucketed by class/time... changed

scheduler/server/task_scheduler.go
JeanetteBruno: was trying to minimize code changes from moving the function into stateful scheduler, but then ended up having to change the whole function anyway, so will remove.
Thanks for drawing my attention to it.

scheduler/server/load_based_sched_alg.go
dgdonut: I mean - the function is not doing what the comment is saying

JeanetteBruno:
wrt to Ryan's question: _when would this be true? just on the first iteration?_ (when is currentPctSpread > lbs.getRebalanceThreshold()) - this is true when a lot of long running tasks have been started on loaned workers. (see algorithm summary above LoadBasedAlg struct definition and in README.md)

wrt to Ryan's question: _why least running tasks vs lowest pct of running tasks relative to waiting?_ we could choose to change the algorithm to that, imo it is more fair to try to give all jobs within a class an equal number of workers.  If we go with lowest pct of running to waiting tasks we could end up with jobs with lots of tasks starving jobs with small number of tasks.

wrt to Ryan's question: _why not? why include it?_. I added a log statement and updated the comment... the algorithm should always return in the loop above, but needed to include a return statement outside of the loop for the code to compile.

scheduler/server/load_based_sched_alg.go
JeanetteBruno: rewrote comments/README

scheduler/server/load_based_sched_alg.go
JeanetteBruno: rewrote the whole section.

scheduler/server/load_based_sched_alg.go
JeanetteBruno: this is true when a lot of long running tasks have been started on 'loaned' workers.

scheduler/server/load_based_sched_alg.go
JeanetteBruno: lowest pct of running tasks would allow jobs with many tasks to starve jobs with few tasks.  I think trying to have an equal number of tasks running for each job in a class is a better solution.

scheduler/server/load_based_sched_alg.go
JeanetteBruno: done

dgdonut (approved):

scheduler/client/cli/sched_alg_params.go
anthonyjpratti: Please follow the current conventions, it makes the code harder to read and reason about when deviations happen.

perftests/scheduler_simulator/fake_worker_cli.go
JeanetteBruno: > Is there a reason to prefer <-time.After over Sleep?

see comment above- ticker is more efficient wrt to garbage collection.

willnorris:
closing PR (and will immediately reopen) to try and bump CLA Assistant.

PR #491, opened 2020-12-02
Migrate from travis to GitHub actions

anthonyjpratti:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,42 @@
+on: [push, pull_request]
+name: ci
+env:
+  GO111MODULE: on
+
+jobs:
+  test:
```
JeanetteBruno: perhaps name validate or validate-change (it does more than test)

also - do you know if we need to consider billing? (https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions and https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration)

JeanetteBruno (approved):
see comment about changing 'test' job name.

.github/workflows/ci.yml
anthonyjpratti: Sure I can change it to validate. I think the free tier should be more the enough, since it includes 2,000 minutes for free.

dgdonut (approved):
LGTM

PR #495, opened 2021-01-14
fixed panic error when adding a worker to "" groups idle list

JeanetteBruno:
scheduler update introduced a panic at ln 273 in scheduler/server/cluster_state.go
**panic root cause:**
The panic was triggered by the following scenario:
-  all workers were allocated to non-"" groupIDs (snapshotID)
- task_scheduler would delete the "" group in cluster_state since there were no workers in that group
- then a worker was lost and re-added to the cluster before cluster_state recognized the worker as suspended
- cluster_state would try to add the worker to the "" group assuming the "" group existed. (ln 273 cluster_state.go)

**summary of the fix**
- The change updates task_scheduler to use a local copy of the node groups and does not update cluster_state's nodeGroups
- The old code also updated cluster_states's nodeGroups directly in the assign() function, then tried to do the same update in cluster_state.taskScheduled() function.  (the second update would be a nop.)
- the unit test was updated to verify that the "" node group exists in cluster state after task assignment

**validation**
I recreated the panic in my devel instance then verified that I could not recreate the panic with this update deployed.

### DIFF
--- a/scheduler/server/task_scheduler.go
+++ b/scheduler/server/task_scheduler.go
@@ -5,6 +5,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 
+	"github.com/twitter/scoot/cloud/cluster"
 	"github.com/twitter/scoot/common/stats"
 )
 
@@ -70,7 +71,7 @@ func (s *statefulScheduler) getTaskAssignments() []taskAssignment {
 	// - Hot node for the given snapshotId (one whose last task shared the same snapshotId).
 	// - New untouched node (or node whose last task used an empty snapshotId)
 	// - A random free node from the idle pools of nodes associated with other snapshotIds.
-	assignments := assign(s.clusterState, tasks, s.stat)
+	assignments := s.assign(tasks)
 	log.WithFields(
 		log.Fields{
 			"numAssignments": len(assignments),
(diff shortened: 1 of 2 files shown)


scheduler/server/task_scheduler.go
```diff
@@ -164,16 +153,20 @@ func assign(
...
-		stat.Counter(stats.SchedScheduledTasksCounter).Inc(1)
+		s.stat.Counter(stats.SchedScheduledTasksCounter).Inc(1)
 	}
 	return assignments
 }
 
-func findIdleNodeInGroup(cs *clusterState, groupID string) *nodeState {
-	for _, ns := range cs.nodeGroups[groupID].idle {
-		if ns.suspended() || cs.isOfflined(ns) {
+// findIdleNodeInGroup find a node in the group's idle nodes that is not suspended or offlined (this method will, pick
```
ryancouto: this isn't entirely accurate - it's looking through both the idle and busy fields of the nodeGroup

scheduler/server/task_scheduler.go
```diff
@@ -81,48 +82,54 @@ func (s *statefulScheduler) getTaskAssignments() []taskAssignment {
...
-) (assignments []taskAssignment) {
-	idleNodeGroupIDs := map[string]string{}
-	for groupID, group := range cs.nodeGroups {
+// Note: there may be a race condition between recognizing idle nodes as available for assignment and
+// nodes becoming offlined or suspended. The code does the best it can, but it may assign a task to
+// a node that clusterState considers offlined/suspended before or as the task is actually being started
+func (s *statefulScheduler) assign(tasks []*taskState) (assignments []taskAssignment) {
+	idleNodesByGroupIDs := map[string]nodeStatesByNodeID{}
+
+	// make a local copy of (non-suspended/offlined) idle nodeIDs by groupID, these will be the nodes assigned to the tasks
```
ryancouto: why do we need a local copy? can we include that reasoning in this comment?

scheduler/server/task_scheduler.go
```diff
@@ -81,48 +82,54 @@ func (s *statefulScheduler) getTaskAssignments() []taskAssignment {
...
-		} else {
+		if nodeSt == nil {
 			// could not find any free nodes in node group for the task's snapshot id.  Look for a free node in the other
 			// node groups, starting with the "" node group
-			if _, ok := idleNodeGroupIDs[""]; ok {
-				nodeSt = findIdleNodeInGroup(cs, "")
+			if nodeGroup, ok := idleNodesByGroupIDs[""]; ok {
+				nodeSt = s.findIdleNodeInGroup(nodeGroup)
 			}
 			if nodeSt == nil {
```
ryancouto, line 128: we're already in an if block for if nodeSt == nil here, we don't need to check again

JeanetteBruno: ln 118 may set nodeSt.

scheduler/server/task_scheduler.go
JeanetteBruno: nodeGroup is the local copy of clusterState's nodeGroups - this local copy only contains idle nodes 

this local node structure confused me in the first place - hence the bug!  I'll try updating comments, perhaps that will help.  Also recommendations on better variable names are welcome.

scheduler/server/task_scheduler.go
JeanetteBruno: good idea - this tripped me up for a while.... updated the comment

scheduler/server/task_scheduler.go
ryancouto: whoops!

dgdonut (approved):
Makes sense to me, thanks for detailed description and comments

PR #496, opened 2021-01-20
add framework for persisting scheduler settings with nop persistence …

JeanetteBruno:
This change adds a setting persistor to scheduler.
If no persistor is provided, a nop persistor is used.


scheduler/scheduler/main.go
```diff
@@ -76,6 +77,10 @@ func main() {
 		func() (*temp.TempDir, error) {
 			return temp.NewTempDir("", "sched")
 		},
+
+		func() server.Persistor {
+			return nil
```
ryancouto, line 82: why not create a NewNoopPersistor and inject it here?

scheduler/server/stateful_scheduler.go
```diff
@@ -292,7 +297,15 @@ func NewStatefulScheduler(
...
 		tasksByJobClassAndStartTimeSec: tasksByClassAndStartMap,
+		persistor:                      persistor,
+	}
+
+	sched.setThrottle(-1)
+
+	if sched.persistor == nil {
+		sched.persistor = Persistor(&nopPersistor{})
 	}
+	sched.persistor.LoadSettings(sched)
```
ryancouto: does this override the setThrottle above?

scheduler/server/stateful_scheduler.go
```diff
@@ -292,7 +297,15 @@ func NewStatefulScheduler(
 		stat:             stat,
 
 		tasksByJobClassAndStartTimeSec: tasksByClassAndStartMap,
+		persistor:                      persistor,
+	}
+
+	sched.setThrottle(-1)
+
+	if sched.persistor == nil {
```
ryancouto, line 334: nit: seems weird to create this statefulScheduler and then check if one of the fields we assigned it is nil... could we check if persistor is nil above the statefulScheduler instantiation?

JeanetteBruno: actually, this is the 'style' I'm trying to get to... use nopPersistor as default behavior - allow so so consumers of statefulScheduler don't have to think/know about Persistor if they don't care about that feature.

scheduler/scheduler/main.go
JeanetteBruno: we could, but I would still like to have statefulScheduler provide a default persistor (NopPersistor) if persistor is nil.  (See my comment below.)

scheduler/server/stateful_scheduler.go
JeanetteBruno: It depends on the implementation of LoadSettings()...the nop persistor doesn't do anything so throttle is not changed.
The Persistor implemented in our twitter only code (source) behaves as follows:
- don't do anything if no setting have been saved
- if any setting is changed, all settings are persisted so LoadSetting will set the correct values.

scheduler/server/persist_settings.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// Persistor interface for persisting scheduler settings and initializing the scheduler
+// from its persisted settings
+type Persistor interface {
+	PersistSettings(s Scheduler)
```
dgdonut: I'm confused by the interface and how it gets used by the scheduler. Assuming it makes sense that a Scheduler contains an implementation of Persistor, why would these methods take a Scheduler as a parameter? Instinctively I would expect to call something like:
`PersistSettings(ps PersistedSettings) error`
`LoadSettings() (ps PersistedSettings, error)`

Seems like at minimum these should return an error type as well - need to be able to propagate errors on set/get operations.

scheduler/server/persist_settings.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (p *nopPersistor) PersistSettings(s Scheduler) {
+	return
+}
+
+func (p *nopPersistor) LoadSettings(s Scheduler) {
+	return
+}
+
+func (s *statefulScheduler) LoadPersistedSettings(settings PersistedSettings) {
```
dgdonut: If we have a nopPersistor do we even need this public API? What if we just always loaded persistent settings when the scheduler starts (with a default of using a nop unless a different one was provided)?

scheduler/server/persist_settings.go
JeanetteBruno: agree that the interface was confusing... I reorganized it - thx for calling that out.

scheduler/server/persist_settings.go
JeanetteBruno: reorged the code only the Persistor's methods are public, statefulScheduler's load/persist are not visible.

scheduler/server/load_based_sched_alg.go
```diff
@@ -733,6 +733,8 @@ func (lbs *LoadBasedAlg) setClassLoadPercents(classLoadPercents map[string]int32
 		log.Errorf("LoadBalanced scheduling %%'s don't add up to 100, normalizing them")
 		lbs.normalizePercents(pctTotal)
 		log.Errorf("LoadBalanced scheduling class percents have been changed to %v", lbs.config.classLoadPercents)
+	} else {
+		log.Infof("set classLoadPercents to %v", lbs.config.classLoadPercents)
```
dgdonut: nit, but prefer we log this info on one line before the "if pctTotal != 100" check, so we don't have to have an else block.

scheduler/server/persist_settings.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (p *nopPersistor) PersistSettings(settings *PersistedSettings) error {
+	return nil
+}
+
+func (p *nopPersistor) LoadSettings() (*PersistedSettings, error) {
+	return nil, nil
+}
+
+func (s *statefulScheduler) persistSettings() {
```
dgdonut: Return an error here and in loadSettings()?

scheduler/server/stateful_scheduler.go
```diff
@@ -1176,7 +1189,9 @@ func (s *statefulScheduler) SetSchedulerStatus(maxTasks int) error {
 	if err != nil {
 		return err
 	}
-	s.config.TaskThrottle = maxTasks
+	s.setThrottle(maxTasks)
```
dgdonut, line 1275: Got confused here again. These functions for GetSchedulerStatus and SetSchedulerStatus sound generic but are only used for the throttle setting. Can we rename or is that going to break APIs (manual only APIs may be OK to break).

scheduler/server/stateful_scheduler.go
```diff
@@ -1176,7 +1189,9 @@ func (s *statefulScheduler) SetSchedulerStatus(maxTasks int) error {
...
+	s.persistSettings()
```
dgdonut: For all setters where we call load or persistSettings - should we interpret failure to persist as an error? Should we not update local values if the persist call fails (ensure all or nothing updated state)?

scheduler/server/stateful_scheduler.go
JeanetteBruno: I that the name is not good... 
Sickle is currently using it to select the cluster.

IMO I should merge the alg params into this.... 
- add a --throttle option, but also keep the current behavior, then
- Update sickle to use the --throttle option, then
- remove the default throttle behavior.

This positions us to extend the api for other parameters.

dgdonut: So we'll end up with one pair of get/set settings APIs for all settings including throttle and algorithm? And we'll use flags for the CLI tools to make shortcuts? (SGTM)

scheduler/server/persist_settings.go
JeanetteBruno: adding it to persist, but not to loadSettings().  loadSettings() already prints a log message that default values were used, the call to loadSettings() would do nothing with the error response.

scheduler/server/stateful_scheduler.go
JeanetteBruno: this is going to be a larger change, will create ticket for it.

scheduler/server/stateful_scheduler.go
JeanetteBruno: I don't think so.... I've updated the code so persist failures return an error.  The local values will have been set, but lack of persist success just means that if the scheduler restarts it will not pick up the updated values.
so the ux is that we manually update the setting, and we should look at the response.  If we see a message that that the value was not persisted, we should retry the update again till we see that the value is persisted (or start debugging why the db write is failing).

scheduler/server/persist_settings.go
```diff
@@ -0,0 +1,88 @@
...
+	_, throttle := s.GetSchedulerStatus()
+	ps := &PersistedSettings{
+		ClassLoadPercents:               sa.getClassLoadPercents(),
+		RequestorToClassMap:             sa.getRequestorToClassMap(),
+		RebalanceMinimumDurationMinutes: int(sa.getRebalanceMinimumDuration().Minutes()),
+		RebalanceThreshold:              sa.getRebalanceThreshold(),
+		Throttle:                        throttle,
+	}
+
+	err := s.persistor.PersistSettings(ps)
```
dgdonut, line 56: can just do `return s.persistor.PersistSettings(ps)` unless the additional Errorf context is strictly necessary

dgdonut (approved):
General comment on "scheduler may use default..." error messages - scheduler will use the last saved values, not defaults

scheduler/server/load_based_sched_alg.go
JeanetteBruno: done

scheduler/server/stateful_scheduler.go
JeanetteBruno: see MGCI-1310

scheduler/server/persist_settings.go
JeanetteBruno: I prefer the additional error context - explaining the impact of the error on the scheduler functionality.

PR #497, opened 2021-02-16
update loan calculation to include current loan counts

JeanetteBruno:
The old worker loan calculation did not factor in a class's current loaned workers.   I saw an example of the following behavior:
diff and tryout were the only classes with running tasks and both were exceeding their allocations - using workers loaned from the other classes that didn't have any tasks running.  Since diff was set to 9% and tryout to 2% load, the normalized %'s were 82% for diff and 18% for tryout.  The diff job was running longer running tasks and over a lot of iterations accumulated many more 'loaned' workers than their targeted 82%.

This change factors adjusts the computed loan % factoring in the current number of loaned workers.

### DIFF
--- a/scheduler/server/load_based_sched_alg_test.go
+++ b/scheduler/server/load_based_sched_alg_test.go
@@ -16,23 +16,6 @@ import (
 	"github.com/twitter/scoot/scheduler/domain"
 )
 
-var (
-	default_sc = SchedulerConfig{
-		MaxRetriesPerTask:    0,
-		DebugMode:            false,
-		RecoverJobsOnStartup: false,
-		DefaultTaskTimeout:   0,
-		TaskTimeoutOverhead:  0,
-		RunnerRetryTimeout:   0,
-		RunnerRetryInterval:  0,
-		ReadyFnBackoff:       0,
-		MaxRequestors:        0,
-		MaxJobsPerRequestor:  0,
-		TaskThrottle:         0,
-		Admins:               nil,
-	}
-)
-
 type classState struct {
 	loadPct              int
 	numRunningTasks      int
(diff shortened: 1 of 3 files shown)


dgdonut:
test comment

dgdonut:
I think I get it... looks mostly OK but I admit I'm glossing over my review of some of the math code. In the summary example, what would be the desired %'ages after proper normalization?
Broken test needs to be fixed.

scheduler/server/load_based_sched_alg_test.go
```diff
@@ -47,8 +30,13 @@ type testDef struct {
 	classes      map[string]classState
 }
 
+// test the algorithm for various scenarios
+// https://docs.google.com/spreadsheets/d/1QP-okxrVkmKBMtc4OzWe8cFU15u84b9aXENeKqGogbc/edit#gid=554928957 is a
```
ryancouto, line 34: this can't be seen by people outside twitter's org

scheduler/server/load_based_sched_alg_test.go
```diff
@@ -113,6 +108,40 @@ func Test_Class_Task_Start_Cnts(t *testing.T) {
 			"c6": {loadPct: 3, numRunningTasks: 977, numWaitingTasks: 209145, numJobs: 30, expectedTasksToStart: 0},
 			"c7": {loadPct: 2, numRunningTasks: 2612, numWaitingTasks: 416781, numJobs: 40, expectedTasksToStart: 0}},
 		},
+		// scenario 9 - mimick just diff and tryout running on staging
+		{totalWorkers: 3270, classes: map[string]classState{
+			"land":       {loadPct: 48, numRunningTasks: 0, numWaitingTasks: 0, numJobs: 0, expectedTasksToStart: 0},
```
ryancouto, line 113: separate from this diff - but I'm realizing now that it seems strange to have all these twitter internal classes in open source. how difficult would it be to pull that out of this repo, use mock classes here, and set the real default percentages in twitter source?

dgdonut: I think we can dump this stuff when persistence works

scheduler/server/load_based_sched_alg_test.go
dgdonut: yes, we shouldn't put these links in open source.

scheduler/server/load_based_sched_alg_test.go
JeanetteBruno: I used these tests to validate the loan change scenario. 

We have the default values (land, diff, etc.) coded in load_based_sched_alg.go..... we could change those to a general c0 value that has 100%, and have our twitter code change it to the land, diff, etc values when scheduler starts up, but I thought that was more complicated than needed.  (since we are pretty much the only users of scoot)

dgdonut: I disagree with this, unless we scrap this rep altogether this should remain generic / not include company settings.

dgdonut (approved):
(can get rid of defaults when we do persistence)

PR #498, opened 2021-03-10
add disk usage monitor to worker

JeanetteBruno:
Add a disk usage monitor to the worker.  It is initialized with the list of directories we want to monitor.  The invoker will record the change in the directories sizes seen when running the task's command.

### DIFF
--- a/common/stats/dir_monitor.go
+++ b/common/stats/dir_monitor.go
@@ -11,18 +11,25 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-// DiskMonitor monitor disk usage for selected directories
-type DiskMonitor struct {
-	pathAbbrevs []string // these abbreviations are added to the end of the paths' usage gauge names
-	paths       []string
-	startSizes  []int64
-	endSizes    []int64
+var NopDirMonitor *DirMonitor = NewDirMonitor([]MonitorDir{})
+
+// MonitorDir a directory to monitor and a shortname (suffix) for reporting the stat
+type MonitorDir struct {
+	Directory  string // the directory to monitor
+	StatSuffix string // the suffix to use on the commandDirUsage_kb stat
 }
 
-// NewDiskMonitor return a DiskMonitor
-func NewDiskMonitor(pathAbbrevs []string, paths []string) *DiskMonitor {
-	dm := &DiskMonitor{pathAbbrevs: pathAbbrevs, paths: paths, startSizes: make([]int64, len(paths)), endSizes: make([]int64, len(paths))}
-	for i := range paths {
+// DirMonitor monitor disk usage for selected directories
+type DirMonitor struct {
+	dirs       []MonitorDir
+	startSizes []int64
+	endSizes   []int64
+}
+
+// NewDirMonitor return a DirMonitor
+func NewDirMonitor(dirs []MonitorDir) *DirMonitor {
+	dm := &DirMonitor{dirs: dirs, startSizes: make([]int64, len(dirs)), endSizes: make([]int64, len(dirs))}
+	for i := range dirs {
 		dm.startSizes[i] = -1
 		dm.endSizes[i] = -1
 	}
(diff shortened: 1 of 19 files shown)


common/stats/disk_monitor.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+		if start {
+			dm.startSizes[i] = asInt
+		} else {
+			dm.endSizes[i] = asInt
+		}
+	}
+}
+
+// GetDiskUsageKB use posix du to get disk usage of a dir, for simplicity vs syscall or walking dir contents
```
JeanetteBruno: the code from dir_config.go was moved here.

common/stats/disk_monitor.go
```diff
@@ -0,0 +1,97 @@
...
+	"os/exec"
+	"strconv"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// DiskMonitor monitor disk usage for selected directories
+type DiskMonitor struct {
+	shortName  []string
```
ryancouto: what's a shortName

common/stats/disk_monitor.go
```diff
@@ -0,0 +1,97 @@
...
+func (dm *DiskMonitor) RecordSizeStats(stat StatsReceiver) {
+	for i, shortName := range dm.shortName {
+		delta := dm.endSizes[i] - dm.startSizes[i]
+		statName := fmt.Sprintf("%s_%s", WorkerDirSizeChange, shortName)
+		stat.Gauge(statName).Update(delta)
+	}
+}
+
+// getStartSizes get the starting sized of the directories being monitored
+func (dm *DiskMonitor) getSizes(start bool) {
```
ryancouto: nit: i think "isStart" is clearer than just "start"

common/stats/stats_names.go
```diff
@@ -353,6 +353,12 @@ const (
 	*/
 	WorkerActiveRunsGauge = "activeRunsGauge"
 
+	/*
+		The disk size change for the indicated directory seen when running the task.
+		The stat will be qualified with a logical name for the directory
+	*/
+	WorkerDirSizeChange = "directorySizeChange_k"
```
ryancouto: why "_k"?

runner/runners/invoke.go
```diff
@@ -334,6 +339,11 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 		stdlog.Write([]byte(header))
 	}
 
+	if inv.diskMonitor != nil {
+		inv.diskMonitor.GetStartSizes()
```
ryancouto: because this is after the snapshot checkout, we aren't tracking the disk usage increase associated with it. is that intended?

JeanetteBruno: yes - this is just for the command execution

common/stats/stats_names.go
JeanetteBruno: the units are kb - I'll change it to kb.

common/stats/disk_monitor.go
JeanetteBruno: short name is the name the suffix that is appended to the disk_usage metric name.  from the test: short name is **cwd**, the metric is reported as **directorySizeChange_kb_cwd**.  I'm changing the field name to pathAbbrevs and adding a comment explaining how the abbreviations are used.  I'm also changing the metric name to: **commandDirUsageKb_cwd**.

common/stats/disk_monitor.go
JeanetteBruno: agreed, changed.

common/stats/disk_monitor.go
ryancouto: sgtm!

common/stats/disk_monitor.go
```diff
@@ -0,0 +1,97 @@
...
+	"os/exec"
+	"strconv"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// DiskMonitor monitor disk usage for selected directories
+type DiskMonitor struct {
+	pathAbbrevs []string // these abbreviations are added to the end of the paths' usage gauge names
```
ryancouto: think i'd prefer having paths & pathAbbrevs more closely linked than just relying on the indices to be correct. Could get confusing if we have a long list of paths. Maybe some pathAndAbbrev struct or even just a map[string]string

dgdonut: Was going to suggest the same thing, prefer a slice of structs with all of the fields that can be iterated over.

dgdonut: I hate to nit naming but pathAbbrevs isn't clear without reading the comment, suggest something like statSuffix

common/stats/stats_names.go
```diff
@@ -353,6 +353,12 @@ const (
 	*/
 	WorkerActiveRunsGauge = "activeRunsGauge"
 
+	/*
+		The disk size change for the indicated directory seen when running the task.
+		The stat will be qualified with a logical name for the directory
```
dgdonut: Qualified how - commandDirUsageKb_X? usageKbX? Typically we use _ to delimit these types of stats.

runner/runners/invoke.go
```diff
@@ -334,6 +339,11 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 		stdlog.Write([]byte(header))
 	}
 
+	if inv.diskMonitor != nil {
```
dgdonut: Let's create a diskMonitor with no paths that is effectively a no-op on runner construction in place of checking for nil every time we want to use it

common/stats/dir_monitor.go
```diff
@@ -0,0 +1,104 @@
...
+type DirMonitor struct {
+	dirs       []MonitorDir
+	startSizes []int64
+	endSizes   []int64
+}
+
+// NewDirMonitor return a DirMonitor
+func NewDirMonitor(dirs []MonitorDir) *DirMonitor {
+	dm := &DirMonitor{dirs: dirs, startSizes: make([]int64, len(dirs)), endSizes: make([]int64, len(dirs))}
+	for i := range dirs {
```
dgdonut: Looks like we did this half way, I meant even including the startSize and endSize in DirMonitor, so you don't need this for loop to initialize these values, and you don't have to index different slices anywhere in this file.

common/stats/dir_monitor.go
```diff
@@ -11,52 +11,49 @@ import (
...
 }
 
 // RecordSizeStats record the disk size deltas to the stats receiver
-func (dm *DirMonitor) RecordSizeStats(stat StatsReceiver) {
+func (dm *DirsMonitor) RecordSizeStats(stat StatsReceiver) {
 	for i, dir := range dm.dirs {
-		delta := dm.endSizes[i] - dm.startSizes[i]
-		statName := fmt.Sprintf("%s_%s", CommandDirUsageKb, dir.StatSuffix)
-		stat.Gauge(statName).Update(delta)
+		if dm.dirs[i].startSize != -1 && dm.dirs[i].endSize != -1 {
```
dgdonut: we are almost there, no need for dm.dirs[i] here, we already have dirs[i] as "dir" from the for loop definition, which we can clean up to
for _, dir := range dm.dirs {

common/stats/dir_monitor.go
```diff
@@ -69,9 +66,9 @@ func (dm *DirMonitor) getSizes(isStart bool) {
 			asInt = int64(dSize)
 		}
 		if isStart {
-			dm.startSizes[i] = asInt
+			dm.dirs[i].startSize = asInt
```
dgdonut: same here.

dgdonut (approved):
thank you

PR #499, opened 2021-03-16
dir monitor use dir size 0 if dir not exist

JeanetteBruno:


JeanetteBruno:
change is not necessary functionality is already in GetDiskUsageKB()

PR #503, opened 2021-03-29
scheduler - update default load class percents

dgdonut:


scheduler/server/load_based_sched_alg.go
```diff
@@ -21,10 +21,10 @@ const (
 // defaults for the LoadBasedScheduler algorithm: only one class and all jobs map to that class
 var (
 	DefaultLoadBasedSchedulerClassPercents = map[string]int32{
-		"land":       48,
+		"land":       40,
```
ryancouto, line 24: unrelated to this change, but this feels twitter internal-y... should we add a TODO to delete these defaults once persistence is deployed?

ryancouto (approved):

scheduler/server/load_based_sched_alg.go
dgdonut: We have an internal ticket but I will add a generic todo.

PR #506, opened 2021-04-01
scheduler - use an LRU cache to track taskDurations

dgdonut:
Address root cause by converting taskDurations to an LRU that will cap at a preset size, instead of a map that grows infinitely or has to be manually pruned.


dgdonut:
edited: test updated

scheduler/server/job_state.go
```diff
@@ -4,7 +4,9 @@ import (
 	"math"
 	"time"
 
+	lru "github.com/hashicorp/golang-lru"
 	log "github.com/sirupsen/logrus"
+
```
anthonyjpratti, line 9: nit: blank line

scheduler/server/stateful_scheduler.go
```diff
@@ -123,9 +129,10 @@ type averageDuration struct {
 	duration time.Duration
 }
 
-func (ad *averageDuration) update(d time.Duration) {
+func (ad *averageDuration) update(d time.Duration) *averageDuration {
 	ad.count++
 	ad.duration = ad.duration + time.Duration(int64(d-ad.duration)/ad.count)
+	return ad
```
anthonyjpratti: I don't see this used anywhere,  maybe remove?

anthonyjpratti (approved):

scheduler/server/job_state.go
dgdonut: added for consistency with other files (separate twitter/scoot pkgs from 3rdparty pkgs)

scheduler/server/stateful_scheduler.go
dgdonut: thanks, added for previous iteration, not needed anymore

PR #508, opened 2021-04-10
endpoints - no default /health endpoint

dgdonut:
Don't add a default handler that is only needed on aurora. Will be replaced internally.

### DIFF
--- a/common/endpoints/endpoints.go
+++ b/common/endpoints/endpoints.go
@@ -53,7 +53,7 @@ func (s *TwitterServer) Serve() error {
 }
 
 func helpHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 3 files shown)


common/endpoints/endpoints.go
```diff
@@ -56,14 +53,10 @@ func (s *TwitterServer) Serve() error {
 }
 
 func helpHandler(w http.ResponseWriter, r *http.Request) {
-	msg := "Common paths: '/health', '/admin/metrics.json', '/output', '/debug/pprof'"
+	msg := "Common paths: '/admin/metrics.json', '/output', '/debug/pprof'"
```
ryancouto: let's remove /output too

ryancouto (approved):

PR #509, opened 2021-04-15
update CI go version to 1.15

dgdonut:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/twitter/scoot
 
-go 1.12
+go 1.13
```
ryancouto, line 3: why not 15

ryancouto (approved):

go.mod
dgdonut: This is the minimum version required to import this package, not really anything to do with what version our repo runs at. Since we've been using 1.13 and that still works I would leave it there.

PR #511, opened 2021-04-26
sched - cap requestorHistory via lru cache

dgdonut:
Cap in same manner as #506. Also, cap the length of the slice that can be in each requestor's history.


dgdonut:
Also, I can't figure out why the "validate-changes (1.13.8, ubuntu-latest)" check keeps showing up and not completing. That was our old go version, but the ci.yml now uses 1.15.11.

JeanetteBruno (approved):

PR #516, opened 2021-06-17
additional logging and stats

JeanetteBruno:
we are seeing instances where a worker completes a task, but it takes a long time (over 20 minutes) for the scheduler to assign the worker a new task.  We are also seeing 10 minute timeout errors getting a job's status.

Added:
logging the time it takes to loop through all active tasks looking for completed tasks (in mailbox.go) and 
stats and logging for time it takes to build a job's status response.

### DIFF
--- a/async/mailbox.go
+++ b/async/mailbox.go
@@ -122,7 +122,7 @@ func (bx *Mailbox) ProcessMessages() {
 			unCompletedMsgs = append(unCompletedMsgs, msg)
 		}
 	}
-	log.Infof("processed (worker) messages in %v", time.Since(start))
+	log.Infof("processed %d (worker) messages in %v", len(bx.msgs), time.Since(start))
 
 	// reset inProgress messages to unCompletedMsgs only
 	bx.msgs = unCompletedMsgs
--- a/async/mailbox.go
+++ b/async/mailbox.go
@@ -1,5 +1,11 @@
 package async
 
+import (
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
 // An AsyncMailbox stores AsyncErrors and their associated callbacks
 // and invokes them once the AsyncError is completed
 //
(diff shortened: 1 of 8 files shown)


async/mailbox.go
```diff
@@ -115,6 +122,7 @@ func (bx *Mailbox) ProcessMessages() {
 			unCompletedMsgs = append(unCompletedMsgs, msg)
 		}
 	}
+	log.Infof("processed (worker) messages in %v", time.Since(start))
```
ryancouto: maybe add len(bx.msgs) here so we know how many we're processing in each cycle

scheduler/api/thrift/get_status.go
```diff
@@ -37,7 +46,15 @@ func GetJobStatus(jobId string, sc s.SagaCoordinator) (*scoot.JobStatus, error)
 		return js, nil
 	}
 
-	return convertSagaStateToJobStatus(state), nil
+	start = time.Now()
+	scootJs := convertSagaStateToJobStatus(state)
```
ryancouto: can we just call it js?

JeanetteBruno: done

async/mailbox.go
JeanetteBruno: done

async/mailbox.go
```diff
@@ -115,6 +122,7 @@ func (bx *Mailbox) ProcessMessages() {
 			unCompletedMsgs = append(unCompletedMsgs, msg)
 		}
 	}
+	log.Infof("processed %d (worker) messages in %v", len(bx.msgs), time.Since(start))
```
ryancouto, line 125: can we pass the stats handler through to here too & emit a latency gauge for each message callback?

async/mailbox.go
```diff
@@ -105,6 +111,7 @@ func (bx *Mailbox) NewAsyncError(cb AsyncErrorResponseHandler) *AsyncError {
 // the callback function and removes the message from the mailbox
 func (bx *Mailbox) ProcessMessages() {
 	var unCompletedMsgs []message
+	start := time.Now()
```
dgdonut, line 114: We already capture this in stateful_scheduler.go (stats.SchedProcessMessagesLatency_ms)

JeanetteBruno:
the latency is already reported in scheduler via stat name: scheduler/schedProcessMessagesLatency_ms

PR #517, opened 2021-06-21
Add logging to node updates

ryancouto:


cloud/cluster/state.go
```diff
@@ -52,6 +56,7 @@ func (s *state) setAndDiff(newState []Node) []NodeUpdate {
 	// reset nodes map, assign to new state
 	s.nodes = make(map[NodeId]Node)
 	for _, n := range newState {
+		log.Infof("Node %s included in new cluster state", n)
```
JeanetteBruno: instead of printing this, how about printing:
number of nodes added, number of nodes removed, then also print same info (number of nodes added, number of nodes removed) from cluster_state.go?

wdyt?

JeanetteBruno (approved):

cloud/cluster/state.go
```diff
@@ -37,13 +39,15 @@ func (s *state) setAndDiff(newState []Node) []NodeUpdate {
 	sort.Sort(NodeSorter(removed))
 	outgoing := []NodeUpdate{}
 	for _, n := range added {
+		log.Infof("NodeAdded update: %s", n)
```
JeanetteBruno, line 43: should the formatting directive be %v?

cloud/cluster/state.go
```diff
@@ -37,13 +39,15 @@ func (s *state) setAndDiff(newState []Node) []NodeUpdate {
...
 		outgoing = append(outgoing, NodeUpdate{
 			UpdateType: NodeAdded,
 			Id:         n.Id(),
 			Node:       n,
 		})
 	}
 	for _, n := range removed {
+		log.Infof("NodeRemoved update: %s", n)
```
JeanetteBruno, line 51: same as above

cloud/cluster/state.go
ryancouto: nope - n is a Node, and Node has a String() method, so %s should work fine

cloud/cluster/state.go
ryancouto: added

PR #519, opened 2021-07-01
fix requestor count stats

JeanetteBruno:


scheduler/server/stateful_scheduler_test.go
```diff
@@ -839,3 +839,44 @@ func validateCompletionCounts(s *statefulScheduler, t *testing.T) {
 	assert.Equal(t, totalMapRunning, running, "running count from tasksByJobClassAndStartTimeSec map (%d) is not equal to running from scheduler (%d)",
 		totalMapRunning, running)
 }
+
+func Test_StatefulScheduler_RequestorCountsStats(t *testing.T) {
+	sc := sagalogs.MakeInMemorySagaCoordinatorNoGC()
+	s, _, _ := initializeServices(sc, false)
+	s.SetClassLoadPercents(map[string]int32{"land": 60, "diff": 40})
+	s.SetRequestorToClassMap(map[string]string{"land": "land", "diff": "diff"})
```
ryancouto: nit: land/diff are internal terminology. would prefer to use some fake classes here

scheduler/server/stateful_scheduler.go
```diff
@@ -560,7 +559,7 @@ func (s *statefulScheduler) updateStats() {
 	// get job and task counts by requestor, and overall jobs stats
 	for _, job := range s.inProgressJobs {
 		requestor := job.Job.Def.Requestor
-		if _, ok := s.requestorsCounts[requestor]; !ok {
+		if _, ok := requestorsCounts[requestor]; !ok {
```
ryancouto, line 562: nice catch

ryancouto (approved):

PR #520, opened 2021-07-12
fix create jobState, order tasks by average duration (desc)

JeanetteBruno:
- newJobState() uses task duration key extractor function when setting a new task's average duration

- addOrUpdateTaskDuration() only add when the task's key is not in the lru cache, otherwise update the current lru cache entry.

### DIFF
--- a/scheduler/server/job_state.go
+++ b/scheduler/server/job_state.go
@@ -71,7 +71,7 @@ func (s taskStatesByDuration) Less(i, j int) bool {
 // The jobState will reflect any previous progress made on this job and logged to the Sagalog
 // Note: taskDurations is optional and only used to enable sorts using taskStatesByDuration above.
 func newJobState(job *domain.Job, jobClass string, saga *saga.Saga, taskDurations *lru.Cache,
-	tasksByJobClassAndStartTimeSec map[taskClassAndStartKey]taskStateByJobIDTaskID) *jobState {
+	tasksByJobClassAndStartTimeSec map[taskClassAndStartKey]taskStateByJobIDTaskID, durationKeyExtractor func(string) string) *jobState {
 	j := &jobState{
 		Job:                            job,
 		Saga:                           saga,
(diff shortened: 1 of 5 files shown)


scheduler/server/job_state.go
```diff
@@ -88,10 +88,12 @@ func newJobState(job *domain.Job, jobClass string, saga *saga.Saga, taskDuration
 
 	for _, taskDef := range job.Def.Tasks {
 		var duration time.Duration
+		origKey := taskDef.TaskID
```
ryancouto: nit: why do we need origKey?

scheduler/server/stateful_scheduler.go
```diff
@@ -306,6 +311,11 @@ func NewStatefulScheduler(
 		return nil
 	}
 
+	dkef := durationKeyExtractorFn
```
ryancouto, line 315: nit: why do we need this this assignment?

scheduler/server/stateful_scheduler.go
```diff
@@ -1276,19 +1283,19 @@ func addOrUpdateRequestorHistory(requestorHistory *lru.Cache, requestor, newHist
...
 	var ad *averageDuration
-	iface, ok := taskDurations.Get(taskId)
+	iface, ok := taskDurations.Get(durationKey)
 	if !ok {
 		ad = &averageDuration{count: 1, duration: d}
+		taskDurations.Add(durationKey, ad)
 	} else {
 		ad, ok = iface.(*averageDuration)
 		if !ok {
 			return
```
ryancouto, line 1297: should we log some error here?

scheduler/server/job_state.go
JeanetteBruno: removed it

scheduler/server/stateful_scheduler.go
JeanetteBruno: this provides a default eliminating if fn == nil tests later in the code

ryancouto: whoops didn't mean to resolve. i just meant - why can't we just say

`
if durationKeyExtractorFn == nil {
		durationKeyExtractorFn = nopDurationKeyExtractor
}
`

dgdonut (approved):
lgtm less open comments

scheduler/server/stateful_scheduler.go
JeanetteBruno: I thought changing parameter values is a frowned upon practice.

PR #521, opened 2021-07-16
fix bug in suspending worker when workerapi can't get task result

JeanetteBruno:
A new task is assigned to a lost worker even though the prior task on that worker recognized that the worker was lost.... We had a scenario where a worker restarted, the task running on it failed but the worker was not put in the suspended list because the task callback function's call to cluster_state.go taskCompleted() had flaky=false.  

root cause:
taskRunner.go's run() method changed the state from UNKNOWN to FAILED (near line 110).  As a result the stateful_scheduler.go ln 1019 set flaky to false.


cloud/cluster/state.go
```diff
@@ -55,7 +55,9 @@ func (s *state) setAndDiff(newState []Node) []NodeUpdate {
 		})
 	}
 
-	log.Infof("Number of nodes added: %d\nNumber of nodes removed: %d\nNumber of nodes in newState: %d\nNumber of nodes in old state: %d", len(added), len(removed), len(newState), oldStateLen)
+	if len(added) > 0 || len(removed) > 0 {
```
JeanetteBruno, line 62: decreasing noise in scheduler log

scheduler/server/cluster_state.go
```diff
@@ -237,25 +237,25 @@ func (c *clusterState) update(updates []cluster.NodeUpdate) {
 			adds += 1
 			// UserInitiated is true only when this NodeUpdate comes from a ReinstateWorker or OfflineWorker request
 			if update.UserInitiated {
-				log.Infof("Reinstating node %s", update.Id)
+				log.Infof("NodeAdded: Reinstating node %s", update.Id)
```
JeanetteBruno, line 241: adding common Add, Remove, Suspend qualifiers to log statements to make it easier to trace a node's handling in the log

scheduler/server/task_runner.go
```diff
@@ -107,7 +107,6 @@ func (r *taskRunner) run() error {
 
 	// Update taskErr state if it's empty or if we're doing deadletter..
 	if taskErr.st.State == runner.UNKNOWN {
-		taskErr.st.State = runner.FAILED
```
JeanetteBruno, line 110: this is the actual fix.

scheduler/server/task_runner_test.go
```diff
@@ -201,6 +201,37 @@ func Test_runTaskAndLog_MarkFailedTaskAsFinished(t *testing.T) {
...
+	msgMatcher = TaskMessageMatcher{Type: &sagaEndTask, JobId: "job1", TaskId: "task1", Data: gomock.Any()}
+	sagaLogMock.EXPECT().LogMessage(msgMatcher)
+
+	sagaCoord := saga.MakeSagaCoordinator(sagaLogMock)
+	s, _ := sagaCoord.MakeSaga("job1", nil)
+
+	err := get_testTaskRunner(s, chaos, "job1", "task1", task, true, stats.NilStatsReceiver()).run()
+
+	assert.NotNil(t, err.(*taskError).runnerErr)
+	assert.Equal(t, err.(*taskError).st.State, runner.UNKNOWN)
```
JeanetteBruno, line 231: returning UNKNOWN will trigger setting flaky flag to true in the task callback in stateful_scheduler.go

scheduler/server/task_runner.go
```diff
@@ -107,7 +107,6 @@ func (r *taskRunner) run() error {
 
 	// Update taskErr state if it's empty or if we're doing deadletter..
```
ryancouto: update comment

scheduler/server/task_runner.go
ryancouto: runner.UNKNOWN state might not be correctly mapped to a failing exit code for the client

JeanetteBruno: we tested the behavior in jbruno/devel (and rcouto/devel): started tryout job, killed one of the workers running a task and saw sickle report the task as failed and restarting.
sample output:
I] 15:02:55] Task Failure: src/go/twitter.biz/mgci/sq/validator:: (pants), (job ID: d775a460-ef42-4297-4504-8600ed23fa80, task prefix: -4), finished in 107s
I] 15:02:55]
I] 15:03:01] Retrying ./scoot/config/scripts/run_task/pants/pantsRun.sh ./pants --pants-config-files=pants.ini --pants-config-files=pants.ini.scoot --pants-supportdir=pants-support --no-enable-pantsd test lint src/go/twitter.biz/mgci/sq/validator::, (job ID: 50b2d799-90d3-414c-6edd-89ffdb68137b)
I] 15:03:01] Starting testing of 1 target(s). Job ID: 50b2d799-90d3-414c-6edd-89ffdb68137b

scheduler/server/task_runner.go
JeanetteBruno: done

scheduler/server/task_runner.go
dgdonut: yeah, only thing I can think of is that this could cause some odd downstream behavior. Sickle does seem to generally group this together with FAILED as "lost" though, and has a few tests referencing unknown.

Ultimately the comment in stateful scheduler around where we set flaky is leading me to believe the ultimate root cause is poor distinction around these states generally. But this might be enough to fix not suspending workers in this particular case

dgdonut (approved):

PR #523, opened 2021-07-22
get rid of os/temp

ryancouto:
we don't need it!


runner/runners/local_output.go
```diff
@@ -44,6 +44,10 @@ func NewHttpOutputCreator(tmp *temp.TempDir, httpUri string) (HttpOutputCreator,
 		}
 		httpPath = strings.TrimSuffix(u.Path, "/") + "/"
 	}
+	tmp, err := ioutil.TempDir("", "output")
```
JeanetteBruno, line 47: with the orig code, the output dir would be under scoot-tmp-<rand #>/ .... should we try to preserve that behavior?

JeanetteBruno (commented):
I'm still looking this over, but it might be easier to understand the change if it was broken into multiple changes - eliminate os/temp from one binary at a time.   

(when you eliminate os/tmp from scheduler you may find that scheduler doesn't need ioutils.tmpDir at all - I just took a look at the running scheduler and it only had an empty scoot-tmp directory)

runner/runners/local_output.go
ryancouto: imo it doesn't really matter. we do still set TMPDIR=$(pwd)/scoot-tmp in our aurora file, so this output dir will be located at < path to aurora sandbox >/scoot-tmp/output-<rand #>

ryancouto:
> I'm still looking this over, but it might be easier to understand the change if it was broken into multiple changes - eliminate os/temp from one binary at a time.
> 
> (when you eliminate os/tmp from scheduler you may find that scheduler doesn't need ioutils.tmpDir at all - I just took a look at the running scheduler and it only had an empty scoot-tmp directory)

i can try to pull this apart if that'd be helpful

snapshot/git/gitfiler/setup.go
```diff
@@ -43,26 +39,3 @@ func NewSingleRepoCheckouter(repoIniter PooledRepoIniter,
 	pool := NewSingleRepoPool(repoIniter, stat, doneCh)
 	return NewCheckouter(pool, stat)
 }
-
-// A Checkouter that creates a new repo with git clone --reference for each checkout
-func NewRefRepoCloningCheckouter(refRepoIniter PooledRepoIniter,
```
dgdonut, line 48: This is still called in closed source (for mac-based workers I believe, which meh, but worth calling out that this should trigger deprecation of that completely)

dgdonut:
Need some time to review everything in more detail, but only thing we may end up tripping over here would be cases where we could have been relying on reuse of a previous tempdir across multiple operations (something expecting files to be leftover there). I can't remember specifically if/where we do that, just something to watch out for.

snapshot/git/gitfiler/setup.go
ryancouto: Ah - could also just leave it in & change the implementation of the refCloner so that it takes a path rather than a *temp.TempDir

dgdonut (approved):

snapshot/git/gitfiler/setup.go
dgdonut: I think I am gonna get rid of macbook filer compatibility altogether, fyi

ryancouto: ah missed this when I merged. fine with me, can remove all this stuff with that removal.

PR #526, opened 2021-07-27
Remove unused Erase() from workerapi & runner interfaces

ryancouto:


bazel/cas/mock_bytestream/bsclient_mock.go
```diff
@@ -38,6 +38,7 @@ func (m *MockByteStreamClient) EXPECT() *MockByteStreamClientMockRecorder {
 
 // QueryWriteStatus mocks base method
 func (m *MockByteStreamClient) QueryWriteStatus(arg0 context.Context, arg1 *bytestream.QueryWriteStatusRequest, arg2 ...grpc.CallOption) (*bytestream.QueryWriteStatusResponse, error) {
+	m.ctrl.T.Helper()
```
dgdonut: Why'd these change?

dgdonut (approved):

bazel/cas/mock_bytestream/bsclient_mock.go
ryancouto: I ran a `make generate` to create new thrift files, forgetting that the thrift generation is done via `make thrift`

Saw that these changed, but didn't change them back.

PR #528, opened 2021-07-28
moving setup to top level directory

ralphmcneal:
This refactor is one of several to eliminate the binaries directory.


ryancouto (commented):
There's still a reference to setup-cloud-scoot in binaries/README.md, that should be removed. Otherwise LGTM!

dgdonut (approved):

PR #530, opened 2021-07-29
refactor smoketest

H-Chaudhary:


scheduler/client/cli/cli.go
```diff
@@ -5,23 +5,16 @@ import (
...
-}
-
-// Implements CLIClient - basic
-type simpleCLIClient struct {
-	rootCmd     *cobra.Command
-	addr        string
-	dial        dialer.Dialer
-	logLevel    string
-	scootClient *client.CloudScootClient
+type SimpleCLIClient struct {
```
ryancouto: Could we name this schedCLI or schedClient or something more descriptive

common/client/client.go
```diff
@@ -0,0 +1,27 @@
+package client
```
ryancouto, line 1: if we renamed this dir/pkg to cli, we could get rid of the import stutter (client.*Client -> cli.Client)

scheduler/client/cli/cli.go
```diff
@@ -5,23 +5,16 @@ import (
...
+	simpleCLIClient commoncli.SimpleClient
```
ryancouto: we could embed the commoncli.SimpleClient here as an unnamed field. Means we can skip a layer of nested calls (e.g. c.simpleCLIClient.RootCmd.Execute() becomes c.RootCmd.Execute())

ryancouto (commented):
Most LGTM! Just a couple naming nits really

scheduler/client/cli/cli.go
harshita-chaudhary: Makes sense, renaming it to SchedCLIClient

common/client/client.go
harshita-chaudhary: There's another package named cli in the client package, which has a cli.go, so it would be the same issue if we renamed it.

scheduler/client/cli/cli.go
harshita-chaudhary: We can embed it as an unnamed field but the struct is being passed to the cmd.Run() function, I'm unsure if we can pass it if it is unnamed.

harshita-chaudhary: Thanks for pointing this out, done.

integration-tests/smoketest/smoketest.go
```diff
@@ -11,18 +11,19 @@ import (
```
dgdonut, line 11: At top of file package cli -> change to package smoketest?

scheduler/client/cli/cli.go
```diff
@@ -5,23 +5,16 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	commoncli "github.com/twitter/scoot/common/client"
```
dgdonut, line 9: nit: some files don't match this, but we usually space the imports like this:
`import (
  "fmt"
  
  "github.com/spf13/cobra"
  <other non-scoot imports>
  
  "github.com/twitter/scoot/common/dialer"
  <other twitter/scoot imports>
)`

common/client/client.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/twitter/scoot/common/dialer"
+	"github.com/twitter/scoot/scheduler/client"
+)
+
+// Client interface that includes CLI handling
+type CLIClient interface {
+	Exec() error
+}
+
+// Implements CLIClient - basic
```
dgdonut: Does this actually implement CLIClient iface? I don't see the Exec() implementation - looks like we use this as base struct ontop of which we build actual implementations - can the comment make that clearer?

dgdonut (approved):
lgtm less some minor suggestions

integration-tests/smoketest/smoketest.go
harshita-chaudhary: Done

scheduler/client/cli/cli.go
harshita-chaudhary: Makes sense, changed it.

common/client/client.go
harshita-chaudhary: Made it clearer, lmk if it needs to be more informative.

dgdonut (approved):

PR #536, opened 2021-07-30
eliminate scheduler's use of ice, make ci works

JeanetteBruno:
summary of change:
- update scheduler's main.go to create the necessary objects directly when starting the server
- eliminated scheduler's use of the config.go generated by bindata: moved the json configurations into variable defnitions (see scheduler/config/config.go)
- removed scheduler ice files.


scheduler/scheduler/config/config_new.go
```diff
@@ -0,0 +1,182 @@
+package config
```
ryancouto: why config_new?

scheduler/scheduler/start_server.go
```diff
@@ -0,0 +1,102 @@
+package main
```
ryancouto: think server.go would be a better name

scheduler/server/stateful_scheduler.go
```diff
@@ -187,6 +195,13 @@ type statefulScheduler struct {
 
 	// durationKeyExtractorFn - function to extract, from taskID, the key to use for tracking task average durations
 	durationKeyExtractorFn func(string) string
+
+	SchedAlg SchedulingAlgorithm
+}
+
+func (s *statefulScheduler) String() string {
+	return fmt.Sprintf("%s, num nodes: %d, num suspended nodes: %d, maxFlakyDuration: %s, maxLostDuration: %s",
```
ryancouto, line 203: would we be interested in seeing which nodes were suspended instead of just the # of them?

scheduler/scheduler/config/config_new.go
JeanetteBruno: renamed to config.go

scheduler/scheduler/start_server.go
JeanetteBruno: I thought that would be confusing with scheduler/api/server.go and scheduler/server directory.

Since this file is not an object definition - its main feature is the StartServer function (which is not a member function for an object), using the action_name (start_server.go) makes more sense to me.

scheduler/server/stateful_scheduler.go
JeanetteBruno: at this point, I think this is only used at scheduler startup (I added it to debug an issue with the change).  I would prefer to hold off changing till we really need it.

JeanetteBruno:
this pr is replaced by pr 543

PR #539, opened 2021-07-30
reuse persistent temp dir where needed

dgdonut:


dgdonut:
branch got messed up when rebasing, see #541

PR #541, opened 2021-08-02
reuse persistent temp dir where needed

dgdonut:


binaries/scoot-snapshot-db/main.go
```diff
@@ -53,6 +56,11 @@ func (i *injector) RegisterFlags(rootCmd *cobra.Command) {
 }
 
 func (i *injector) Inject() (snapshot.DB, error) {
+	dbTempDir, err := ioutil.TempDir("", "")
+	if err != nil {
+		return nil, err
+	}
+
```
ryancouto, line 61: not part of this PR, but could we remove the comment below (or at least the parts about talking to the daemon)

binaries/scoot-snapshot-db/main.go
```diff
@@ -36,11 +37,13 @@ func main() {
 
 	inj := &injector{}
 	cmd := cli.MakeDBCLI(inj)
-	if err := cmd.Execute(); err != nil {
-		removeTemp()
+	err = cmd.Execute()
+	if dbTempDir != "" {
+		os.RemoveAll(dbTempDir)
```
ryancouto: when will dbTempDir not be ""?

snapshot/git/gitdb/setup.go
```diff
@@ -38,6 +38,9 @@ func (m module) Install(b *ice.MagicBag) {
 		func() AutoUploadDest {
 			return AutoUploadBundlestore
 		},
+		func() string {
```
ryancouto, line 41: honestly don't really understand why we need this... is this dbTempDir?

binaries/scoot-snapshot-db/main.go
dgdonut: If tempdir creation fails. Otherwise it will do os.RemoveAll("") - which I'm not sure exactly what that would do. Nothing? Or delete everything in your cwd?

snapshot/git/gitdb/setup.go
dgdonut: oss apiserver main.go implicitly calls MakeDBNewRepo here via ice, and that func needs a string for (a preexisting) tmpdir path. In closed source we call it explicitly, without ice

binaries/scoot-snapshot-db/main.go
dgdonut: checked the os pkg implementation and "" is special cased to do nothing and return nil, so removing the check.

snapshot/git/gitdb/setup.go
ryancouto: Does that mean that the oss MakeDBNewRepo will use / as temp then? Should we do os.TempDir() here instead of ""?

dgdonut: It will effectively pass "" through which will result in ioutil.TempDir("", ...) calls, which is equivalent to using os.TempDir()

dgdonut: ex:

> $ go run github.com/twitter/scoot/apiserver
INFO[0000] Making new FileStore at dir: /var/folders/y0/6frrxtkn495dgbzr_s_7zcgr0000gp/T/bundles074177289  sourceLine="store/file_store.go:26"
INFO[0000] Starting new bundlestore.Server with root: /var/folders/y0/6frrxtkn495dgbzr_s_7zcgr0000gp/T/bundles074177289  sourceLine="bundlestore/server.go:34"
INFO[0000] git.NewRepository: /var/folders/y0/6frrxtkn495dgbzr_s_7zcgr0000gp/T/gitdb-repo-541018836, top: /private/var/folders/y0/6frrxtkn495dgbzr_s_7zcgr0000gp/T/gitdb-repo-541018836  sourceLine="repo/repo.go:132"

ryancouto (approved):

PR #543, opened 2021-08-03
remove scheduler's use of ice and bindata

JeanetteBruno:
this pr replaces #536 

summary of change:
- update scheduler's main.go to create the necessary objects directly when starting the server
- eliminated scheduler's use of the config.go generated by bindata: moved the json configurations into variable defnitions (see scheduler/config/config.go)
- removed scheduler ice files.


cheyenne-martinez (approved):

perftests/scheduler_simulator/test_alg.go
```diff
@@ -427,8 +427,8 @@ func (st *SchedulingAlgTester) makeTestCluster(num int) *testCluster {
 	return h
 }
 
-func (st *SchedulingAlgTester) getTestConfig() server.SchedulerConfig {
-	return server.SchedulerConfig{
+func (st *SchedulingAlgTester) getTestConfig() server.SchedulerConfiguration {
```
ryancouto, line 430: jw - why the rename to SchedulerConfiguration? Makes this diff bigger/harder to pick out which parts are actual changes

scheduler/scheduler/config/configs.go
```diff
@@ -0,0 +1,72 @@
...
+	  "MaxRetriesPerTask" : 0,
+	  "DefaultTaskTimeout" : "30m"
+	},
+	"SagaLog": {
+	  "Type": "memory"
+	}
+}
+`
+
+// localLocal config for local.local - !!! make sure this constant is added to SchedulerConfigs map above !!!
```
ryancouto, line 30: is the !!! part needed in this comment?

scheduler/scheduler/config/configs.go
```diff
@@ -0,0 +1,72 @@
...
+	  "DefaultTaskTimeout" : "30m"
+	},
+	"SagaLog": {
+	  "Type": "file",
+	  "Directory": ".scootdata/filesagalog"
+	}
+  }
+  `
+
+// localMemory config for local.memory - !!! make sure this constant is added to SchedulerConfigs map above !!!
```
ryancouto, line 54: ditto

scheduler/scheduler/main.go
```diff
@@ -45,50 +44,39 @@ func main() {
 	}
 	log.SetLevel(level)
 
-	configText, err := jsonconfig.GetConfigText(*configFlag, config.Asset)
+	schedulerJSONConfigs, err := config.GetSchedulerConfigs(*configFlag)
 	if err != nil {
-		log.Fatal(err)
+		panic(fmt.Errorf("error parsing schedule server config.  Scheduler not started. %s", err))
```
ryancouto: why panic instead of log.Fatal?

scheduler/starter/modules.go
```diff
@@ -0,0 +1,208 @@
...
+	return &socksSocket{
+		conn:    nil,
+		addr:    addr.String(),
+		timeout: 0, // thrift.TSocket gives us no way to extract the timeout
+		dialer:  f.dialer,
+	}
+}
+
+// socksSocket creates a socket via the proxy dialer.
+// (Ideally, there would be a form of TSocket that took a Dialer instead of a net.Conn)
```
ryancouto, line 128: remove?

scheduler/starter/modules.go
```diff
@@ -0,0 +1,208 @@
...
+			return err
+		}
+	}
+
+	s.conn = thrift.NewTSocketFromConnTimeout(conn, 0)
+	return nil
+}
+
+func (s *socksSocket) Read(buf []byte) (int, error) {
+	return s.conn.Read(buf)
```
ryancouto: should we embed conn as an unnamed field? means we can skip defining these funcs

ryancouto (commented):
Mostly lgtm! Couple little nits.

perftests/scheduler_simulator/test_alg.go
JeanetteBruno: I (and json marshaling) was tripping over multiple uses of SchedulerConfig - 'SchedulerConfig' in the json structure is being parsed to the SchedulerJSONConfig field, then the code translates that to the scheduler's SchedulerConfiguration object.

scheduler/scheduler/config/configs.go
JeanetteBruno: to draw attention to it - if the config is not added to the map, it will not be available.

dgdonut: can we please add a comment to the map above that indicates that is what defines what configs are available

perftests/scheduler_simulator/test_alg.go
ryancouto: oof gotcha

scheduler/starter/modules.go
```diff
@@ -0,0 +1,208 @@
...
+
+	// An Address is Socks-able if it starts with 10, because Twitter's production network lives in
+	// 10/8 in IPv4 space.
+	if ip[0] != 10 {
+		return trans
+	}
+
+	if f.dialer == nil {
+		// We want to use SOCKS, but can't, because we have no dialer
+		log.Info("Aurora address specified, but cannot use SOCKS proxy because SOCKS_PORT is unset", addr.String())
```
dgdonut: oss shouldn't reference "Aurora". not sure what that implies as to what this logic was supposed to be doing internally

scheduler/scheduler/config/configs.go
ryancouto: sure - but it seems self-evident that it would need to be there imo

scheduler/starter/modules.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/twitter/scoot/runner"
+	"github.com/twitter/scoot/runner/runners"
+	"github.com/twitter/scoot/saga"
+	"github.com/twitter/scoot/saga/sagalogs"
+	"github.com/twitter/scoot/scheduler/scheduler/config"
+	"github.com/twitter/scoot/scheduler/server"
+	"github.com/twitter/scoot/workerapi/client"
+)
+
+const DEFAULT_SOCKS_PORT = "50001"
```
dgdonut, line 25: Does all the low level proxy code belong here, vs common?

scheduler/starter/modules.go
JeanetteBruno: Not sure if you are suggesting remove sockSocket structure or the comment.... afaict, we have this sockSocket so we can inject the proxy.SOCKS5 dialer and with localhost:<port>.  I'm trying to minimize opportunity for introducing errors, so prefer to initially move this from closed source to here.

We can create a ticket to eliminate sockSocket struct altogether, but I don't think we want to do that as part of this change.

scheduler/starter/modules.go
JeanetteBruno: see prior comment

scheduler/starter/modules.go
JeanetteBruno: imo common should only contain functionality that is used by > 1 service.

scheduler/scheduler/config/configs.go
dgdonut: yeah - mainly as opposed to commenting the each of the config definitions separately

scheduler/starter/modules.go
JeanetteBruno: modified the comment

scheduler/scheduler/main.go
JeanetteBruno: changed back to log.Fatal()

scheduler/scheduler/config/configs.go
ryancouto: oh weird idk how these 2 convos got merged. wasn't responding to you drew! but yeah would rather have the one comment than the multiple

JeanetteBruno: I think we want the comment on each config definition in case someone cuts and paste's a definition without the top map visible in their editor screen....hopefully when they update the pasted comment they will see the message to also add the new config up top of the file.

scheduler/starter/modules.go
ryancouto: not sure i follow... if we had the socksSocket struct defined like this

`
type socksSocket struct {

	*thrift.TSocket

	addr    string

	timeout time.Duration

	dialer  proxy.Dialer

}
`

we could remove Read, Write, Close, Flush, & RemainingBytes from here, as socksSocket.Read(buf) would just delegate to the *thrift.TSocket

common/endpoints/endpoints.go
```diff
@@ -32,6 +34,10 @@ type TwitterServer struct {
 	Handlers map[string]http.Handler
 }
 
+func (s *TwitterServer) String() string {
+	return fmt.Sprintf("TwitterServer:Addr:%s, Stats:%s, Handlers:%v", s.Addr, s.Stats, s.Handlers)
```
ryancouto: stats.StatsReceiver doesn't have a String() method, so the %s won't be that helpful. Recommend using %+v instead (maybe same with the Handlers)

scheduler/scheduler/main.go
```diff
@@ -1,27 +1,25 @@
...
 	"github.com/twitter/scoot/common/log/hooks"
-	"github.com/twitter/scoot/common/stats"
-	"github.com/twitter/scoot/config/jsonconfig"
-	"github.com/twitter/scoot/config/scootconfig"
 	"github.com/twitter/scoot/scheduler"
-	"github.com/twitter/scoot/scheduler/api"
 	"github.com/twitter/scoot/scheduler/scheduler/config"
-	"github.com/twitter/scoot/scheduler/server"
-	"io/ioutil"
+	starter "github.com/twitter/scoot/scheduler/starter"
```
ryancouto: don't need alias here

scheduler/server/stateful_scheduler.go
```diff
@@ -346,6 +361,8 @@ func NewStatefulScheduler(
 	}
 	sched.loadSettings()
 
+	log.Info(sched.String())
```
ryancouto: you can just use sched here, it automatically runs String() to log

ryancouto (approved):
lgtm less nits

scheduler/scheduler/config/config.go
```diff
@@ -1,283 +1,174 @@
...
-}
+// How long to wait between runner status queries to determine [init] status.
+const DefaultReadyFnBackoff = 5 * time.Second
 
-type bindataFileInfo struct {
-	name    string
-	size    int64
-	mode    os.FileMode
-	modTime time.Time
+// SchedulerServerConfig config structure holding original parsed json configs and
```
dgdonut: comment for "SchedulerServerConfig" referencing "JSONConfigs" struct

dgdonut (approved):

PR #544, opened 2021-08-04
Get rid of async package & remove unused runner interfaces

ryancouto:
we don't need them!


workerapi/client/client.go
```diff
@@ -143,14 +122,16 @@ func (c *simpleClient) StatusAll() ([]runner.RunStatus, runner.ServiceStatus, er
...
+	svc := runner.ServiceStatus{Initialized: ws.Initialized, Error: svcErr}
 	if err != nil {
 		return nil, runner.ServiceStatus{}, err
 	}
 	st, err = runners.StatusesRO(st).QueryNow(q)
 	return st, svc, err
 }
+
+func (c *simpleClient) Query(q runner.Query, w runner.Wait) ([]runner.RunStatus, runner.ServiceStatus, error) {
+	// TODO: implement
```
dgdonut, line 135: ?

dgdonut (commented):
There is a significant behavior change here from changing the async callback portions to execute at a specific part in the sched step loop, to happening immediately after the async function portion completes. This could have performance implications and these modify the inProgressJobs struct which has no concurrency protection (and could be in the middle of e.g. task schedule and assignment). From a purely "refactor" perspective it would be good to preserve the async ProcessMessages functionality in some way to keep processing behavior as-is.

workerapi/client/client.go
ryancouto: Whoops - had to add this to collapse the StatusQueryNower iface into the StatusQuerier, even though we don't use this. Meant to add a better comment.

JeanetteBruno:
I agree with Drew's comment (keep processing the tasks' results as a specific action in the loop).  I was expecting the changes to be along the following lines:

- create a scheduler/server/task_handler package
- move task_runner.go to task_handler
- AsyncError object renamed to taskResult and moved to taskRunner
- mailbox renamed to  resultsCollector and moved to taskRunner
- runner can actually stay as runner moved to taskRunner
- the anonymous callback fn in stateful_scheduler.go also be moved into task_handler/

we can attack cleaning up the worker api implementation (Controller, Reader, StatusQuerier, QuerierNow, etc.) at a later date.

dgdonut:
For the record, I'm totally on board with the worker api simplification in this diff. IF it does not cause breaking API changes, i.e. when this lands we cannot redeploy scoot components safely without additional changes.

PR #547, opened 2021-08-11
Remove worker binary config files

cheyenne-martinez:


dgdonut (dismissed):
LGTM - not sure what the problem with ci is, make ci passed for me locally. It may just be stuck waiting for resources.

harshita-chaudhary:
I noticed that deleting binaries/workerserver dir is causing this, don't know why.

dgdonut:
Yeah, I think we broke a bunch of things in #548 actually. Because we still had a binaries/workerserver dir in that branch, everything worked, because a bunch of things actually depend on "workerserver". In this diff we renamed workerserver to worker, which is going to cause issues. We're going to need to preserve the "workerserver" binary name, because we rely on a bunch of ps commands to deal with that binary, and "worker" is way too generic and likely most machines have dozens of procs running as the name "worker"

dgdonut:
I opened #549 and it looks like CI is passing there, recommend we merge that and then update this PR based on that.

harshita-chaudhary:
Makes sense, I also have lots of files being added/changed under worker. Will merge after Cheyenne's PR. Thanks for fixing this issue.

cheyenne-martinez:
Ok I'll update this with changes from [#549](https://github.com/twitter/scoot/pull/549)

dgdonut (approved):
there we go

PR #550, opened 2021-08-12
Refactor workerapi

harshita-chaudhary:


scheduler/server/stateful_scheduler.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/twitter/scoot/runner"
 	"github.com/twitter/scoot/saga"
 	"github.com/twitter/scoot/scheduler/domain"
```
cheyenne-martinez, line 24: should this also have an alias so the domain isn't confused as the worker domain?

harshita-chaudhary: Makes sense, but here, since we're aliasing the worker domain as 'worker', I think it's simpler to keep domain for scheduler(it's the domain for the top level dir we're working under).

cheyenne-martinez (approved):

JeanetteBruno:
if you're still making changes, the following could be moved to worker/worker directory - they're part of instantiating the worker service. If you don't want to make the change now, that's also fine, they will be eliminated when we get rid of ice creation of worker.
- worker/api/setup.go
- runner/runners/setup.go

Also, are the changes to twitter (source) scoot ready?   If this change requires path updates in twitter scoot, it shouldn't be landed till we have the twitter scoot updates ready.

JeanetteBruno (approved):

harshita-chaudhary:
As per our discussion, skipping moving setup.py files for now. Also, will land this once the ice removal from scheduler diff has landed.

dgdonut (approved):

PR #551, opened 2021-08-17
convert json configs to data structures

JeanetteBruno:


harshita-chaudhary (approved):

scheduler/scheduler/config/configs.go
```diff
@@ -1,74 +1,84 @@
...
+	SagaLogConfig{
+		Type: "memory",
 	},
-	"SagaLog": {
-	  "Type": "memory"
-	}
 }
-`
 
 // localLocal config for local.local - !!! make sure this constant is added to SchedulerConfigs map above !!!
```
dgdonut, line 37: we could avoid this situation now by defining the configs inline directly in the map now

scheduler/scheduler/config/configs.go
```diff
@@ -1,74 +1,84 @@
...
+	"time"
+
+	"github.com/twitter/scoot/scheduler/server"
+	"github.com/twitter/scoot/workerapi/client"
+)
+
 // SchedulerConfigs the map of available configurations
-var SchedulerConfigs = map[string]string{
+var ServiceConfigs = map[string]ServiceConfig{
 	"default":      defaultConfig,
```
dgdonut, line 12: could use const strings to define these keys now, looks like a few random bits of code use these magic strings

JeanetteBruno:
I'm abandoning this diff.  It should be picked up by someone else later.

PR #552, opened 2021-08-18
adding logging for cluster update issue.  This is a branch off 67439c…

JeanetteBruno:
added logging to help debug delayed cluster update issue.  This branch's parent was 67439c not mater.

Do not merge this branch!!!!!

### DIFF
--- a/cloud/cluster/cluster.go
+++ b/cloud/cluster/cluster.go
@@ -78,7 +78,7 @@ func (c *Cluster) loop() {
 				outgoing = c.state.setAndDiff(nodes)
 			}
 			elapsed := time.Since(c.priorIterationTime)
-			log.Infof("putting changes on channel, time since last iteration: %s", time.Since(c.priorIterationTime))
+			log.Infof("putting %d changes on channel, time since last iteration: %s", len(outgoing), elapsed)
 			if c.Stats != nil {
 				c.Stats.Gauge(stats.ClusterTimeSinceLastUpdate_ms).Update(elapsed.Milliseconds())
 			}
--- a/cloud/cluster/cluster.go
+++ b/cloud/cluster/cluster.go
@@ -37,6 +37,7 @@ func NewCluster(state []Node, updateCh chan ClusterUpdate) *Cluster {
 		subs:               nil,
 		priorIterationTime: time.Now(),
 	}
+	s.Stats = c.Stats
 	go c.loop()
 	return c
 }
--- a/cloud/cluster/cluster.go
+++ b/cloud/cluster/cluster.go
@@ -5,14 +5,20 @@ package cluster
 
 import (
 	"sort"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/twitter/scoot/common/stats"
 )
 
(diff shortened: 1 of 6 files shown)


cloud/cluster/cluster.go
```diff
@@ -70,9 +77,16 @@ func (c *Cluster) loop() {
 				sort.Sort(NodeSorter(nodes))
 				outgoing = c.state.setAndDiff(nodes)
 			}
+			elapsed := time.Since(c.priorIterationTime)
+			log.Infof("putting changes on channel, time since last iteration: %s", time.Since(c.priorIterationTime))
```
dgdonut: recalculating elapsed here

cloud/cluster/cluster.go
```diff
@@ -70,9 +77,16 @@ func (c *Cluster) loop() {
...
+			if c.Stats != nil {
```
dgdonut, line 83: Only if we end up merging this to master:  I'm not a fan of putting if checks around the stats receiver every time we want to use it, we should initialize it to a nil/fake stats receiver on construction if none is provided instead.

dgdonut (commented):
new logging lgtm

cloud/cluster/cluster.go
JeanetteBruno: agreed

cloud/cluster/state.go
```diff
@@ -4,11 +4,15 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"sort"
+
+	"github.com/twitter/scoot/common/stats"
```
harshita-chaudhary, line 8: nit: import order

scheduler/server/cluster_state.go
```diff
@@ -214,11 +217,18 @@ func (c *clusterState) updateCluster() {
...
+		if c.stats != nil {
+			c.stats.Gauge(stats.ClusterStateTimeSinceLastUpdate_ms).Update(elapsed.Milliseconds())
+		}
+		c.priorUpdateTime = time.Now()
 		if !ok {
 			c.updateCh = nil
 		}
 		c.update(updates)
 	default:
+		log.Info("no updates were on the channel")
```
harshita-chaudhary, line 231: nit: Should we use log.Debug() here to avoid adding extra log lines(if this is not important to know in every step)?

harshita-chaudhary (approved):

scheduler/server/cluster_state.go
dgdonut: Generally yes, this would be more of a debug type log, and it's a pretty lousy thing to be logging to be honest. However for a one-off deploy this will help us debug what's going on with delays here; we won't ever merge this code into master. We also only log Debug level on staging, and we need to see this behavior on prod, so has to be Info or higher here.

harshita-chaudhary: Oh, I see. That makes sense.

JeanetteBruno:
abandoning pr, this was a temporary change to collect data.  pr 556 resulted from the analysis of the logging.

PR #555, opened 2021-08-24
Record worker idle latency

harshita-chaudhary:


runner/runners/queue.go
```diff
@@ -327,6 +327,7 @@ func (c *QueueController) loop() {
 	var watchCh chan runner.RunStatus
 	var updateDoneCh chan interface{}
 	updateRequested := false
+	var idleLatency = c.inv.stat.Latency(stats.WorkerIdleLatency_ms).Time()
```
cheyenne-martinez: [nit] assign "c.inv.stat.Latency(stats.WorkerIdleLatency_ms)" to a new variable so it isn't accessed 3 separate times.

cheyenne-martinez (approved):

runner/runners/queue.go
harshita-chaudhary: Yup, agree!

runner/runners/queue.go
```diff
@@ -409,6 +413,7 @@ func (c *QueueController) loop() {
 			case abortReq:
 				st, err := c.abort(r.runID)
 				r.resultCh <- result{st, err}
+				idleLatency.Time()
```
dgdonut: Does this work without this line in the abort handler here? From what I can tell even in case of abort, we should still hit the normal <-watchCh case below (in fact the abort call here effectively measures when we've signaled the real run to abort by closing c.runningAbort - does not guarantee the run has really exited).

harshita-chaudhary: Yes, agree that the watchCh will receive the signal with AbortStatus and it will anyway reset the current state. Removing this redundant line, thanks!

dgdonut (approved):

PR #556, opened 2021-08-30
Jbruno/remove subscription from cluster update

JeanetteBruno:
eliminate subscribe
fetcher/cluster/cluster_state/groupcache now operate as follows:
fetcher
- when cron fetcher is started it returns the channel it will use to send the latest node list to cluster.  This will be a buffered
channel preventing cluster from blocking fetcher

cluster
- when cluster is created it creates a cron fetcher, getting the (buffered) channel that will contain fetched node lists
- cluster processing loop 
  . empties the fetched node list channel (or pulls at most buffer size entries from the channel) and either
  . converts the node list to a list of NodeUpdates  and puts the node udpate events on the (buffered) node update channel (sending them to cluster_state)  or
  . sends the list of current nodes on a nodes request channel, when the nodes are requested (by apiserver)
- we are using buffered channels to prevent cluster blocking fetcher and to prevent cluster_state blocking cluster

cluster_state
- update() , pulls all []NodeUpdate lists off the nodes updates channel (shared with cluster), and updates the cluster

groupcache
- sends a nodes request channel to cluster
- pulls the nodes from the nodes request channel
- updates peers with this list of nodes

Also added were metrics for tracking the various cluster update frequencies, fetch duration and fetch errors.

Also updated a number of fmt.Errorf() statements per go's conventions (don't start error messages with capital letter, nor ending punctuation)


scheduler/server/cluster_state.go
```diff
@@ -205,47 +213,44 @@ func (c *clusterState) taskCompleted(nodeId cluster.NodeId, flaky bool) {
...
-func (c *clusterState) getNodeState(nodeId cluster.NodeId) (*nodeState, bool) {
+func (c *clusterState) getNodeState(nodeId cc.NodeId) (*nodeState, bool) {
+	c.clusterUpdatesMu.RLock()
+	defer c.clusterUpdatesMu.RUnlock()
+
 	ns, ok := c.nodes[nodeId]
 	return ns, ok
 }
 
+func (c *clusterState) getSuspendedNodeState(nodeId cc.NodeId) (*nodeState, bool) {
```
dgdonut: this function looks unused

cloud/cluster/cluster.go
```diff
@@ -5,121 +5,133 @@ package cluster
...
 // Cluster's ch channel accepts []Node and []NodeUpdate types, which then
 // get passed to its state to either SetAndDiff or UpdateAndFilter
-func NewCluster(state []Node, updateCh chan ClusterUpdate) *Cluster {
-	s := makeState(state)
-	c := &Cluster{
-		state:    s,
-		reqCh:    make(chan interface{}),
-		updateCh: updateCh,
-		subs:     nil,
+func NewCluster(stat stats.StatsReceiver) Cluster {
```
dgdonut: prefer to return *cluster instead of Cluster from constructors

scheduler/server/cluster_state.go
```diff
@@ -128,38 +133,38 @@ func newNodeState(node cluster.Node) *nodeState {
...
 	// This can go negative due to lost nodes, set lower bound at zero.
 	return max(0, len(c.nodes)-c.numRunning)
 }
 
-// Update ClusterState to reflect that a task has been scheduled on a particular node
+// update ClusterState to reflect that a task has been scheduled on a particular node
 // SnapshotId should be the value from the task definition associated with the given taskId.
-func (c *clusterState) taskScheduled(nodeId cluster.NodeId, jobId, taskId, snapshotId string) {
+func (c *clusterState) taskScheduled(nodeId cc.NodeId, jobId, taskId, snapshotId string) {
+	c.clusterUpdatesMu.RLock()
```
dgdonut: should be a write lock?

scheduler/server/cluster_state.go
```diff
@@ -179,10 +184,13 @@ func (c *clusterState) taskScheduled(nodeId cluster.NodeId, jobId, taskId, snaps
 	c.numRunning++
 }
 
-// Update ClusterState to reflect that a task has finished running on
+// update ClusterState to reflect that a task has finished running on
 // a particular node, whether successfully or unsuccessfully.
 // If the node isn't found then the node was already suspended and deleted, just decrement numRunning.
-func (c *clusterState) taskCompleted(nodeId cluster.NodeId, flaky bool) {
+func (c *clusterState) taskCompleted(nodeId cc.NodeId, flaky bool) {
+	c.clusterUpdatesMu.RLock()
```
dgdonut: should be a write lock?

scheduler/server/cluster_state.go
```diff
@@ -380,3 +375,50 @@ func (c *clusterState) status() string {
 	return fmt.Sprintf("now have %d healthy (%d free, %d running), and %d suspended",
 		len(c.nodes), c.numFree(), c.numRunning, len(c.suspendedNodes))
 }
+
+// the following functions implement (async) user initiated onlining and offlining a node
+// since these functions modify c.nodes and c.offlinedNodes, we need mutex around
+// uses of these structures.
+
+func (c *clusterState) HasOnlineNode(nodeId cc.NodeId) bool {
```
dgdonut, line 394: it looks like we had to introduce the cluster_state mutexes so that these APIs don't create race conditions. that could be a high price and the the mutex locking could introduce a bit of overhead into the processing of every task.

can we try to group the other mutex locks so that we don't have to do that? I believe that the scheduleTask operations happen all in one "chunk" ultimately based from the schedule tasks portion of scheduler's step(). The completedTask operations happen all in one chunk during the ProcessMessages portion of step(). The cluster update portion is another discrete part of step() [it looks like update() the way as-is here is already doing one top level lock, so that's good]. If we wrap those with mutex locks at a higher level we should be able to avoid individual locking & unlocking per each schedule task/complete task operation.

apiserver/main.go
```diff
@@ -111,9 +121,13 @@ func main() {
 	bundlestore.RunServer(bag, schema, configText)
 }
 
-func createCluster() *cluster.Cluster {
-	f := local.MakeFetcher("apiserver", "http_addr")
-	nodes, _ := f.Fetch()
-	updates := cluster.MakeFetchCron(f, time.NewTicker(time.Duration(1*time.Second)).C)
-	return cluster.NewCluster(nodes, updates)
-}
+// func createCluster(stat stats.StatsReceiver) (cc.Cluster, error) {
```
dgdonut: are we ever going to use this again?

cloud/cluster/cluster.go
```diff
@@ -5,121 +5,133 @@ package cluster
...
-	state    *state
-	reqCh    chan interface{}
-	updateCh chan ClusterUpdate
-	subs     []chan []NodeUpdate
+var ClusterUpdateLoopFrequency time.Duration = time.Duration(250) * time.Millisecond
+
+// Cluster represents a group of Nodes and has mechanism for
+// setting current Node list and receiving correlating Node updates.
+type Cluster interface {
+	RetrieveCurrentNodeUpdates() []NodeUpdate // used by scheduler's cluster state to get updates
```
dgdonut: nit: iface comments shouldn't rely on external things (scheduler, fetcher) for usage context

cloud/cluster/cluster.go
```diff
@@ -5,121 +5,133 @@ package cluster
...
+
+	stat stats.StatsReceiver
+
+	// latestFetchedNodes, when not nil, contains the latest node list from fetcher.  When
+	// cluster start processing the latestFetchedNodes, it grabs a copy an sets this to nil
+	latestFetchedNodes   []Node
+	latestFetchedNodesMu sync.RWMutex
+
+	// currentNodeUpdates, the node updates that the scheduler needs to process.  When
+	// scheduler gets the contents of currentNodeUpdates, currentNodeUpdates is reset to nil.
```
dgdonut: nit: same - caller could be anything, not just the scheduler

cloud/cluster/cluster.go
```diff
@@ -5,121 +5,133 @@ package cluster
...
 // get passed to its state to either SetAndDiff or UpdateAndFilter
-func NewCluster(state []Node, updateCh chan ClusterUpdate) *Cluster {
-	s := makeState(state)
-	c := &Cluster{
-		state:    s,
-		reqCh:    make(chan interface{}),
-		updateCh: updateCh,
-		subs:     nil,
+func NewCluster(stat stats.StatsReceiver, initialNodes []Node) Cluster {
+	s := makeState([]Node{})
```
dgdonut, line 56: why don't we want to makeState with initialNodes?

Makefile
```diff
@@ -84,7 +84,7 @@ smoketest: clean-procs
 	# Setup a local schedule against local workers (--strategy local.local)
 	# Then run (with go run) scootcl smoketest with 10 jobs, wait 1m
 	# We build the binaries because 'go run' won't consistently pass signals to our program.
-	$(FIRSTGOPATH)/bin/setup-cloud-scoot --strategy local.local run scootcl smoketest --num_jobs 10 --timeout 1m
+	$(FIRSTGOPATH)/bin/setup-cloud-scoot --strategy local.local run scootcl smoketest --num_jobs 10 --timeout 2m
```
dgdonut, line 87: was this flaky or do we think this diff has substantially slowed something down?

cloud/cluster/node.go
```diff
@@ -83,14 +83,14 @@ func NewRemove(id NodeId) NodeUpdate {
...
-	nu.UserInitiated = true
-	return nu
-}
-
-func NewUserInitiatedRemove(id NodeId) NodeUpdate {
-	nu := NewRemove(id)
-	nu.UserInitiated = true
-	return nu
-}
+// func NewUserInitiatedAdd(node Node) NodeUpdate {
```
dgdonut: are we going to reuse these?

common/stats/stats_names.go
```diff
@@ -18,6 +18,9 @@ const (
 	ClusterRunningNodes   = "runningNodes"
 	ClusterLostNodes      = "lostNodes"
 
+	ClusterNodeUpdateFreqMs = "clusterGetNodeUpdatesFreqMs"
```
dgdonut: nit: existing stats_names convention is "_ms" vs "Ms"

scheduler/server/cluster_state.go
JeanetteBruno: removed

scheduler/server/cluster_state.go
JeanetteBruno: we've removed mutexes

Makefile
JeanetteBruno: yes it was flaky

common/stats/stats_names.go
JeanetteBruno: fixed

cloud/cluster/node.go
JeanetteBruno: yes. I've reenabled the code

cloud/cluster/cluster.go
JeanetteBruno: starting the loop will acquire the first set of nodes.  The code is simpler without having to do an outer fetch.

scheduler/server/cluster_state.go
```diff
@@ -205,40 +205,42 @@ func (c *clusterState) taskCompleted(nodeId cluster.NodeId, flaky bool) {
...
+// it will process at most DefaultClusterSize sets of updates with each call
 func (c *clusterState) updateCluster() {
 	defer c.stats.Latency(stats.SchedUpdateClusterLatency_ms).Time().Stop()
-	select {
-	case updates, ok := <-c.updateCh:
-		if !ok {
-			c.updateCh = nil
+	allUpdates := []cc.NodeUpdate{}
+LOOP:
+	for i := 0; i < common.DefaultClusterChanSize; i++ {
```
dgdonut, line 219: The buffer sizes seem arbitrary - using the const here at 100, but internally we set a fetcher/cluster buffer of 1000 (and 100 for apiserver). Suggest using the const value in all of those places unless the other numbers were picked for a reason.

scheduler/server/cluster_state.go
```diff
@@ -380,3 +389,29 @@ func (c *clusterState) status() string {
 	return fmt.Sprintf("now have %d healthy (%d free, %d running), and %d suspended",
 		len(c.nodes), c.numFree(), c.numRunning, len(c.suspendedNodes))
 }
+
+// the following functions implement (async) user initiated onlining and offlining a node
+// since these functions modify c.nodes and c.offlinedNodes, we need mutex around
```
dgdonut: comment - these don't use mutexes anymore

cloud/cluster/cluster.go
```diff
@@ -5,121 +5,153 @@ package cluster
...
+	// latestFetchedNodesCh, contains the latest node list from fetcher.
+	fetchedNodesCh chan []Node
 
-func (c *Cluster) Subscribe() Subscription {
-	ch := make(chan Subscription)
-	c.reqCh <- ch
-	return <-ch
-}
+	// nodesReqCh, used (by groupcache) to request the current list of nodes
+	NodesReqCh NodeReqChType
```
dgdonut: this and NodesUpdatesCh don't need to be exported

cloud/cluster/cluster.go
```diff
@@ -5,121 +5,153 @@ package cluster
...
+// The processing loop continuously gets the latest nodes list from the fetched nodes channel and either
+// put the list of added/removed nodes on the nodes updates channel or if the nodes list is requested put the
+// current list of nodes on the nodes list channel
+func NewCluster(stat stats.StatsReceiver, fetcher Fetcher, useNodesUpdatesCh bool, updateFreq time.Duration, chBufferSize int) (chan []NodeUpdate, NodeReqChType) {
+	fetchedNodesCh := StartFetchCron(fetcher, updateFreq, chBufferSize, stat)
+
+	s := makeState([]Node{})
+	var updatesCh chan []NodeUpdate = nil
+	var nodesReqCh NodeReqChType = nil
+	if useNodesUpdatesCh {
```
dgdonut, line 59: This bool usage is odd, it looks like most of cluster internals supports doing both ways; it's up to the caller to decide which return value to use.

cloud/cluster/cluster.go
```diff
@@ -5,121 +5,153 @@ package cluster
...
+	}
+	// logging
+	chDesc := []string{}
+	if c.NodesUpdatesCh != nil {
+		chDesc = append(chDesc, "NodesUpdateCh")
+	}
+	if c.NodesReqCh != nil {
+		chDesc = append(chDesc, "NodesReqCh")
+	}
+	log.Infof("cluster loop starting with frequency %s, reporting on %s with channel buffer size %d", c.updateFreq, strings.Join(chDesc[:], ","), chBufferSize)
```
dgdonut: nits: chDest doesn't need to be a slice. Also the "[:]" in the Join call is unnecessary.

cloud/cluster/fetch_cron.go
```diff
@@ -2,42 +2,74 @@ package cluster
...
 // Defines the way in which a full set of Nodes in a Cluster is retrieved
 type Fetcher interface {
 	Fetch() ([]Node, error)
 }
 
-// Given a Fetcher implementation and a Ticker, returns a channel over which
-// ClusterUpdates will be sent to periodically from a new Goroutine
-func MakeFetchCron(f Fetcher, tickCh <-chan time.Time) chan ClusterUpdate {
-	outCh := make(chan ClusterUpdate)
+// Given a Fetcher implementation and a Ticker, start a ticker loop that
```
dgdonut: comment nit: doesn't take a ticker, takes a duration

cloud/cluster/cluster.go
```diff
@@ -5,121 +5,153 @@ package cluster
...
 
-func (c *Cluster) Subscribe() Subscription {
-	ch := make(chan Subscription)
-	c.reqCh <- ch
-	return <-ch
-}
+	// nodesReqCh, used (by groupcache) to request the current list of nodes
+	NodesReqCh NodeReqChType
+
+	// nodesUpdatesChan, the node updates that the scheduler needs to process.
```
dgdonut: these shouldn't define themselves in terms of scheduler and groupcache.

dgdonut (approved):
Seems like this could work

scheduler/server/cluster_state.go
JeanetteBruno: - 100 here was just an arbitrary number
- 1000 for scheduler vs 100 for apiserver in internal scoot was based on worker updates being batched at 1000 workers.

cloud/cluster/cluster.go
JeanetteBruno: we need to know which channel the caller will be pulling values from, to avoid blocking

cloud/cluster/cluster.go
JeanetteBruno: I fixed [:].  Using slice for chDest to get 'nice' commas printed in the log statement

cloud/cluster/cluster.go
ralphmcneal: I'm not a huge fan of the bool either. But it seems to highlight a bigger flaw in the design of the cluster, which services two callers differently.

cloud/cluster/cluster.go
ralphmcneal: nit: chDesc will only ever be one channel. You can add your comma in the format string vs. using the string join.

cloud/cluster/cluster.go
```diff
@@ -5,121 +5,153 @@ package cluster
...
+	return currentNodes
 }
 
-func (c *Cluster) closeSubscription(s *subscriber) {
-	c.reqCh <- s.inCh
+// addToCurrentNodeUpdates put the node updates on the nodes updates channel.
+func (c *Cluster) addToCurrentNodeUpdates(updates []NodeUpdate) {
+	c.NodesUpdatesCh <- updates
+	if len(updates) > 0 {
+		log.Infof("cluster has %d new node updates", len(updates))
```
ralphmcneal, line 147: should we only send nodes updates on the channel if len(updates) > 0?

cloud/cluster/cluster.go
```diff
@@ -5,121 +5,153 @@ package cluster
...
 }
 
-func (c *Cluster) current() []Node {
-	var r []Node
-	for _, v := range c.state.nodes {
-		r = append(r, v)
+// GetNodes get (a copy of) the list of nodes last fetched by fetcher
+func (c *Cluster) getNodes() []Node {
+	ret := []Node{}
+	for _, node := range c.state.nodes {
```
ralphmcneal, line 157: using built-int copy and/or preallocating your slice will be a big performance boost. Otherwise go will create many new new arrays behind the scenes( one on each append).

PR #557, opened 2021-08-31
Update polling frequency to be exponential 

cheyenne-martinez:


runner/runners/polling.go
```diff
@@ -48,6 +48,10 @@ func (r *PollingStatusQuerier) Query(q runner.Query, wait runner.Wait) ([]runner
 			return st, service, err
 		}
 		time.Sleep(r.period)
+
+		// Exponentially increase time between polls
+		r.period *= 2
```
ralphmcneal: This backoff is changing the state of the PollingStatusQuerier. Is that by design? I think you only want to increase the period for a specific method invocation, which means introducing a variable on the local call stack.

runner/runners/polling_test.go
```diff
@@ -149,3 +149,27 @@ func TestPolling_ErrorPolling(t *testing.T) {
...
 
 }
+
+func TestPollingWorker_PollFrequency(t *testing.T) {
+	defer teardown(t)
+	sim := execers.NewSimExecer()
+
+	outputCreator, err := NewHttpOutputCreator("")
+	if err != nil {
+		panic(err)
```
ralphmcneal: nit: change to t.Fatal

runner/runners/polling_test.go
```diff
@@ -149,3 +149,27 @@ func TestPolling_ErrorPolling(t *testing.T) {
 	ex.Resume()
 
 }
+
+func TestPollingWorker_PollFrequency(t *testing.T) {
+	defer teardown(t)
+	sim := execers.NewSimExecer()
```
ralphmcneal: Should this test setup a poller for testing?

runner/runners/polling_test.go
```diff
@@ -149,3 +149,27 @@ func TestPolling_ErrorPolling(t *testing.T) {
...
+	filerMap[runner.RunTypeScoot] = snapshot.FilerAndInitDoneCh{Filer: snapshots.MakeInvalidFiler(), IDC: nil}
+
+	r := NewSingleRunner(sim, filerMap, outputCreator, nil, stats.NopDirsMonitor, runner.EmptyID)
+	args := []string{"pause", "complete 0"}
+	startTime := time.Now()
+	expectedTime := startTime.Add(10 * time.Microsecond)
+	firstRun := run(t, r, args)
+	assertWait(t, r, firstRun, pending(), args...)
+
+	assert.WithinDuration(t, expectedTime, time.Now(), 1*time.Millisecond)
```
dgdonut: I'm trying to figure out exactly what is being tested here - looks like it is verifying that one poll interval has elapsed?  Which may not verify that the poll interval is increasing (I could be missing something). This is also a very high resolution time delta, so I'm wondering if this will end up being flaky depending on where it's being run.

FWIW I don't see any great way to test this behavior. Bad practice generally but I'm not averse to skipping if we can't make one in this instance.

runner/runners/polling.go
```diff
@@ -48,6 +48,10 @@ func (r *PollingStatusQuerier) Query(q runner.Query, wait runner.Wait) ([]runner
 			return st, service, err
 		}
 		time.Sleep(r.period)
+
+		// Exponentially increase time between polls
```
dgdonut: Suggest adding a max duration that we can just hardcode as a const in this file, (e.g. 10 sec?)

runner/runners/polling_test.go
dgdonut: +1

runner/runners/polling.go
dgdonut: +1

JeanetteBruno: also think max value should be at least 1minute

runner/runners/polling_test.go
JeanetteBruno: jbruno/testPollingFreq branch shows how we could instrument a test to expose the polling frequency for test assertions (warning ... it is not pretty)

runner/runners/polling.go
cheyenne-martinez: I set the max at 10s. Jeanette and I talked about the limit yesterday and agreed on 10s as suggested by Drew in another comment.

runner/runners/polling_test.go
```diff
@@ -149,3 +148,77 @@ func TestPolling_ErrorPolling(t *testing.T) {
...
+	single := NewSingleRunner(ex, filerMap, NewNullOutputCreator(), nil, stats.NopDirsMonitor, runner.EmptyID)
+	chaos := NewChaosRunner(single)
+	var nower runner.StatusQueryNower
+	nower = &instrumentedNower{ChaosRunner: *chaos}
+	poller := NewPollingService(chaos, nower, period)
+	return ex, nower, poller
+}
+
+type instrumentedNower struct {
+	ChaosRunner
```
dgdonut: I think the current go vet error can be fixed by making this a *ChaosRunner and passing the pointer on line 175.

That being said I tried this out locally and it failed 50% of the time for me - trying to test using sleep on the microsecond scale is going to be extremely flaky due to cpu speed and contention. Strongly recommend if we are going to spend more than ~minutes getting this particular test to work reliably that we scrap it and move on, it is clear enough that the fix in polling.go is going to work.

cheyenne-martinez: I reverted the test changes since it was also quite flaky for me.

dgdonut (approved):

PR #558, opened 2021-09-07
revert pr 521's task runner result state change

JeanetteBruno:
we can't have scheduler marking batches of workers as suspended when we are restarting a batch of 1000 workers.  We need to rethink what to do with 'lost' workers in this situation.

### DIFF
--- a/scheduler/server/task_runner_test.go
+++ b/scheduler/server/task_runner_test.go
@@ -11,6 +11,8 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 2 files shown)


scheduler/server/task_runner_test.go
```diff
@@ -11,6 +11,8 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 
+	"io/ioutil"
```
ryancouto, line 14: this should go in the top block with the other std lib stuff

ryancouto (approved):

PR #560, opened 2021-09-10
fix fetch error stat reporting

JeanetteBruno:
reset error to 0 when fetch succeeds


cloud/cluster/fetch_cron.go
```diff
@@ -68,6 +68,7 @@ func (c *fetchCron) doFetch() {
 		log.Infof("num fetched nodes changed from %d to %d", c.priorNumNodes, len(nodes))
 		c.priorNumNodes = len(nodes)
 	}
+	c.stat.Gauge(stats.ClusterFetchedError).Update(0)
```
dgdonut: Suggest using a Counter  in the original error check and just incrementing it and not trying to manage the 0/1 state, and then we can dashboard it up with the accumulated sum or a rate() that shows whenever a new error occurs.

dgdonut (approved):

JeanetteBruno:
abandoning pr - no longer needed

PR #561, opened 2021-09-13
Make retry log an error

ryancouto:


snapshot/store/http_store.go
```diff
@@ -21,7 +21,7 @@ func MakePesterClient() *pester.Client {
 	client.Backoff = pester.ExponentialBackoff
 	client.MaxRetries = DefaultHttpTries
 	client.LogHook = func(e pester.ErrEntry) {
-		log.Infof("Retrying after failed attempt: %+v", e)
+		log.Errorf("Retrying after failed attempt: %+v", e)
```
dgdonut, line 24: Can we also do lines 101, 119, 136, 161, 165 (I think that's it)

dgdonut (approved):

PR #562, opened 2021-09-14
Optimize update saga state

harshita-chaudhary:


saga/saga.go
```diff
@@ -195,54 +197,85 @@ func (s *Saga) updateSagaState(msgs []SagaMessage) error {
 // is one per saga currently executing.  This ensures all updates are applied
 // in order to a saga.  Also controls access to the SagaState so its is
 // updated in a thread safe manner
+// Processes at most SagaUpdateChSize number of updates at a time
 func (s *Saga) updateSagaStateLoop() {
+	updates := []sagaUpdate{}
 	for update := range s.updateCh {
-		s.updateSaga(update)
+		updates = append(updates, update)
+		if len(s.updateCh) == 0 || len(updates) == common.DefaultSagaUpdateChSize {
```
ryancouto, line 213: can we add a stats handler to *Saga & record how long it takes to loop through these operations?

harshita-chaudhary: Yeah sure, it would be interesting to record this, along with the number of updates processed each time.

saga/saga.go
```diff
@@ -195,54 +197,85 @@ func (s *Saga) updateSagaState(msgs []SagaMessage) error {
...
-	s.state, err = logMessages(s.state, update.msgs, s.log)
-	update.resultCh <- err
+	var validUpdates []sagaUpdate
+	var validUpdateMsgs []SagaMessage
+
+	for _, update := range updates {
+		err = nil
+
+		// verify that the applied message(s) results in a valid state
+		if len(update.msgs) == 1 {
```
dgdonut, line 247: Why have a separate case for len == 1?

saga/saga_state.go
```diff
@@ -244,6 +244,23 @@ func (state *SagaState) addTaskData(taskId string, msgType SagaMessageType, data
 	}
 }
 
+/*
+ * Applies the supplied messages atomically to the supplied sagaState.
+ * Returns the mutated sate if all the messages are valid,
+ * otherwise returns the original state if any of the messages are invalid along with the error
+ */
+func bulkUpdateSagaState(state *SagaState, msgs []SagaMessage) (*SagaState, error) {
+	originalState := copySagaState(state)
```
dgdonut, line 253: Tried thinking about how to optimize this and can't come up with anything that's not super complicated or awful, like bailing out of all updates if one fails.  I got a 25-30% bump by commenting it out entirely.

saga/sagalogs/file_test.go
```diff
@@ -170,3 +172,54 @@ func TestGetMessages_SagaDoesNotExist(t *testing.T) {
...
+	* DefaultSagaUpdateChSize=10:	 1 iteration(s)  2209322245 ns/op	6610864 B/op  29788 allocs/op
+	* DefaultSagaUpdateChSize=100:	 3 iteration(s)	  368981765 ns/op 	2278261 B/op  23838 allocs/op
+	* DefaultSagaUpdateChSize=1000:	 7 iteration(s)	  174756943 ns/op	1861685 B/op  23132 allocs/op
+
+	For sagasCount = 10, updatesPerSaga = 100 and following DefaultSagaUpdateChSize values
+	* DefaultSagaUpdateChSize=1:     1 iteration(s)	17434411997 ns/op  6995688 B/op  45573 allocs/op
+	* DefaultSagaUpdateChSize=10:	 1 iteration(s)  2553031047 ns/op  2832392 B/op  28176 allocs/op
+	* DefaultSagaUpdateChSize=100:	 2 iteration(s)	  705908630 ns/op  1966188 B/op  23760 allocs/op
+*/
+func BenchmarkProcessUpdatesInFileSagaLog(b *testing.B) {
```
dgdonut: I hate to say it but I was reading these for a minute before I realized it was file log - this is good info but we only use file log in testing and don't care about the perf. Just worried about it possibly causing confusion.

saga/sagalogs/memory_test.go
```diff
@@ -119,3 +121,54 @@ func TestMemorySagaGC(t *testing.T) {
...
+/*
+	Set different saga update channel size using DefaultSagaUpdateChSize
+	Benchmarked on cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
+
+	For sagasCount = 10, updatesPerSaga = 10000 and following DefaultSagaUpdateChSize values
+	* DefaultSagaUpdateChSize=1:     1 iteration(s)	124623537403 ns/op
+	* DefaultSagaUpdateChSize=10:	 1 iteration(s)  12299526501 ns/op
+	* DefaultSagaUpdateChSize=100:	 1 iteration(s)	  1864785416 ns/op
+	* DefaultSagaUpdateChSize=1000:	 3 iteration(s)	   451137271 ns/op
+	* DefaultSagaUpdateChSize=10000: 7 iteration(s)	   176294414 ns/op
```
dgdonut, line 134: Hopefully gives a big boost to utilization!

dgdonut (commented):
Would also like to have stats around when the scheduler adds StartTask (happens twice) and EndTask messages in scheduler/server/task_runner.go, so we know how long the sched is waiting there. Fine to do that here or in a separate PR.

saga/saga.go
harshita-chaudhary: Added this separate case to avoid copying sagaState(which takes some time) in case of single message, and call updateSagaState directly. Most updates will contain a single message. The only place I see bulk messages being used is while killing multiple jobs and aborting tasks that is called by getTaskAssignments(). 

If we wanna completely skip the copySagaState() call in bulkUpdateSagaState function, we can avoid this case of len == 1 without any impact on time. I'm up for either way.

saga/saga_state.go
harshita-chaudhary: Agree, a thing to note is that this case(updates with bulk messages) does not occur so often(more info in the above comment). If we don't save the old state and error occurs, we'll have to error the entire batch of updates(upto 10,000 in this case), since the current state would be unusable.

saga/sagalogs/file_test.go
harshita-chaudhary: Sure, since we're not using it anywhere, I'll remove this benchmark.

saga/sagalogs/memory_test.go
harshita-chaudhary: Yeah, fingers crossed! Might have to tune the DefaultSagaUpdateChSize or the limit on the number of updates processed together and see how it impacts the performance in real-time.

harshita-chaudhary:
> Would also like to have stats around when the scheduler adds StartTask (happens twice) and EndTask messages in scheduler/server/task_runner.go, so we know how long the sched is waiting there. Fine to do that here or in a separate PR.

Would it be more useful to measure the time taken by each update, i.e., by updateSagaState() in general, rather than measuring StartTask & EndTask only?

saga/saga.go
```diff
@@ -199,21 +205,32 @@ func (s *Saga) updateSagaState(msgs []SagaMessage) error {
 // updated in a thread safe manner
 // Processes at most SagaUpdateChSize number of updates at a time
 func (s *Saga) updateSagaStateLoop() {
+	var loopLatency = s.stat.Latency(stats.SagaUpdateStateLoopLatency_ms)
```
dgdonut: you can append .Time() here and ditch the additional line below

saga/saga.go
```diff
@@ -199,21 +205,32 @@ func (s *Saga) updateSagaState(msgs []SagaMessage) error {
...
+			loopLatency.Stop()
+			s.stat.Gauge(stats.SagaNumUpdatesProcessed).Update(int64(len(updates)))
+
 			s.updateSaga(updates)
+
 			updates = []sagaUpdate{}
+			loopLatency.Time()
 		}
 	}
+	loopLatency.Stop()
```
dgdonut: I think we do not actually want to call this as we will end up measuring an arbitrary amount of time between the last update and when s.updateCh is getting closed, which is unrelated

saga/saga.go
harshita-chaudhary: yup!

saga/saga.go
harshita-chaudhary: yeah, makes sense

dgdonut (approved):

common/stats/stats.go
```diff
@@ -425,28 +425,50 @@ type metricLatency struct {
...
-func (l *metricLatency) Stop()         { l.Update(Time.Since(l.start).Nanoseconds()) }
 func (l *metricLatency) Capture() Latency {
-	return &metricLatency{l.Histogram.Snapshot(), l.start, l.precision}
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
+	return &metricLatency{l.Histogram.Snapshot(), l.start, l.precision, sync.RWMutex{}}
 }
+
 func (l *metricLatency) GetPrecision() time.Duration {
+	l.mutex.Lock()
```
dgdonut: Can use Rlock/RUnlock GetPrecision. Otherwise LGTM

harshita-chaudhary: yeah, that works. Updated.

PR #566, opened 2021-10-29
Clean and reset repo even if snapshot is currently checked out

ryancouto:
This fixes a bug where if a snapshot is currently checked out on a worker & the worker receives a new task with the same snapshot, we immediately run the task without modification to the working directory. This is an issue if the previous task has modified the working state at all, and can lead to inconsistent results.


ryancouto:
abandoning in favor of https://github.com/twitter/scoot/pull/567

PR #568, opened 2021-12-03
Jbruno/mgci1674 investigate bazel memory usage

JeanetteBruno:
objective: to monitor for task based memory leaks - including memory usage on processes whose pids are not related to the workerserver pid. (did the task leave a process running that was not cleaned up with cleanupProcs()

approach:
- before exec'ing a task, get the total memory in use by all processes being run by the current user
- after the exec is finished, get the total memory in use by all processes being run by the current user
- compute the delta between finished memory usage and starting memory usage
- report >0 delta's to monitoring (qualified with the command that was run)
- if the delta is over a threshold, log the memory usage and task info to help identify the specific tasks that are leaving processes that are consuming memory. 

also refactored the list of process maps (allProcesses, groupProcesses, parentProcesses) into a common processMaps structure so as to simplify function signatures.


runner/execer/os/os_execer.go
```diff
@@ -21,10 +23,19 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+var memNotFreedMemAlertThreshold = 1000000 // TODO should we use a value other than 1M?
```
ryancouto: Can we remove this TODO & pass the value through as a flag to workerserver? Allows us to update it without making changes to open source

runner/execer/os/os_execer.go
```diff
@@ -78,6 +90,14 @@ func (e *osExecer) Exec(command execer.Command) (execer.Process, error) {
 		return nil, errors.New("No command specified.")
 	}
 
+	var startMem int
```
ryancouto: we can skip these two var declarations and just use := on line 95

runner/execer/os/os_execer_test.go
```diff
@@ -148,7 +176,7 @@ func TestAbortCatch(t *testing.T) {
 	}
 
 	proc.Abort()
-	usage, err = e.memUsage(pid)
+	usage, _ = e.memUsage(pid)
```
ryancouto: should we check this err

JeanetteBruno: I removed the call e.memUsage(pid).... it was not needed.

ryancouto (approved):

runner/execer/os/os_execer.go
```diff
@@ -577,5 +597,52 @@ func cleanupProcs(pgid int) (err error) {
...
+		return 0, err
+	}
+	cUser, err := user.Current()
+	if err != nil {
+		return 0, err
+	}
+
+	// compute the total user's memory usage
+	totalMem := 0
+	for _, proc := range pm.byUser[cUser.Username] {
```
dgdonut, line 623: Do we know how this will work e.g. on a shared machine with 2 separate scoot workers running?

JeanetteBruno: good question - I don't really know.

JeanetteBruno: I was able to verify that the ps command results included processes from multiple scoot workers running on the same machine.... I didn't see a simple way to restrict the process list to those for a specific scoot worker, so I updated the code to only report memory usage when it can find only 1 './workerserver' process.  

Also, this code should work on linux machines, but it will fail on mac because it uses ps options (user, cmd) that are not available on macs.  The code could be refactored to be platform specific, but I'm not sure it's worth it.

I will not merge this branch.

PR #570, opened 2021-12-09
Jbruno/remove ice from worker

JeanetteBruno:
eliminate worker's use of ice.

was also able to eliminate worker config stuff since worker doesn't use any config structure.


ryancouto (approved):
awesome!

PR #572, opened 2022-02-03
Make processWatcher methods public

ryancouto:
This diff
1. Renames proc[ess]Getter to proc[ess]Watcher to more accurately reflect its use case
2. Makes proc[ess]Watcher's methods public for use in other packages
3. Removes package name stutter from the runner/execer/os package
4. Splits os_execer.go into separate files for each struct, i.e. execer.go, process.go, & process_watcher.go


cheyenne-martinez (approved):

ralphmcneal (approved):
lgtm

PR #579, opened 2022-02-22
upgrade mock to match internal version & run go mod tidy

ryancouto:


dgdonut:
Was talking with @ralphmcneal the other day about this - the other issue here is in Makefile we get mockgen with:
`go get github.com/golang/mock/mockgen@58cd061d09382b6011f84c1291ebe50ef2e25bab`
which is an older pinned version and conflicts with the go mod value here. So we need to find a version here that versoin stable and still works with make generate.

ryancouto:
ah gotcha. will close and cut a ticket

PR #581, opened 2022-02-24
pre/post processors

ryancouto:
Introduce the ability to define pre/post processing funcs that run before/after a task

### DIFF
--- a/runner/runners/invoke.go
+++ b/runner/runners/invoke.go
@@ -108,10 +108,11 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 
 	// set up pre/postprocessors
 	for _, pp := range inv.preprocessors {
+		var rf *runtime.Func
 		msg := "running preprocessor"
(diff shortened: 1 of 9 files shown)


cheyenne-martinez (approved):

runner/runners/invoke.go
```diff
@@ -100,6 +104,16 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
 	var co snapshot.Checkout
 	checkoutCh := make(chan error)
 
+	// set up pre/postprocessors
+	for _, pp := range inv.preprocessors {
```
dgdonut, line 108: Suggest we at least add an error return to pre/post, and log the func and error value if non-nil. Consider also logging the func to begin with at debug or info (e.g. "running pre func <x>")

dgdonut (approved):

runner/runners/invoke.go
ryancouto: had that in the internal diff but i like having it here better, good call.

PR #586, opened 2022-03-10
Added interface for log uploading by workers, replacing ingesting bundles

harshita-chaudhary:
Existing way of uploading logs: create a snapshot of log directory and upload to bundlestore
Updated way: upload logs directly to a permanent storage by implementing LogUploader, skipping output snapshot creation and ingestion

### DIFF
--- a/common/stats/stats_names.go
+++ b/common/stats/stats_names.go
@@ -486,13 +486,28 @@ const (
 	WorkerTimeSinceLastContactGauge_ms = "timeSinceLastContactGauge_ms"
 
 	/*
-		the number of times the worker uploaded a snapshot to bundlestore
+		the number of times the worker uploaded a log to storage
 	*/
 	WorkerUploads = "workerUploads"
 
 	/*
-		the amount of time spent uploading snapshots to bundlestore.  This includes time for
+		the number of times the worker log upload timed out
+	*/
+	WorkerLogUploadTimeouts = "workerLogUploadTimeouts"
+
+	/*
+		the number of times the worker log upload was retried
+	*/
+	WorkerLogUploadRetries = "workerLogUploadRetries"
+
+	/*
+		the amount of time spent uploading the log to storage. This includes time for
 		successful as well as erroring uploads
+	*/
+	WorkerLogUploadLatency_ms = "workerLogUploadLatency_ms"
+
+	/*
+		the overall time spent uploading logs(stderr, stdout & stdlog)
 	*/
 	WorkerUploadLatency_ms = "workerUploadLatency_ms"
 
--- a/common/stats/stats_names.go
+++ b/common/stats/stats_names.go
@@ -491,6 +491,11 @@ const (
 	WorkerUploads = "workerUploads"
 
 	/*
+		the number of times the worker log upload failed
+	*/
+	WorkerLogUploadFailures = "WorkerLogUploadFailures"
+
+	/*
 		the number of times the worker log upload timed out
 	*/
 	WorkerLogUploadTimeouts = "workerLogUploadTimeouts"
--- a/common/stats/stats_names.go
+++ b/common/stats/stats_names.go
@@ -486,22 +486,22 @@ const (
 	WorkerTimeSinceLastContactGauge_ms = "timeSinceLastContactGauge_ms"
 
 	/*
-		the number of times the worker uploaded a log to storage
+		the number of times the worker uploaded a log(stdout/stderr/stdlog) to storage
 	*/
 	WorkerUploads = "workerUploads"
 
 	/*
-		the number of times the worker log upload failed
+		the number of times the worker log(stdout/stderr/stdlog) upload failed
 	*/
-	WorkerLogUploadFailures = "WorkerLogUploadFailures"
+	WorkerLogUploadFailures = "workerLogUploadFailures"
 
 	/*
-		the number of times the worker log upload timed out
+		the number of times the worker log(stdout/stderr/stdlog) upload timed out
 	*/
 	WorkerLogUploadTimeouts = "workerLogUploadTimeouts"
 
 	/*
-		the number of times the worker log upload was retried
+		the number of times the worker log(stdout/stderr/stdlog) upload was retried
 	*/
 	WorkerLogUploadRetries = "workerLogUploadRetries"
 
@@ -512,7 +512,7 @@ const (
 	WorkerLogUploadLatency_ms = "workerLogUploadLatency_ms"
 
 	/*
-		the overall time spent uploading logs(stderr, stdout & stdlog)
+		the overall time spent uploading all three logs(stderr, stdout & stdlog)
 	*/
 	WorkerUploadLatency_ms = "workerUploadLatency_ms"
 
--- a/runner/runners/invoke.go
+++ b/runner/runners/invoke.go
@@ -5,10 +5,9 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"os"
-	"path/filepath"
 	"time"
 
+	uuid "github.com/nu7hatch/gouuid"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/twitter/scoot/common/errors"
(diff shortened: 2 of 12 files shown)


runner/runners/log_uploader.go
```diff
@@ -0,0 +1,31 @@
...
+// Creates a new LogUploader that will just block on cancel channel
+func NewWaitingLogUploader() *NoopWaitingLogUploader {
+	return &NoopWaitingLogUploader{}
+}
+
+type NoopWaitingLogUploader struct{}
+
+func (u *NoopWaitingLogUploader) UploadLog(blobId string, filepath string, cancelCh chan struct{}) (string, error) {
+	<-cancelCh
+	return "fake_blob_url", nil
```
dgdonut: nit: may make more sense to return an err in this case although it doesn't look like the test cares either way.

runner/runners/single_test.go
```diff
@@ -105,6 +105,34 @@ func TestAbort(t *testing.T) {
...
+	if err != nil {
+		panic(err)
+	}
+	filerMap := runner.MakeRunTypeMap()
+	filerMap[runner.RunTypeScoot] = snapshot.FilerAndInitDoneCh{Filer: snapshots.MakeInvalidFiler(), IDC: nil}
+	r := NewSingleRunner(sim, filerMap, outputCreator, nil, stats.NopDirsMonitor, runner.EmptyID, []func() error{}, []func() error{}, NewWaitingLogUploader())
+	args := []string{"complete 0"}
+	runID := run(t, r, args)
+	assertWait(t, r, runID, running(), args...)
+	time.Sleep(10 * time.Millisecond)
```
dgdonut: Is this prone to fail or skip the upload without the Sleep here? Wonder if asa an alternative we could add an additional channel to NoopWaitingLogUploader such that UploadLog would wait to receive on before sending or closing its cancelCh, and we could signal that additional channel here, verifying we actually hit that code as intended. WDYT?

runner/runners/invoke.go
```diff
@@ -368,71 +369,62 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
...
-			if err = stageLogFiles(tmp, stdoutName, stderrName, stdlogName, stdout, stderr, stdlog); err != nil {
-				return runner.FailedStatus(id, errors.NewError(err, errors.PostExecFailureExitCode),
-					tags.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID, Tag: cmd.Tag})
-			}
+			var stderrUrl, stdoutUrl string
+			// only upload logs to a permanent location if a log uploader is initialized
+			if inv.uploader != nil {
+				uploadTimer := inv.stat.Latency(stats.WorkerUploadLatency_ms).Time()
+				defer func() {
+					uploadTimer.Stop()
```
dgdonut: can just `defer uploadTimer.Stop()` I believe. `defer inv.stat.Latency(stats.WorkerUploadLatency_ms).Time().Stop()` might even work

runner/runners/invoke.go
```diff
@@ -452,38 +444,33 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
...
-	}
-	if err := copyLogFile(tmpDir, stderrName, errPath); err != nil {
-		return err
-	}
-	if err := copyLogFile(tmpDir, stdlogName, logPath); err != nil {
-		return err
-	}
-	return nil
-}
+	uploadTimer := inv.stat.Latency(stats.WorkerLogUploadLatency_ms).Time()
```
dgdonut: same comment re defers

common/stats/stats_names.go
```diff
@@ -486,14 +486,29 @@ const (
...
-		the number of times the worker uploaded a snapshot to bundlestore
+		the number of times the worker uploaded a log to storage
 	*/
 	WorkerUploads = "workerUploads"
 
 	/*
-		the amount of time spent uploading snapshots to bundlestore.  This includes time for
+		the number of times the worker log upload timed out
+	*/
+	WorkerLogUploadTimeouts = "workerLogUploadTimeouts"
```
dgdonut, line 501: These don't seem used and it doesn't look like we do retries

runner/runners/invoke.go
```diff
@@ -452,38 +444,33 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
...
+		url, err = inv.uploader.UploadLog(logId, filepath, uploadCancelCh)
+		uploadCh <- err
+	}()
+	select {
+	case abort := <-abortCh:
+		uploadCancelCh <- abort
+		return "", true
+	case err := <-uploadCh:
+		if err != nil {
+			log.Errorf("error uploading stdlog to storage: %s", err)
```
dgdonut: error metric would be useful here

ryancouto (approved):

runner/runners/invoke.go
```diff
@@ -368,71 +369,62 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
...
+
+				// generate a unique id that's appended to the job id to create a unique identifier for logs
+				logUid, _ := uuid.NewV4()
+				stdlogName := "stdlog"
+				stderrName := "stderr"
+				stdoutName := "stdout"
+
+				var isAborted bool
+				// upload stdlog
+				logId := fmt.Sprintf("%s_%s/%s", cmd.JobID, logUid, stdlogName)
```
dgdonut, line 385: Seems like this solidifies the ID of the log in the worker as opposed to leaving it in the hands of the LogUploader. Even if we end up doing something similar in our Uploader it would give us more flexibility there.

runner/runners/log_uploader.go
```diff
@@ -0,0 +1,31 @@
+package runners
+
+// LogUploader provides functionality to upload logs to permanent storage, used by invoker to upload task logs
+type LogUploader interface {
+	UploadLog(blobId string, filepath string, cancelCh chan struct{}) (string, error)
```
dgdonut: Was going to say we tend to prefer `blobID` to `blobId` as far as naming, but we have occurrences of the latter around anyway.

common/stats/stats_names.go
harshita-chaudhary: These are being used by the implementation of LogUploader. I think it can moved to the stats file in closed source for better maintainability.

dgdonut: Ignore, commented before I read internal diff.

runner/runners/invoke.go
harshita-chaudhary: Oh yes, makes sense.

runner/runners/invoke.go
```diff
@@ -368,71 +369,62 @@ func (inv *Invoker) run(cmd *runner.Command, id runner.RunID, abortCh chan struc
...
-			// Note: only modifying stdout/stderr refs when we're actively working with snapshotID.
-			status := runner.CompleteStatus(id, snapshotID, st.ExitCode,
+			status := runner.CompleteStatus(id, "", st.ExitCode,
 				tags.LogTags{JobID: cmd.JobID, TaskID: cmd.TaskID, Tag: cmd.Tag})
-			if cmd.SnapshotID != "" {
-				status.StdoutRef = snapshotID + "/" + stdoutName
-				status.StderrRef = snapshotID + "/" + stderrName
+
+			// Note: only modify stdout/stderr refs when logs are successfully uploaded to storage
+			if stderrUrl != "" {
```
dgdonut: Nit but these ifs are unnecessary if following std golang convention of setting these to "" in the error case (which we are)

runner/runners/invoke.go
harshita-chaudhary: Agree, will update.

runner/runners/invoke.go
harshita-chaudhary: Yep, will add one

runner/runners/log_uploader.go
harshita-chaudhary: Updating to blobID

runner/runners/log_uploader.go
harshita-chaudhary: Will update to return err.

runner/runners/single_test.go
harshita-chaudhary: Yeah, that's a good idea, will update this test. Had added sleep to make sure we're not aborting too early and skipping log upload.

runner/runners/invoke.go
harshita-chaudhary: Creating id here because this ensures that the urls for all stdout/err/log files are similar as sickle just replaces the last "stderr" part with "stdlog". This won't be required once we start passing just stdlog in task status and maybe we can get back to this once we do that. Also, wanted to use the job id in the unique identifier for the log but it does not matter.

runner/runners/invoke.go
harshita-chaudhary: yep, removed!

dgdonut (approved):
LGTM, thanks

PR #589, opened 2022-06-21
Ability to override worker memory utilization function

harshita-chaudhary:
Introduce ability to pass a memory utilization function when starting workerserver. If passed, use this function to calculate worker memory and compare against the memory cap.

### DIFF
--- a/runner/execer/os/execer.go
+++ b/runner/execer/os/execer.go
@@ -31,7 +31,7 @@ type WriterDelegater interface {
 type execer struct {
 	// Best effort monitoring of command to kill it if resident memory usage exceeds this cap
 	memCap            scootexecer.Memory
-	getMemUtilization func() (int64, error)
+	getMemUtilization func(int) (scootexecer.Memory, error)
 	stat              stats.StatsReceiver
 	pw                ProcessWatcher
 }
@@ -45,8 +45,18 @@ func NewBoundedExecer(memCap scootexecer.Memory, getMemUtilization func() (int64
 	if stat != nil {
 		oe.stat = stat
 	}
+	// if not nil, use the provided function to get memory utilization,
+	// otherwise get the memory usage of the current process and its subprocesses
 	if getMemUtilization != nil {
-		oe.getMemUtilization = getMemUtilization
+		oe.getMemUtilization = func(int) (scootexecer.Memory, error) {
+			mem, err := getMemUtilization()
+			if err != nil {
+				return 0, err
+			}
+			return scootexecer.Memory(mem), err
+		}
+	} else {
+		oe.getMemUtilization = oe.pw.MemUsage
 	}
 	return oe
 }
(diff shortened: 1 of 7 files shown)


runner/execer/os/execer.go
```diff
@@ -157,11 +161,25 @@ func (e *execer) monitorMem(p *process, memCh chan scootexecer.ProcessStatus) {
...
-			mem, _ := e.pw.MemUsage(pid)
+			var mem scootexecer.Memory
+			// if not nil, use the provided function to get memory utilization,
+			// otherwise get the memory usage of the current process and its subprocesses
+			if e.getMemUtilization != nil {
+				var memory int64
+				memory, err = e.getMemUtilization()
+				mem = scootexecer.Memory(memory)
+			} else {
+				mem, err = e.pw.MemUsage(pid)
```
dgdonut: Just one style comment: is there a way we can initialize such that we assign getMemUtilization() to some wrapper that calls e.pw.MemUsage(pid) in this case, and we don't need the if/else branch here?

harshita-chaudhary: Moved to NewBoundedExecer() func.

runner/execer/os/execer.go
```diff
@@ -30,20 +30,34 @@ type WriterDelegater interface {
...
 	if stat != nil {
 		oe.stat = stat
 	}
+	// if not nil, use the provided function to get memory utilization,
+	// otherwise get the memory usage of the current process and its subprocesses
+	if getMemUtilization != nil {
+		oe.getMemUtilization = func(int) (scootexecer.Memory, error) {
+			mem, err := getMemUtilization()
+			if err != nil {
+				return 0, err
```
ryancouto, line 54: nit: would -1 make more sense here

runner/execer/os/execer.go
```diff
@@ -212,7 +232,7 @@ func (e *execer) monitorMem(p *process, memCh chan scootexecer.ProcessStatus) {
 							"tag":    p.Tag,
 							"jobID":  p.JobID,
 							"taskID": p.TaskID,
-						}).Tracef("ps after increasing mem_cap utilization for pid %d", pid)
+						}).Tracef("ps after increased memory utilization for pid %d", pid)
```
ryancouto: not really related, but do we use trace logs anywhere? i think staging/devel use debug.

ryancouto (approved):

runner/execer/os/execer.go
harshita-chaudhary: scootexecer.Memory is of type uint64, hence made it 0 as we receive -1 if there's an error.

runner/execer/os/execer.go
harshita-chaudhary: yeah, don't think so. Will update this to debug.

dgdonut (approved):
thanks!

PR #591, opened 2022-07-26
Add ability to mark worker unhealthy in case of persistent high memory & add monitoring metrics

harshita-chaudhary:
**Problem**

Workers can get into bad states where they continuously have high memory consumption exceeding the soft cap threshold(could be due to persistent processes or memory leaks from previous runs). This would fail tasks with the memory cap exceeded exception as soon as they start running. Fixing these workers would require manual intervention to kill and have the worker reassigned to a fresh, uncorrupted host.

**Solution**

Add ability to mark worker status as unhealthy if memory consumption is observed to be higher than the threshold as soon as a task starts, and allow scheduler to query if the worker is unhealthy. The scheduler marks unhealthy workers as suspended to prevent further tasks from being scheduled on the corrupt host. The unhealthy suspended worker nodes are also marked lost by the scheduler since we don't yet have a way to recover an unhealthy node. The worker health check API(/health) can be used to identify and setup an automatic restart mechanism. Also, record metrics for better monitoring and tracking of memory related errors.

### DIFF
--- a/runner/execer/os/execer.go
+++ b/runner/execer/os/execer.go
@@ -141,10 +141,6 @@ func (e *execer) monitorMem(p *process, memCh chan scootexecer.ProcessStatus) {
 	} else {
 		defer cleanupProcs(pgid)
 	}
-	thresholdsIdx := 0
-	reportThresholds := []float64{0, .25, .5, .75, .85, .9, .93, .95, .96, .97, .98, .99, 1}
-	memTicker := time.NewTicker(500 * time.Millisecond)
-	defer memTicker.Stop()
 	log.WithFields(
 		log.Fields{
 			"pid":    pid,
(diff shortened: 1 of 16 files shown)


cheyenne-martinez (approved):

runner/execer/os/execer.go
```diff
@@ -197,11 +197,29 @@ func (e *execer) monitorMem(p *process, memCh chan scootexecer.ProcessStatus) {
 					Error:    msg,
 					ExitCode: 1,
 				}
+				// check whether memory consumption is above threshold immediately on process start
+				// mostly indicates that memory utilization was already above cap when the process started
+				if thresholdsIdx == 0 {
```
ryancouto: this is sort of confusing... are we checking if thresholdsIdx is 0 because that means we haven't measured the memory above 25% of the cap yet?
I think this would be easier to understand if we refactored this monitorMem func a bit. Pull out the block inside of `if mem >= e.memCap`, e.g.`if mem >= e.memCap {
    memCapKillHandler()
}`, and then we could just explicitly check the mem usage once before we start the monitorMem goroutine

runner/status.go
```diff
@@ -155,9 +155,10 @@ func CompleteStatus(runID RunID, snapshotID string, exitCode errors.ExitCode, ta
 // This is for overall runner status, just 'initialized' status and error for now.
 type ServiceStatus struct {
 	Initialized bool
+	IsHealthy   bool
 	Error       error
 }
 
 func (s ServiceStatus) String() string {
-	return fmt.Sprintf("--- Service Status ---\n\tInitialized:%t\n", s.Initialized)
+	return fmt.Sprintf("--- Service Status ---\n\tInitialized:%t\n\tIsHealthy:%t\n", s.Initialized, s.IsHealthy)
```
ryancouto: nit, but can we add spaces in front of the colons? increases readability imo

worker/starter/start_server.go
```diff
@@ -27,11 +27,19 @@ type WorkerStatusHTTPHandler struct {
...
 func (h *WorkerStatusHTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	_, svcStatus, err := h.svc.QueryNow(runner.Query{})
-	if err == nil && svcStatus.Initialized {
+	if err == nil {
+		if !svcStatus.Initialized {
+			fmt.Fprintf(w, "not initialized")
+			return
+		}
+		if !svcStatus.IsHealthy {
+			fmt.Fprintf(w, "failed")
```
ryancouto: nit: let's put "not healthy" or something similar here

ryancouto (approved):
overall lgtm!

runner/execer/os/execer.go
harshita-chaudhary: We're checking if threshold index is 0 because it implies that it's the first time we're checking memory. I agree that this could be a bit confusing, going to refactor this.

runner/status.go
harshita-chaudhary: Sure.

worker/starter/start_server.go
harshita-chaudhary: Yep, will update

runner/execer/os/execer.go
harshita-chaudhary: Refactored

runner/status.go
harshita-chaudhary: Done

worker/starter/start_server.go
harshita-chaudhary: Done.

ryancouto (approved):
lgtm!

